PR #19, opened 2024-11-17
Unify agents

yellowcap:
Combined agents using subgraphs, this includes glad, location, searching layers, and getting info from blog posts.

An API streams the results to a streamlit app. No rendering so far, need to make this pretty.

This requires two `chroma` RAG databases, and one `glad.csv` table.


srmsoumya:
Racoon bot works great, added uvicorn to run the fastapi app.

PR #21, opened 2024-11-22
Feature/dockerize rag setup

leothomas:
This PR implements a dockerized setup for the unified RAG agents that @yellowcap put together. 

Notes: 
- This PR implements a dynamically set model, however the OpenAI model doesn't accept the `format="json"` parameter. It must instead be set as follows: 
```python
json_llm = llm.bind(response_format={"type": "json_object"})
ai_msg = json_llm.invoke(
    "Return a JSON object with key 'random_ints' and a value of 10 random ints in [0-99]"
)
```
- Only the chat model is dynamically set, the embeddings model is still hardcoded to use the Ollama nomic embeddings model


frontend/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Use the official Python image from the Docker Hub
+FROM python:3.9-slim
+
+# Set the working directory in the container
+WORKDIR /app
+
+# Copy the requirements file into the container
+COPY requirements.txt .
```
yellowcap, line 8: Why is this not using the uv environment?

zeno/agents/maingraph/models.py
```diff
@@ -0,0 +1,100 @@
...
+        "model_id": "qwen2.5:7b",
+        "model_name": "Ollama Qwen2.5 (7 billion)",
+        "constructor_class": ChatOllama,
+        "json_mode": {"param_name": "format", "param_value": "json"},
+        "additional_params": {
+            "base_url": os.environ.get("OLLAMA_BASE_URL"),
+        },
+    },
+    # {
+    #     "required_env_var": "ANTHROPIC_API_KEY",
```
yellowcap, line 31: Why is this commented out, it would be great to have Claude as an option.

zeno/agents/docfinder/utils/nodes.py
```diff
@@ -45,15 +42,21 @@ class grade(BaseModel):
         input_variables=["context", "question"],
     )
 
+    model_id = config["configurable"].get("model_id")
+    model = ModelFactory().get(model_id)
```
yellowcap, line 46: Not sure how much overhead this is to instanciate every time. Maybe we need to pull these out of the tools and instanciate on module load? Something that does not need change right away but to keep in mind.

docker-compose.yaml
```diff
@@ -0,0 +1,88 @@
+services:
```
yellowcap, line 1: I am getting a port error when running `docker-compose up`. Could be something on my end though, not sure.

```
[+] Running 6/5
 ✔ Network project-zeno_default              Created                                                                                0.0s 
 ✔ Volume "project-zeno_database_data"       Created                                                                                0.0s 
 ✔ Container project-zeno-api-1              Created                                                                                0.1s 
 ✔ Container project-zeno-frontend-1         Created                                                                                0.1s 
 ✔ Container project-zeno-db-1               Created                                                                                0.1s 
 ✔ Container project-zeno-langfuse-server-1  Created                                                                                0.0s 
Attaching to api-1, db-1, frontend-1, langfuse-server-1
Gracefully stopping... (press Ctrl+C again to force)
Error response from daemon: driver failed programming external connectivity on endpoint project-zeno-db-1 (cb3d16eb9595ee17601e03ec769e8faa92ff443fc77e7b1cf2493f2132ec6b56): Bind for 0.0.0.0:5432 failed: port is already allocated
```

yellowcap (commented):
Looks pretty good, nice job at centralizing the model loading @leothomas !

There are a bunch of merge conflicts unfortunately. If these are resolved I think we can go ahead and merge.

zeno/agents/docfinder/utils/nodes.py
leothomas: Good call. I think there are 3 ways to address this: 
- Instantiate the user-specified model and store is as an environment variable, before running the `graph.stream` command. Each tool can then pull the model from that environment variable. I'm not a huge fan of using environment variables in this way. 
- Instantiate the user-specified model and store it in the `config` object passed to the `graph.stream()` method. (I'm not sure what the restrictions are on the values in the config object - they may need to be hashable for example)
- Possibly an easier options is to add a `functools.lru_cache` to the `ModelFactory.get()` method to cache the instantiated model. Since the method output is cached based on the hashed parameters, the cache would maintain one version of the model for JSON mode and one for regular mode for each model

leothomas: As discussed earlier, we'll leave the model instantiation as is for now, as we may want to explore the option of the tools/nodes using different models

zeno/agents/maingraph/models.py
leothomas: Anthropic doesn't have a `format="json"` parameter readily available on the `ChatAnthropic()` model instantiation method. We can force JSON output by using `llm.with_structured_output()` and pass an output definition, but since the various nodes/tools are already using `llm.with_structured_output()` _after_ having fetched the instantiated model from the `ModelFactory`, I was unsure of the potential effects of the double `with_structured_output()` call, so I disabled the Anthropic model class for now

frontend/Dockerfile
leothomas: I just knew the syntax using the requirements file off the top of my head - plus since we're building each service within a docker container, there's no need for the overhead of building a whole python package!

PR #22, opened 2024-11-25
Build docker image and publish to GHCR

sunu:
Building a single image that can be used for both the api and the streamlit app for now.

Images are published to https://github.com/wri/project-zeno/pkgs/container/project-zeno


yellowcap (approved):
Yup, looks good 🦘

PR #23, opened 2024-11-26
Unify agents

leothomas:
Merges the work that @yellowcap, @srmsoumya  and I have been doing to put together a containerized agent with multiple routes and access to tools, such as a fuzzy location matcher, GLAD alterts look up, etc. 

I want to merge the containerized structure of the code base so that future work won't create merge conflicts


sunu (commented):
@leothomas I took a brief look at the changes. Few comments/suggestions:
- I would propose we use a single Dockerfile for the api and streamlit given that `streamlit` is the only extra dependency (and soon we might move away from streamlit to a proper ui). That would make things simpler in terms of image management since we'll build a single image instead of two.
- Lets keep the `uv` based package management. It's a lot faster than `pip`.

I have a docker image for the api that uses `uv` in https://github.com/wri/project-zeno/pull/22. The same image with a different command can be used to run streamlit if we want.

yellowcap (approved):
Looks good, we might reconsider passing a model id to the api, but that we can do later.

Issue #24, opened 2024-11-27
Alerts agent based on GEE LCL layers


yellowcap:
An agent to feed the first alerts workflow. The agent needs to

- [x] Find GADM areas based on bbox or text
- [x] Compute DIST alert statistics for the area
- [x] Create a text description of the result
- [x] https://github.com/wri/project-zeno/issues/34
- [ ] Add human in the loop checkpoints
- [ ] Improve LLM interpretation
- [ ] Allow for buffer around the AOI for "context"
- [ ] Increase number of context / landcover layers
- [ ] Extract dates for each dist alert
- [ ] Confidence level
- [ ] Offer satellite imagery for before/after for reference

Based on LCL layers on GEE.

yellowcap:
Som insight on getting geojson from tools: this is a lot of tokens for llms, but its hard to shield the llm from tool output.

It's not possible to update state from within a tool. We need to find a way to pass larger amounts of data from tools to the frontend without giving it to the LLM for interpretation. We probably want to pass part of the data as-is to the frontend, and part to the LLM for interpretation.

We can inject the state into the tool using [InjectedState](https://langchain-ai.github.io/langgraph/reference/prebuilt/#langgraph.prebuilt.tool_node.InjectedState) , but not update the state (or at least its highly discouraged).

For instance, pass disturbance statistics data to LLM, but send vectorized disturbances directly to frontend. I have not figured out how to solve that, and the vectorized geojson generates quite a lot of tokens.

A related discusison is
https://github.com/langchain-ai/langgraph/discussions/1616

cc @srmsoumya

yellowcap:
Ok this was a bit hard to figure out, but there is a simple solution to this actually! 🫏 

The [`@tool`](https://python.langchain.com/v0.2/api_reference/core/tools/langchain_core.tools.convert.tool.html#tool) decorator has an additional argument: `response_format (Literal['content', 'content_and_artifact'])`.

If this is is et to `content_and_artifact`, any kind of data can be passed to downstream as artifacts, and they will not be passed to the model. We can leverage this to feed data to the user.

yellowcap:
Closing this as the individual items have been ticketed out separately.

PR #28, opened 2024-11-28
Gee agent

yellowcap:
Eessntially solves an initial minimal version of #24 but is still a bit buggy.


yellowcap:
Merging this to test the build while folks are on holidays 😄

Issue #37, opened 2024-12-06
Add more raster layers for analysis


yellowcap:
### Goal
Include additional datasets that can be included in analysis (masking, aggregation) and interpretation. See https://gfw.atlassian.net/wiki/spaces/LCL/pages/edit-v2/1550254128#Datasets

Priority list:
- Global Tree Cover [GEE](https://developers.google.com/earth-engine/datasets/catalog/JRC_GFC2020_subtypes_V0)
  - Aggregate by tree cover type
- Grasslands [GEE](https://global-pasture-watch.projects.earthengine.app/view/ggc-30m)
  - consider disturbances to natural grasslands
- Above ground carbon biomass (Mg/ha) [GEE](https://code.earthengine.google.com/4f57d1f4a659b8abf22a52c83e127f39)
  - highlight alerts that coincide with high biomass areas and could have high potential to release carbon emissions

### Scope
...

### User Stories & Acceptance Criteria
- When I [context], I want to [action], so that I can [outcome]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]

### Future work

PR #38, opened 2024-12-06
Improve context layer tool

yellowcap:
Moves tool into separate folder, and adds choices beyond the NaturalLands layer.

Addresses most of #37 , need to add grasslands and biomass. But since the addition is from a metadata file, its just collecting the descriptions and maybe some thresholds.


srmsoumya (approved):
LGTM.

Note: We can move to vector-db for layers (even though it is just 10) & switch from fiona to gpd in a different PR.

PR #41, opened 2024-12-12
Add notebook for generating embeddings and initial RAG implementation for fetching the appropriate layer

leothomas:
Notes: 
- Embeddings for some of the layers from the Confluence document still need to be generated (see note in IPyNb)
- The RAG uses LanceDB as a vector store, backed directly to S3. Appropriate AWS Credentials/IAM role will be required to access the LanceDB instance
- The vector store search step is now implemented as a retriever - I'm not sure if the LangChain model automatically knows to use the retriever of if the prompt needs to be amended to tell the Agent to use the retriever tool first


yellowcap:
Works locally for some layers. Some others need lookup tables to work, added some more lookup tables but not all.

Changed default query, natural lands works well.

Will merge and we can fix things so that we have working statistics for all context layers.

PR #51, opened 2024-12-18
Add HITL for location

srmsoumya:
Adds HITL for location, human needs to provide feedback on the desired location before it can pull dist alerts.


yellowcap:
💬

Issue #61, opened 2024-12-18
Extract confidence level from DIST alerts


01painadam:
### Goal
The Raw Assets in [GEE](https://code.earthengine.google.com/e5efd1064ec5b258eef195505513064c) contain several bands that could be used as filtering criteria. See [documentation](https://lpdaac.usgs.gov/products/opera_l3_dist-alert-hls_v1v001/) 

NOTE: #64 deals with date filters already

Bands used:
- `VEG-DIST-DATE` date filter
  - d=0: no anomaly detected
  - d>0: day of initial anomaly detection in the last year
- `VEG-ANOM-MAX` alerts are based on a 3 year average measurement for a similar time window. This and is the normalised pixel value difference from that baseline. WRI filters on ≥30% as disturbance less than this are more difficult to corroborate/verify with EO scenes.
- `VEG-DIST-COUNT` Proxy for confidence. The number of detected vegetation loss anomalies observations.
  - 1: No confidence (ignore, typically artefacts or cloud cover)
  - 2-3: Low confidence
  - ≥4: High confidence
- `VEG-DIST-DUR` Vegetation disturbance duration. Shorter term (<30 days) tend to be caused by climate events or wildfires. Persistant events (>70 days) indicate permanent land cover change.
- `VEG-LAST-DATE` date of last assessed observation for a disturbance

We can use these bands and allow users to filter by date, confidence etc.

### User Stories & Acceptance Criteria
- When I [context], I want to [action], so that I can [outcome]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]

### Future work

yellowcap:
We have already implemented the confidence level as an input [here](https://github.com/wri/project-zeno/blob/main/zeno/tools/distalert/dist_alerts_tool.py#L85) but its not yet exposed in the promt side. We can try to add that in by mentioning the parameter in the propmt.

PR #64, opened 2024-12-18
Add temporal filters for distalert

yellowcap:
Adds time filter to distalerts tool.

Also converts units of distalerts to hectares.

Resolves the time filter in #61


01painadam (approved):
LGTM!

PR #84, opened 2025-01-07
Add data-api example notebook for monitoring tools

01painadam:
## Summary

Adds a few examples to illustrate using the data-API OTF endpoints for the monitoring use case. Specifically:
- Querying for ecosystem health: Area, Forest Extent, Loss, Gain 
- Filtering & aggregation: forest type, land cover, year
- Querying for threats: fire and integrated deforestation alerts

### DIFF
--- a/nbs/09_data_api_otf_analysis_examples.ipynb
+++ b/nbs/09_data_api_otf_analysis_examples.ipynb
@@ -1,0 +1,1286 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


yellowcap (approved):
Looks great, thanks @01painadam this will be very useful to develop the agent.

Only small comment would be to import the API Key from the environment. If its set directly in the notebook, there is a danger to check it into the repo by accident.

i.e. instead of `DATA_API_KEY = ''`, use `DATA_API_KEY = os.environ.get("DATA_API_KEY")`

01painadam:
Good point! Fixed and merging

PR #89, opened 2025-01-08
Adds nbk with  examples for styleing and tileing gee asssets

01painadam:
Style some GEE tiles! 
e.g. https://earthengine.googleapis.com/v1/projects/earthengine-legacy/maps/57c9739e7f6e1b5b3e2437fa983d9dc7-e4793c526943aed90a2aff87f0f77d26/tiles/{z}/{x}/{y}

Shows examples for priority layers (I may also add examples for DIST driver classifications too if useful?)

Screenshot 2025-01-08 at 18 56 25


yellowcap (approved):
Really cool, we can create a tile endpoint that to the layer selector RAG as additional metadata. The frontend will then be able to add this to a map very easilty @kamicut .

Issue #62, opened 2024-12-18
[API] Offer satellite imagery for nearest before/after scenes for reference


01painadam:
### Goal
For a given analysis, extract alert dates and allow users to inspect the nearest EO scenes before and after the event to inspect the disturbance visually.

### Scope
Define the limits of the work to be carried out; what is included or not included?

- P1: Sentinel (MVP)
- P2: Planet monthly mosaic (more difficult, tropics only)

### User Stories & Acceptance Criteria
- When I [context], I want to [action], so that I can [outcome]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]

### Future work

yellowcap:
As far as I understood, Planet would be higher value, but if Sentinel is easier to implement, we should go with that one first.

01painadam:
Just a thought @yellowcap @LanesGood - would it be more simple (UI wise) to remove the ability for users to select from multiple dates? Instead we either animate the scenes between the dates or show a before after slider?

LanesGood:
Frontend & design implications discussed in standup:
- STAC tool returns images with dates (most recent five?)
- People can click through time slider
- Or, show thumbnails in filmstrip
    - ideally STAC response includes jpeg or other smaller image
- When I click on a thumbnail, load it behind the geojson of the alert
- Add satellite image to layer panel

yellowcap:
We added a tool for that in #98, but we still need to hook it into the API.

LanesGood:
@kamicut I are have discussed requirements for this "filmstrip" approach, and he's considering what is possible. We also discussed the need for overlaying both satellite imagery and still having access to generated charts, which led to this proposed "tabbed secondary widget" design. This proposed design also allows users to collapse this panel, but does not yet explore the idea of removing this data from the map - this seems out of scope for the prototype phase.

![image](https://github.com/user-attachments/assets/226a4f47-7155-425c-b96f-7ddbb4b3313f)

PR #98, opened 2025-01-10
Initial STAC search tool for sentinel-2

yellowcap:
Initial tool for stac searches. Still needs to be integrated into an agent, this is only the tool. Might be something to add to the new agent setup @srmsoumya .

@kamicut this will return a bunch of STAC items as geosjon FeatureCollection. What is the most useful way for the frontend to get this data? We could pre-create TiTiler endpoints for instance to put on the map. Your input here on what is needed on the frontend to visualize STAC items would be valuable.

Refs #62

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,8 @@ dependencies = [
     "geopandas>=1.0.1",
     "geemap>=0.35.1",
     "langgraph>=0.2.56",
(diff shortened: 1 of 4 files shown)


srmsoumya (approved):
LGTM - we can change what we want to pass to the model later, merging for now.

PR #100, opened 2025-01-10
Add tms url generation to context layer tool

### LINKED ISSUE #90: [API] Add layer tile url with styles
Style data layers on map for DIST alerts plus [priority layers](https://gfw.atlassian.net/wiki/spaces/LCL/pages/1550254128/001.+Improved+Alerting+workflow#Datasets)

See this [notebook](https://github.com/wri/project-zeno/blob/main/nbs/10_gee_tiles.ipynb) for examples

yellowcap:
This adds a dynamically generated TMS url to the context layer output. It also adds the rest of the metadata to the output through an `artifact`. 

Closes #90 

The only thing missing is a better viz parameter styling. This is currently using the default, which will look nice if the default is set. For instance, the NatrualLands one has a palette built in and will look nice. The world cereal will simply be a grayscale, so that one should be changed. We can make this dynamic in a later step.

### DIFF
--- a/tests/test_context_layer_tool.py
+++ b/tests/test_context_layer_tool.py
@@ -4,14 +4,13 @@ from zeno.tools.contextlayer.context_layer_retriever_tool import (
 
 
 def test_context_layer_tool_cereal():
-    result = context_layer_tool.invoke(
-        input={"question": "Summarize disturbance alerts by type of cereal"}
+    msg = context_layer_tool.invoke(
+        {
+            "name": "context-layer-tool",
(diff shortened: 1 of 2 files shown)


01painadam:
LGTM! 

> The only thing missing is a better viz parameter styling.
On this @yellowcap , where should the styling be stored? 

In the [notebook](https://github.com/wri/project-zeno/blob/main/nbs/10_gee_tiles.ipynb) I used the default style definitions for everything except Tree Canopy Height `projects/meta-forest-monitoring-okw37/assets/CanopyHeight` and DIST alerts.

01painadam (approved):

PR #101, opened 2025-01-13
Distalerts drivers

yellowcap:
Creates an `ee.Image` from the available distalert driver layers. This image should then be used in the distalert tools the same way that a "normal" landcover from the `context-layer-tool` works.

The driver layer composition is put into a separate module. The integration with the distalerts tool is rudimentary and needs improvement. But we can do that as part of the HIL refactoring.

### DIFF
--- a/tests/test_dist_agent.py
+++ b/tests/test_dist_agent.py
@@ -1,3 +1,4 @@
+import datetime
 import uuid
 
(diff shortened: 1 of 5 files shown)


srmsoumya (approved):
LGTM.

PR #102, opened 2025-01-13
Fix distalert ID handling, add unit test

yellowcap:
Fix some of the ID handling.

Add unit tests for distalert tool:

- Used GEE to get info from a specific location.
- Patched location finder in unit test, obtaining tiny geom around test point
- Used tests to drill into date and threshold settings to confirm these work
- Verified correct landcover type and drivers are returned


01painadam (approved):
Great stuff! Good to see some human-verified results. LGTM!

PR #113, opened 2025-01-19
Add support for stac-tool to query for satellite imagery

srmsoumya:
Pulls satellite imagery for an AOI - can be used to verify the alerts. Fixed it to E84 catalog & sentinel imagery for MVP.


yellowcap (approved):
🌻 let's go STAC!

PR #120, opened 2025-01-23
Fix gee image type instantiation errors

yellowcap:
Incomplete but addressing the errors on instanciation. This happens using grasslands for instance.


yellowcap (commented):
🎸 looks good thanks @leothomas

PR #122, opened 2025-01-27
Hide json on the streamlit-frontend, fix issue with artifacts

srmsoumya:
- Collapse window for json strings
- Adds formatting for AI response
- try-catch block for artifacts to prevent front-end issues


yellowcap (approved):
🐐 looks good!

PR #145, opened 2025-02-07
WIP: KBA to return insights as cards 

srmsoumya:
KBA to return insights as cards that can be:
- table
- chart
- markdown

TODO:
- [ ] Format the frontend to render data based on types
- [ ] Fix multiple rendering of charts


zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+    @validator("categories")
+    def validate_categories_length(cls, v, values):
+        if "values" in values and len(v) != len(values["values"]):
+            raise ValueError("Categories and values must have the same length")
+        return v
+
+
+class Insight(BaseModel):
+    type: InsightType = Field(..., description="Type of insight visualization")
+    title: constr(min_length=5) = Field(
```
yellowcap, line 51: Why is the constraint necessary here?

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+        elif insight_type == InsightType.CHART and not isinstance(
+            v, (dict, ChartData)
+        ):
+            raise ValueError("Data must be ChartData for chart insights")
+        return v
+
+
+class InsightsResponse(BaseModel):
+    insights: List[Insight] = Field(
+        ...,
```
yellowcap, line 92: I should know this, but what do the `...` actually add here?

zeno/agents/kba/state.py
```diff
@@ -6,12 +6,10 @@
...
 from typing_extensions import TypedDict
 
-from zeno.agents.kba.agent import KBAResponse
-
 
 class KbaState(TypedDict):
     messages: Annotated[Sequence[BaseMessage], add_messages]
     user_persona: str
-    report: KBAResponse
+    kba_within_aoi: str
```
yellowcap, line 13: why not `dict` or even a geojson pydantic object?

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+@tool("kba-insights-tool")
+def kba_insights_tool(question: str, state: Annotated[Dict, InjectedState]):
+    """Find insights relevant to the user query for the Key Biodiversity Areas (KBAs).
+
+    Args:
+        question: The user's question or query
+    """
+    print("kba insights tool")
+    kba_within_aoi = state["kba_within_aoi"]
+    kba_within_aoi = gpd.GeoDataFrame.from_features(json.loads(kba_within_aoi))
```
yellowcap, line 120: If this was a dict or geojson object we would not have to load it from string.

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+        columns.remove("geometry")
+
+    kba_within_aoi_filtered = kba_within_aoi[columns]
+
+    kba_insights_prompt = KBA_INSIGHTS_PROMPT.format(
+        user_persona=user_persona,
+        question=question,
+        column_description=column_description[
+            column_description["column"].isin(columns)
+        ].to_csv(index=False),
```
yellowcap, line 150: Interesting that we pass the data as csv. Kind of :magic_wand: that this works! In the future, we could consider to return the data as artifacts in combination with the insights text.

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+        data=kba_within_aoi_filtered.to_csv(index=False),
+    )
+
+    response = insights_agent.invoke(kba_insights_prompt)
```
yellowcap, line 154: Is there a way to ask for markdown output explicitly? Or is that already happening somewhere?

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+
+    return json.loads(response.json())
```
yellowcap, line 156: What is the benefit of this encode/decode routine?

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+from langgraph.prebuilt import InjectedState
+from pydantic import BaseModel, Field, constr, validator
+
+from zeno.agents.kba.prompts import (
+    KBA_TS_COLUMN_SELECTION_PROMPT,
+    KBA_TS_INSIGHTS_PROMPT,
+)
+
+data_dir = Path("data/kba")
+kba_ts_data = pd.read_parquet(data_dir / "kba_timeseries_data.parquet")
```
yellowcap, line 21: Did you already sync the data into S3?

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @validator("year")
+    def validate_year(cls, v):
+        current_year = datetime.now().year
+        if v > current_year + 50:
+            raise ValueError("Year cannot be more than 50 years in the future")
+        return v
+
+
+class TimeSeriesInsight(BaseModel):
```
yellowcap, line 51: Really nice how much detail there is for analysing the time series! :sparkels:

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+    max_change: Optional[dict] = Field(
+        None,
+        description="Details about the maximum change between consecutive periods",
+    )
+
+    @validator("data")
+    def validate_data_ordering(cls, v):
+        # Ensure years are in chronological order
+        years = [point.year for point in v]
+        if years != sorted(years):
```
yellowcap, line 81: Why not simply sort it if its not already sorted instead of raising an error?

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+            raise ValueError(
+                "Time series points must be in chronological order"
+            )
+
+        # Check for duplicate years
+        if len(years) != len(set(years)):
```
yellowcap, line 87: Also here, duplicates could be dropped instead of error.

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+
+class TimeSeriesResponse(BaseModel):
+    insights: List[TimeSeriesInsight] = Field(
+        ...,
+        min_items=1,
+        description="List of time series insights from the analysis",
+    )
+
+
+class ColumnSelectionOutput(BaseModel):
```
yellowcap, line 135: Consider importing the column selection logic from other agent instead of duplicating?

zeno/agents/kba/graph.py
```diff
@@ -1,22 +1,25 @@
...
+    tools_with_hil,
+    tools_with_hil_names,
+)
+from zeno.agents.kba.prompts import KBA_PROMPT
 from zeno.agents.kba.state import KbaState
 
-column_description = pd.read_csv("data/kba/kba_column_descriptions.csv").to_csv(
-    index=False
-)
+column_description = pd.read_csv(
```
yellowcap, line 20: Centralize the loading of this? I think that happens in several places.

zeno/agents/kba/tool_kba_timeseries.py
```diff
@@ -0,0 +1,192 @@
...
+        description="List of column names relevant to the user query based on knowledge base, user persona, and user query",
+    )
+
+
+column_selection_agent = haiku.with_structured_output(ColumnSelectionOutput)
+time_series_agent = haiku.with_structured_output(TimeSeriesResponse)
+
+
+@tool("kba-timeseries-tool")
+def kba_timeseries_tool(question: str, state: Annotated[Dict, InjectedState]):
```
yellowcap, line 147: How does it know if there are time series and for which variables? I.e. how does the graph decide to call this tool? I guess this tool only make sense if the columns selected have time series data. Is there a way to control for this?

yellowcap (commented):
Looks really nice! Made some comments, will try this out in a moment.

zeno/agents/kba/tool_kba_data.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Args:
+        name: Name of the location to search for. Can be a city, region, or country name.
+        gadm_id: GADM ID of the location to search for.
+        gadm_level: GADM level of the location to search for.
+    """
+    print("kba data tool")
+    aoi = get_aoi(gadm_id=gadm_id, gadm_level=gadm_level)
+
+    kba_within_aoi = kba[kba.geometry.within(aoi)]
```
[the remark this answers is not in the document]
srmsoumya: ah yes, didn't think about it - will test that out for different regions & update in the next PR

zeno/agents/kba/tool_kba_insights.py
```diff
@@ -0,0 +1,156 @@
...
+
+class ChartData(BaseModel):
+    categories: List[Union[str, int, float]] = Field(
+        ..., description="Categories for the chart axis"
+    )
+    values: List[Union[int, float]] = Field(
+        ..., description="Numerical values corresponding to categories"
+    )
+
+    @validator("categories")
```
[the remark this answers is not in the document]
srmsoumya: yes, changing this

zeno/agents/kba/tool_kba_insights.py
srmsoumya: just ensuring the LLM sends good titles that are long enough to interpret

zeno/agents/kba/tool_kba_insights.py
srmsoumya: used in pydantic docs, but can be removed, they don't add anything special.

zeno/agents/kba/tool_kba_insights.py
srmsoumya: Can be done by explicit prompting

zeno/agents/kba/tool_kba_timeseries.py
srmsoumya: umm nope, need to check my `aws` access again :-(

srmsoumya:
a lot of this is just me scrambling through to get the code base working - there are still a few edge cases I need to fix.

zeno/agents/kba/tool_kba_timeseries.py
srmsoumya: There is a prompt for this & the llm knows when to call this tool when the user requests for specific time-series workflows.

yellowcap (approved):
Just tried the streamlit, looks great! 🌺

PR #156, opened 2025-03-19
This will allow using zeno from anywhere in the virtualenv.

yellowcap:
This removes pain when it comes to pythonpaths. Basically uv will install the package in a n "editable" mode into the venv and 🎉 no more path problems.


srmsoumya (approved):
This is super useful.

PR #158, opened 2025-04-21
Authentication demo

leothomas:
This PR is a small demo of authenticating against the RW API. Run it using docker-compose from the project root

The demo redirects the user to the RW API's auth page, with a callback URL to the application. The user can login, either directly against the RW data API, or with 3rd party auth (Google, Facebook, Apple). In all three cases, the user will returned to the application url with a `token` query parameter containing the JWT. The streamlit application fetches the token from the query parameter and uses it to fetch the logged in user's info (when running the demo, you'll briefly see the token flash in the URL after being redirected to the application). 

I think that it's technically not a best practice to pass the token through the URL (because man-in-the-middle attacks can intercept the token in the URL and then make any sort of authenticated requests on the user's behalf). I know that the RW API offers browser cookie based authentication (which also enables revoking the browser cookie) but I'm not seeing any cookies set in the browser when being redirected to the Zeno frontend. Either way, cookie based authentication wouldn't work with Streamlit (since streamlit runs a python server to power the frontend, which can't access HTTPOnly cookies) but would, in theory, work with a REACT frontend - we should further investigate this. 

Now that we can fully make use of the RW hosted auth functionalities (including password reset, etc) I think we can skip using `fastapi-users` (which integrates directly with the auth provider, such as Okta) and simply set up a user's table, with a UUID primary key, corresponding the UUID from the RW auth endpoint (see demo). 

The last remaining question to answer is if we want to use an ORM, or just build the run the SQL queries directly using `pyscopg`

https://github.com/user-attachments/assets/382b2925-44b0-4dd8-b4aa-bbbefadf2cdf


batpad:
@leothomas this is super!

I made a minimal example showing how one can do this purely client-side for the frontend: https://github.com/batpad/wri-auth-demo

> I think that it's technically not a best practice to pass the token through the URL (because man-in-the-middle attacks can intercept the token in the URL and then make any sort of authenticated requests on the user's behalf).

If one is using https, man-in-the-middle attacks should not be a concern. I think the concern with passing authentication details in the URL is more that there maybe places where one is logging URLs, etc. and ensuring that tokens are not being logged . These are things we should have control over though. I'm not too concerned about handling the token in the query parameter of the callback, but happy to defer if anyone has strong objections and tangible reasons for why we shouldn't do this.

> but would, in theory, work with a REACT frontend - we should further investigate this.

I think we can just use the same redirect approach and grab the token on the frontend - I made a small PoC of this: https://github.com/batpad/wri-auth-demo

> Now that we can fully make use of the RW hosted auth functionalities (including password reset, etc) I think we can skip using fastapi-users (which integrates directly with the auth provider, such as Okta) and simply set up a user's table, with a UUID primary key, corresponding the UUID from the RW auth endpoint (see demo).

That sounds great to me!

> The last remaining question to answer is if we want to use an ORM, or just build the run the SQL queries directly using pyscopg

My preference would be to use an ORM - if we choose not to use an ORM, at the very least, I would want to have a very solid way to handle database migrations. I do think it's nice to use a mature ORM framework that would handle that for us.

leothomas:
Awesome, so we'll move forward with the URL token based authentication. I'm on board with ORM + Database migrations manager. I've worked with SQLAlchemy + Alembic in the past - it's very comprehensive and offers fine grained control, but can sometimes be a little verbose and cumbersome. ChatGPT suggests [Tortoise](https://tortoise.github.io/) as an alternative ORM, supposed to have slightly less advanced features (eg: missing composite keys) but much more simple, Django-style syntax (not exactly sure what that means) and comes with Aerich as a migration tool

Any other options we should consider for our ORM + Database Migrations?

batpad:
@leothomas this looks really good to me. Can we add tests for the fetch_user / user authentication functionality?

The token authentication and fetching and caching of user data, etc. looks good to me.

It seems to me like adding Threads and saving conversations might take a little work to:

 - Work out the exact relationship between Users, Projects, Threads, related to https://gfw.atlassian.net/browse/PZB-84
 - Work out the exact schema and what we want to store for conversations

I think my suggestion might be to limit the scope of this PR to adding the user authentication stuff, but then have a bit more architecture / discussion around the modelling of Projects / Threads and saving conversations.

I would suggest either:

 - If conversation saving is roughly working, just merge this roughly as is, and then ticket out more in detail the handling of projects and conversations and make those changes in a separate PR with the required db migrations
 - Remove the Threads model and conversation saving from this PR for now, merge the auth work, and implement Threads / Projects as a separate PR

I'm fine with leaving this in one PR, but it seems like it might get a bit much, and I do think we might need to do a bit of discussion around modelling of projects vs threads, etc.

db/models.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class UserModel(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    id: str
+    name: str
+    email: str
+    created_at: datetime
+    updated_at: datetime
+    threads: list[ThreadModel] = []
```
yellowcap, line 51: One question, why is this necessary? Isn't the `user_id` foreign key on the Threads table enough?

docker-compose.yaml
```diff
@@ -12,12 +12,18 @@ services:
       - ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY}
       - OPENAI_API_KEY=${OPENAI_API_KEY}
       - GEE_SERVICE_ACCOUNT_PATH=${GEE_SERVICE_ACCOUNT_PATH}
+      - WATCHFILES_FORCE_POLLING=1
+      - DATABASE_URL=postgresql+psycopg2://postgres:postgres@db:5432/zeno-data
```
yellowcap, line 16: Is this on the same PG instance as the langfuse tracing? How much separation are we thinking about on the db instance level?

frontend/pages/5_🐊_Investi_Gator.py
```diff
@@ -55,6 +56,41 @@ def reset_state():
     """
     )
 
+    if not st.session_state.get("token"):
```
yellowcap, line 59: Any way we can centralize this streamlit part in a separate file and import it instead of dupliating? Not sure how that would work in streamlit, not worth spending a lot of time on I think, just a thought.

db/models.py
```diff
@@ -0,0 +1,51 @@
...
+class ThreadOrm(Base):
+    __tablename__ = "threads"
+
+    id = Column(String, primary_key=True, unique=True, nullable=False)
+    user_id = Column(String, ForeignKey("users.id"), nullable=False)
+    created_at = Column(DateTime, nullable=False, default=datetime.now())
+    updated_at = Column(
+        DateTime, nullable=False, default=datetime.now(), onupdate=datetime.now()
+    )
+    content = Column(JSONB, nullable=False)
```
yellowcap, line 31: Is this intened to store the entire thread history? This could become pretty big, because it will contain all the output, including tools etc. Maybe we want a thread model and then a separate messages table where we split the individual messages into one row each. This would also make appending history easier.

yellowcap (commented):
Thanks  @leothomas this looks pretty good already.

My main comment is about the thread data table. So its a bit tangential from this PR. I think @kamicut should weigh in on how we store the thread data. 

So we could consider removing the thread model from this PR and putting it into another one so this PR becomes more focused.

For the auth part it looks good but I have not done test drive.

db/models.py
yellowcap: cc @kamicut how are you thinking about loading the thread data?

db/models.py
leothomas: The foreign key in the `UserOrm` will load the threads for a User, but the lines you've highlighted are for the pydantic UserModel - since the ORM only loads the relationship when accessed, we need to declare it as a field on the model. I'm following this logic: https://docs.pydantic.dev/latest/examples/orms/#sqlalchemy to integrate the pydantic models with the SQLAlchemy ORM models - there might be a neater way to do this

docker-compose.yaml
leothomas: Yes - it's one the same PG instance, but in a separate database. My instinct is to say one PG instance for both but we can run that by Bitner to see if it's a bad idea in production (it definitely works for the demo)

frontend/pages/5_🐊_Investi_Gator.py
leothomas: Yeah - repeating the authentication logic throughout the app is definitely very hacky, I was running into issues where navigating from one page to the other would sometimes refresh the state and lose the token. Can definitely look into centralizing the code!

db/models.py
leothomas: There are a couple discussion threads in slack on this subject - but I agree that the conversations will become very big. A couple ideas: 
- We store conversation metadata in the postgres table, and then the actual conversation content in S3 
- We use a NoSQL database (DynamoDB, ElasticSearch) for the whole thing
- (as you suggested) We create a `messages` table where we store each message as a separate entry
- We store the conversation in the `threads` table but store any geometries separately, with references in the `threads.content` column 

Since there's still a lot of conversations happening around the Users<>Projects<>Threads structure I haven't pushed on this any more

leothomas:
Hey folks - I think we can go ahead and merge, I've recorded a quick demo of the user + thread data management [here](https://drive.google.com/file/d/17hDPDdCgEwclV1jLj9WzogG12MRdaGhh/view?usp=sharing) since I won't be online for our tagup with the Zeno folks

leothomas:
Note: I've also added the emails allowlist feature, it works off of a comma separated string of accepted emails. The emails allow list is also cached in the user validation functionality so that if the allowlist is updated, the user will be allowed to make requests against the API before that 24 cache invalidation period

api.py
```diff
@@ -29,6 +29,8 @@
 engine = create_engine(DATABASE_URL)
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
+EMAILS_ALLOWLIST = os.environ.get("EMAIL_ALLOWLIST", "").split(",")
```
batpad, line 32: @leothomas I think what we want here is to be able to whitelist domains and not just specific email addresses.

cc @kamicut can you confirm that is what the frontend is doing?

batpad:
@leothomas is this PR ready to merge? I see it still marked as Draft.

batpad:
There's a few things we should add here:

 - Tests
 - Better documentation for the alembic setup to run locally

But I'm going to go ahead and merge this, get this deployed with the database connection bits and get back to those.

PR #159, opened 2025-05-05
Activating langfuse callback in all agents

yellowcap:
The langfuse callback was not in all configs for the agents.

### DIFF
--- a/api.py
+++ b/api.py
@@ -4,24 +4,24 @@ import os
 import uuid
 from typing import Annotated, Optional
 
-import requests
 import cachetools
+import requests
 from dotenv import load_dotenv
-from sqlalchemy import create_engine
-from sqlalchemy.orm import sessionmaker
 from elevenlabs.client import ElevenLabs
-from fastapi import Body, FastAPI, Header, HTTPException, status, Depends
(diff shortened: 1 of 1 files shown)


batpad:
@yellowcap sorry - I went ahead and merged @leothomas's auth PR and now there's some conflicts here - would you be able to rebase / merge main and fix?

api.py
```diff
@@ -473,11 +480,7 @@ def event_stream_gfw_data_api(
     else:
         raise ValueError(f"Invalid query type from frontend: {query_type}")
 
-    # Stores the stream to later write to the database
-    messages_store = []
```
yellowcap, line 477: @batpad the linter was not happy about this line, as this var was never used. I guess the `messages_to_store` one below is the real variable being used.

yellowcap:
Looks like this PR did not use the pre-commit linters, so the rebase introduced a bunch of changes here that are purely from the linters. Hope that is ok.

api.py
batpad: Oops! So sorry @yellowcap - getting used to the flow of this code-base - I did not run the linter locally. Thank you so much for fixing these.

PR #161, opened 2025-05-06
add workflow to build and push image for db container

batpad:
Refs https://github.com/wri/project-zeno-deploy/issues/16

To deploy the latest changes, we need to publish the new `db` container to pull and run it in the cluster.

This PR adds a Github workflow to publish the db image to the same ECR we use for the main image.

I'm not super happy with this as this will mean two different commit hashes that we need to keep track of between `project-zeno` and `project-zeno-deploy`. I'll work with @sunu to outline a plan to do this a bit better, but I think this is the most straightforward way to deploy these changes right now and unblock / move ahead.

cc @sunu @yellowcap

### DIFF
--- a/.github/workflows/docker-build-db.yml
+++ b/.github/workflows/docker-build-db.yml
@@ -4,10 +4,8 @@ on:
   push:
     branches:
       - main
-      - docker-build-publish
-      - publish-docker-db
-    # paths:
-    #   - 'db/**'
+    paths:
+      - 'db/**'
 
 env:
   REGISTRY: public.ecr.aws/b7u8b0a6
--- a/.github/workflows/docker-build-db.yml
+++ b/.github/workflows/docker-build-db.yml
@@ -1,0 +1,46 @@
+name: Build and Publish DB Docker Image
+
+on:
+  push:
+    branches:
+      - main
+      - docker-build-publish
+      - publish-docker-db
+    paths:
+      - 'db/**'
+
+env:
+  REGISTRY: public.ecr.aws/b7u8b0a6
+  IMAGE_NAME: project-zeno/zeno-db
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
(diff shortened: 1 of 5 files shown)


batpad:
Also wrapped around the `init-multiple-dbs.sh` script to ensure that's run before running migrations so the db gets created the first time this is run. cc @leothomas

batpad:
@sunu I made some more changes to:
 - Create the ECR repository if it does not exist in the Action
 - Consolidating the env vars passed to the db container to not have to pass in DATABASE_URL and things like POSTGRES_HOST separately, so now it just needs POSTGRES_HOST, POSTGRES_USER, POSTGRES_PASSWORD as separate env vars and the DATABASE_URL is constructed within the script

I'll test with the image built from this branch.. If that works, let me know if you have any suggestions to improve or if I can merge this.

db/entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+export DATABASE_URL="postgresql+psycopg2://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:5432/${APP_DB}"
+./init-multiple-dbs.sh && \
```
sunu, line 3: Should we merge the init-multiple-dbs script into the entrypoint script?

.github/workflows/docker-build-db.yml
```diff
@@ -0,0 +1,52 @@
+name: Build and Publish DB Docker Image
+
+on:
+  push:
+    branches:
+      - main
+      - docker-build-publish
+      - publish-docker-db
```
sunu, line 8: Just commenting as a reminder that we might want to remove the extra branches before merging

.github/workflows/docker-build-db.yml
sunu, line 1: Maybe we could add this as a separate step in the current workflow, but I think it's fine for now since we'll be consolidating everything into a single image soon anyway.

sunu (approved):
Added some minor comments, but overall, this looks good to me!

batpad:
@sunu I commented out the special-cased branches. Also commented back in the line to only build if changes have been made to the `db/` folder to avoid too many unnecessary builds - not sure if you feel that makes sense to have or just let it build with all pushes to main?

sunu:
Yeah, I think it makes sense to trigger the build only when there are changes in the `db/` directory.

PR #168, opened 2025-05-07
Add instructions to setup db locally.

srmsoumya:
- Add alembic, sqlalchemy as dependency
- Instructions to setup db locally
- `app.py` now uses .env vars


batpad:
Thanks so much for this @srmsoumya !

PR #173, opened 2025-05-14
refactor: add .env file loading status check in frontend app

Quantisan:
Bug occurs when running `uv run streamlit run frontend/app.py` from the root directory per README instructions. The `.env` file path is ambiguous, so I added a check in the Streamlit app to verify the correct location and provide suggestion to fix when needed.

### DIFF
--- a/frontend/app.py
+++ b/frontend/app.py
@@ -3,13 +3,25 @@ import os
 import requests
 from dotenv import load_dotenv
 
-_ = load_dotenv("../.env")
-
-API_BASE_URL = os.environ["API_BASE_URL"]
-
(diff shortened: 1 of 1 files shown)


batpad:
@leothomas or @srmsoumya would you mind having a look at and approving this? It seems to make sense to me but am not so familiar with the Streamlit app.

Thanks @Quantisan !

yellowcap:
Same as the other PR, to avoid solving merge conflicts through the gofentris fork, I took the relevant changes and added them here https://github.com/wri/project-zeno/commit/de8746aa12fd3d5749fece7673dc03ca5aecae89

PR #175, opened 2025-05-16
Dockerized STAC ingestion

yellowcap:
Dockerized ingestion of data into PG STAC. Works for DIST alerts and Natural Lands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -177,4 +177,3 @@ data/
 
 # Custom zeno
 ee-zeno-service-account.json
-.pgdata/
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -174,4 +176,5 @@ data/
 .vscode
 
 # Custom zeno
-ee-zeno-service-account.json
\ No newline at end of file
+ee-zeno-service-account.json
+.pgdata/
```
batpad, line 180: Minor: Seems like .pgdata/ got added twice to gitignore?

yellowcap:
This should go into a separate zeno-data repository. But will add here for now.

PR #180, opened 2025-05-29
Refactor zonal stats as tool for node in single agent

lillythomas:
Refactoring https://github.com/wri/project-zeno/blob/main/nbs/13_cog_sights.ipynb into a callable tool function for integration with the unified agent.


yellowcap:
@lillythomas I moved your file into the `stac` directory for future reference and going to merge. We are probably not going to build this kind of data analysis tools, stats will be handled by a WRI internal team.

PR #181, opened 2025-05-29
Api tests

batpad:
This is a first pass at adding some scaffolding to run API tests that involve the database, and adds tests for user authentication.

This:

 - Makes some small changes to the user cache and reading of the DOMAINS_ALLOWLIST var to fix small bugs and work well with passing in mocks for tests
 - Adds a conftest.py to do db creation and teardown
 - Creates a separate tests/api/ folder for these kinds of more deterministic API tests (as opposed to the agent tests) with a test_auth.py, that will grow to test the CRUD API end-points, etc

TODO: I'll wire up CI to run these tests as part of CI - I'd really like things like testing our auth and basic db CRUD functions to have tests that run in CI

I have not really written tests for a FastAPI application before, and some of the import stuff got a little weird - @leothomas I might appreciate talking that through with you. My understanding is we might want to change the application structure a little bit to make it work better as a python module, but will run it by you.

You should be able to run the tests added here with `pytest tests/api` and you should see 5 tests pass.

### DIFF
--- a/api.py
+++ b/api.py
@@ -30,7 +30,12 @@ DATABASE_URL = os.environ["DATABASE_URL"]
 engine = create_engine(DATABASE_URL)
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
-DOMAINS_ALLOWLIST = os.environ.get("DOMAINS_ALLOWLIST", "").split(",")
+def get_domains_allowlist() -> tuple[str, ...]:
+    """Get allowed domains as a tuple for immutability."""
+    domains = os.environ.get("DOMAINS_ALLOWLIST", "").split(",")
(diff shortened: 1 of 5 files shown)


yellowcap:
@batpad after our conversation yesterday I think we can close this as out of date? Or is there parts of this that we could merge?

batpad:
@yellowcap can we keep this open and in draft, please? This is easy for me to rebase, and esp. with the latest changes removing the GEE dependency, I might be able to unblock to get this working in CI, in which case I would like to merge this.

The tests are pretty isolated from the rest of the code-base, and most of this should all still work even with the refactors in `main`.

geohacker (approved):

PR #183, opened 2025-06-05
Zonal stats dist

yellowcap:
This adds generalized dist alert function in a new file. The function computes zonal statistics on DIST alerts by date for a geometry, with an optional contextual layer. The contextual layer is assumed to be a discrete integer layer.

The intention is to add this as one of the "branches" of the zeno statistics tool. If DIST alerts on-the-fly computation is needed, call this function.

We can further generalize this by allowing the primary layer to not be dist alerts. But the custom date filtering and grouping might make this hard.

```python
# Read the geopackage containing admin boundaries
gadm = gpd.read_file("gadm_410-levels.gpkg", layer="ADM_3")

data = compute_dist_alert_stats_for_area(
    gadm.iloc[10324]["geometry"], "natural-lands-map-v1-1"
)
print(data)
data = compute_dist_alert_stats_for_area(gadm.iloc[10324]["geometry"])
print(data)
```


lillythomas:
@yellowcap where can I access `gadm_410-levels.gpkg`? I don't see it in `s3://zeno-static-data/`. Also, I have my credentials exported for the internal WRI account but I still am not able to access `s3://gfw-data-lake/umd_glad_dist_alerts/v20250329/raster/epsg-4326/cog/default.tif` did the permissions change for that?

yellowcap:
I am merging this to keep the function as reference in the future. We are no longer going to build our own stats retriaval tools, that will be handled by another team from WRI.

PR #184, opened 2025-06-05
Add overview generator script

yellowcap:
A utility to create a global overview layer from a tiled STAC collection for visualization purposes.

This is hard coded against the Natural Lands STAC collection but should be easy to generalize.


yellowcap:
This is the result. Natural Lands is a beautiful product!

image

Issue #165, opened 2025-05-06
Refactor: split up api.py


batpad:
api.py is getting a big and unwieldy. Let's move it to an `api/` folder and separate it into logical components.

Need to think a bit about the best way to do this / how to split it up.

@leothomas this can wait until you're back and we can chat about it a bit.

Issue #164, opened 2025-05-06
Refactor: move models.py out of db folder


batpad:
Let's keep the db folder just for `alembic` stuff - perhaps renamed to `alembic` - it should hold just the migrations related things.

Let's move the models to a `models/` folder, also so that we can then separate things neatly like `models/users.py`, `models/projects.py`, etc. to hold individual models which will keep things cleaner as our model definitions grow.

cc @leothomas (again, sorry for not catching this in PR - getting a bit used to the FastAPI structure, and caught some of these things as I worked through the deployment)

sunu:
@batpad +1 to this! A couple of extra thoughts we touched on briefly but didn’t dive into:

It’d be great to split out the SQLAlchemy models and Pydantic models into their own dedicated modules.

Also, we should make sure to enable [orm_mode](https://docs.pydantic.dev/1.10/usage/models/#orm-mode-aka-arbitrary-class-instances) on relevant Pydantic models -- so we can go directly from SQLAlchemy objects to Pydantic without having to manually convert to a dict.

I’d suggest all of this lives under the zeno/ directory as part of the main Zeno Python module. Something like:

```
zeno/
  - agents/
  - models/     # SQLAlchemy models
  - migrations/ 
  - api/
  - schemas/    # Pydantic models/schemas
```

The migrations directory won't be a full Python module, but I do like keeping it near the db models for clarity. Personally, I slightly prefer naming it for what it does (migrations) rather than tying it to the tool we’re using (alembic).

batpad:
Thanks for this clear guidance, @sunu ! That all sounds good to me. I'll work through the implementation.

PR #186, opened 2025-06-12
Feature/thread persistence

leothomas:
## Pull request type

Please check the type of change your PR introduces:
- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the new behavior?
This PR introduces 2 main changes: 

- Drop support for the distributed agents (Earthy Eagle, Keeper Koala, etc) in favor a single unified agent (UniGuana). This includes: 
  - Refactoring the backend structure to host all tools/agent definitions under `/src`
  - Updating the streamlit frontend to only interact with the `/stream/chat` endpoint
- Adds thread persistence using the Postgres Langraph Checkpointer to the graph 

Related issues: 
- https://github.com/wri/project-zeno/issues/165
- https://github.com/wri/project-zeno/issues/164

## Does this introduce a breaking change?

- [X] Yes
- [ ] No

Note: this drops support for the existing "distributed agents". I've tried to keep the same docker build process to create the API image, but with the update repo structure, we should test the deployment in a dev environment first.


batpad:
@leothomas this looks fantastic!

Happy to just try testing this on staging soon, once @srmsoumya and @yellowcap have had a chance to take a look. Thanks again, this is looking really super!

Dockerfile
```diff
@@ -18,14 +18,14 @@ RUN sh /uv-installer.sh && rm /uv-installer.sh
 # Ensure the installed binary is on the `PATH`
 ENV PATH="/root/.local/bin/:$PATH"
 
-# Copy the project into the image
-ADD . /app
+## Copy the project into the image
+ADD ./pyproject.toml ./uv.lock ./README.md ./ee-zeno-service-account.json /app/
```
sunu, line 22: If `ee-zeno-service-account.json` is a secret file, we should not copy it into the docker file -- it should be mounted as a volume instead.

db/alembic/versions/b2c5d0a31a8b_add_langgraph_persistence_tables.py
```diff
@@ -0,0 +1,160 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = "b2c5d0a31a8b"
+down_revision: Union[str, None] = "fa1922057e80"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
sunu, line 22: If I understand correctly, because we’re managing these tables manually, we’ll need to add new migrations whenever LangGraph is upgraded and its checkpoint schema changes. It would be helpful to document this or file a ticket to ensure we don’t overlook it in future updates.

db/Dockerfile
```diff
@@ -12,8 +12,7 @@ RUN pip install --no-cache-dir alembic psycopg2-binary pydantic
 # Copy migration scripts and models
```
sunu, line 12: Should we switch this to psycopg 3 too?

db/entrypoint.sh
```diff
@@ -1,4 +1,2 @@
-#!/bin/bash
-export DATABASE_URL="postgresql+psycopg2://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:5432/${APP_DB}"
-./init-multiple-dbs.sh && \
-alembic upgrade head
+#!/bin/sh
+DATABASE_URL="postgresql+psycopg2://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:5432/${APP_DB}" alembic upgrade head
```
sunu, line 2: Should we switch this to psycopg 3 too?

docker-compose.yaml
```diff
@@ -35,18 +33,24 @@ services:
...
-        condition: service_healthy
+      migrate:
+        condition: service_completed_successfully
     ports:
       - "3000:3000"
-    #env_file:
-    #  - .env
+    healthcheck:
+      # Uses wget because curl is not available in the langfuse image
+      test: ["CMD", "wget", "-qO-", "http://172.20.0.3:3000/api/public/ready"]
```
sunu, line 48: this internal ip might not be stable. Should we use localhost or the service name instead?
```suggestion
      test: ["CMD", "wget", "-qO-", "http://localhost:3000/api/public/ready"]
```

yellowcap (approved):
Great last update on the streaming @leothomas . After talking to @srmsoumya we think we should just move forward and merge this. He is ok with the changes and like this @batpad can start deploying this.

Dockerfile
yellowcap: I think we can get rid of that with the new agents, we are no longer using GEE.

PR #188, opened 2025-06-18
Refactor Zeno - Simplify tools to pick an aoi, select dataset & run analysis.

srmsoumya:
Refactor zeno by simplifying the tool calls. 

We have 4 tools now:
1. `pick-aoi` - Selects an AOI & optionally returns any subregion if requested. This works for GADM level 0-5 & KBA/WDPA/Landmarks
2. `pick-dataset` - RAG setup to filter datasets & then selects a dataset & optionally any contexual layer + threshold, dateranges
3. `pull-data` - This is hacky right now - taken most of the code from `investigator` agent, currently only works with `tcl` - tree cover loss data. Will have to update to handle new data streams coming along
4. `generate-insights` - Simple insight generator for now, provides a title, description & table

PR also adds a `RAG` app - `rag/app.py`, this can be used to compare different rag approaches to filter what works best for us - currently, we have vector based rag for nomic & openai embeddings & colbert with lighton's modernbert.


yellowcap (approved):
Merged some of the upstream work and reactivated the postgres checkpointer. 

- Running the agent works fine on my machine
- But I did test it without the postgres checkpointer, so something to watch out for in deployment

We now assume that the data is synced to the nodes on startup. The command to sync the new parquet files is[ documented here](https://github.com/wri/project-zeno/blob/44956c3aaebe78360ee46f6878c2e80519a7c2a3/README.md#getting-started) and reads as

```bash
aws s3 sync s3://zeno-static-data/unified-agent-data/ data/
```

@batpad and @sunu please move this forward as soon as poosible after the langfuse v3 upgrade.

batpad:
@sunu would you have some time to help with figuring out the best way to get this s3 sync to run before the container starts up?

PR #202, opened 2025-07-07
Update insights to return charts

srmsoumya:
Add support to return insights along with charts data in `rechart.js` format. Currently, support simple chart types like: line, bar, pie, area, scatter, table.


src/tools/generate_insights.py
```diff
@@ -39,139 +63,206 @@ class Insight(BaseModel):
...
-        logger.warning("No raw data found to generate insights from.")
+    if not state or "raw_data" not in state:
+        error_msg = "No raw data available in state. Please pull data first."
+        logger.error(error_msg)
         return Command(
             update={
                 "messages": [
                     ToolMessage(
-                        content="No raw data found to generate insights from.",
+                        content=error_msg,
```
yellowcap, line 131: This is the pattern that we need to handle better, as it is hard to catch  on the frontend. Let's discuss how to flag these differntly.

src/tools/generate_insights.py
```diff
@@ -39,139 +63,206 @@ class Insight(BaseModel):
...
+        )
+
+    except Exception as e:
+        error_msg = f"Error generating insights: {str(e)}"
+        logger.error(error_msg, exc_info=True)
+        return Command(
+            update={
+                "messages": [
+                    ToolMessage(
+                        content=error_msg,
```
yellowcap, line 215: Same here, if this is the generate-insights response it has to be caught separately by the frontend.

src/tools/generate_insights.py
```diff
@@ -39,139 +63,206 @@ class Insight(BaseModel):
...
 )
 
 
-@tool("generate-insights")
+@tool
 def generate_insights(
     query: str,
-    aoi: str,
-    dataset: str,
+    aoi: str = "",
```
yellowcap: `aoi` and `dataset` are not used in this function.

yellowcap (approved):
This worked for me locally. I removed some unused vars and linted the code. Merging this now to keep things moving. @srmsoumya or @batpad when you come online please deploy and test.

The error catching here has been flaged by @kamicut as difficult to handle on the frontend because it breaks the expected pattern. Let's find a way to make this different so that its easy to catch on the frontend.

PR #213, opened 2025-07-14
Draft raster visualization tool

lillythomas:
Draft for the visualization tool, currently focused on generating a composite of Sentinel-2 true color imagery for the AOI returned from the `pick_aoi` tool and parsed daterange.

### DIFF
--- a/src/agents/agents.py
+++ b/src/agents/agents.py
@@ -1,5 +1,6 @@
 import os
 from datetime import datetime
+from typing import Optional
 
 from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
 from langgraph.graph.state import CompiledStateGraph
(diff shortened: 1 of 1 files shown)


batpad:
@yellowcap @lillythomas is this still relevant? If yes, can we figure out fixing merge conflicts and merging this in?

yellowcap:
This is definitively still relevant, but needs to be coordinated with the frontend, otherwise if this tool is called the frontend might not know what to do with the output and break. I.e. this implies a new additional schema output type.

We should only merge this if we have time to adapt the frontend to show the layers. Probably would need adaption to async approach as well.

lillythomas:
I've changed the tool to return a URL in slippy map format. An example from the tool `http://127.0.0.1:8080/collections/sentinel-2-l2a/WebMercatorQuad/tiles/WebMercatorQuad/{z}/{x}/{y}.png?assets=red&assets=green&assets=blue&rescale=0,3000&datetime=2025-08-01/2025-08-07&minzoom=11` is correctly loading when testing in QGIS (zoomed in on the Sunset District in San Francisco here):

Screenshot 2025-09-13 at 10 10 40 AM

yellowcap:
Will close this because its getting out of date and we don't have it on an immediate roadmap. We can re-visit the appraoch when this comes up again.

Issue #217, opened 2025-07-16
Return errors without trying to raise HTTPException


geohacker:
For example:

https://github.com/wri/project-zeno/blob/main/src/api/app.py#L190

This fails if the exception occurs after the connection was started (with a 200). So we should try for this https://github.com/wri/project-zeno/blob/main/src/api/app.py#L181-L188 and during an exception yeild the pack with an errors object with the message.

@batpad @srmsoumya

PR #223, opened 2025-07-17
Improve Streaming Error handling in chat API

srmsoumya:
When errors occurred during streaming responses, `HTTPException` was raised mid-stream, making it difficult for the frontend to handle gracefully.

**Fix**
- Send errors as stream events with node: "error" instead of raising exceptions
- Include error details: message, error_type, and type for better frontend handling
- Distinguish between fatal (stream stops) and non-fatal errors (stream continues)

Addresses Issue #217 

cc @kamicut

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -167,26 +167,56 @@ def stream_chat(
             )
         )
 
-    try:
-        state_updates["messages"] = messages
-        state_updates["user_persona"] = user_persona
+    state_updates["messages"] = messages
+    state_updates["user_persona"] = user_persona
 
+    try:
         stream = zeno.stream(
             state_updates,
             config=config,
             stream_mode="updates",
             subgraphs=False,
         )
+        
         for update in stream:
-            node = next(iter(update.keys()))
-            yield pack(
-                {
-                    "node": node,
-                    "update": dumps(update[node]),
-                }
-            )
+            try:
+                node = next(iter(update.keys()))
+                yield pack(
+                    {
+                        "node": node,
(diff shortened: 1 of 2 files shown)


src/api/app.py
```diff
@@ -167,26 +167,56 @@ def stream_chat(
...
+                    }
+                )
+            except Exception as e:
+                # Send error as a stream event instead of raising
+                yield pack(
+                    {
+                        "node": "error",
+                        "update": dumps({
+                            "error": True,
+                            "message": str(e),  # String representation of the error
```
yellowcap, line 197: This does not seem like a good approach for production, we should log the error instead and only return a simple error message instead of an error dump.

src/api/app.py
```diff
@@ -167,26 +167,56 @@ def stream_chat(
...
-        raise HTTPException(status_code=500, detail=str(e))
+        # Initial stream setup error - send as error event
+        yield pack(
+            {
+                "node": "error",
+                "update": dumps({
+                    "error": True,
+                    "message": str(e),  # String representation of the error
+                    "error_type": type(e).__name__,  # Exception class name
+                    "type": "stream_initialization_error",
```
yellowcap, line 215: A lot of repetieion here, any way to not having duplication.

yellowcap (commented):
Not sure about the approach of printing the error statement, otherwise ok.

src/api/app.py
srmsoumya: `str(e)` will print just the error message & not the entire stack trace.

PR #229, opened 2025-07-22
Add thread resume/update/delete functionality

leothomas:
This PR implements the following: 
- Adds an editable `name:str` column to the `threads` table, set to `Unnamed Thread` by default
- Updates the API to enable modifying a thread name 
- Update the API to enable deleting a thread (will delete the record from the `threads` table and all associated checkpoints)
- Adds a functionality to the streamlit frontend to demonstrate resuming a conversation from a persisted checkpoint


leothomas:
Quick demo of functionality here: https://drive.google.com/file/d/1g7N4YROFqv4aSv4g1Kqm5kcReAr1xsLn/view?usp=sharing

srmsoumya (approved):

Issue #226, opened 2025-07-22
Move GADM geocoding and geometry fetch to a separate service


geohacker:
I propose that we move the GADM aoi service into a different service that can be scaled independently and queries outside of the zeno api. This makes scaling the zeno api easier due to the large amount of data needs (parquet files) for the GADM queries. 

We can also have a proxy API that's part of the geostore https://github.com/wri/project-zeno/issues/214 so the frontend can seamless fetch the GADM geojson. As it looks like the vector approach for selecting the geometry is not working out well.

cc @batpad @kamicut @yellowcap @srmsoumya @sunu

sunu:
This is happening in https://github.com/wri/project-zeno/pull/241. We are now storing all the geometries in PostGIS

sunu:
Yes, this is done in #241 and all the geometries live in postgis now

PR #230, opened 2025-07-23
Separate geometry handling from AOI responses

srmsoumya:
Refactored AOI architecture to separate geometry data from metadata responses for improved performance and cleaner API design.

**Key Changes**
- Geometry API endpoints: Added `/api/geometry/{source}/{src_id}` and `/api/geometry/batch` for fetching geometry data separately
- Streamlined AOI responses: Removed geometry from `pick_aoi` tool responses, now returns only metadata with `source` and `src_id` for geometry lookups
- Streamlit frontend lazy loading: Streamlit fetches geometry via API endpoints only when needed for map rendering
- Database schema updates: Updated table references and parquet file structure for optimized geometry storage
- Documentation updates: Added ID field mappings and updated schema examples to reflect new architecture

**Breaking Changes**
- AOI responses no longer include `geometry` field
- Subregion AOIs now returned as simple arrays instead of wrapped objects
- Frontend to use geometry API endpoints for map rendering

cc @kamicut @danielfdsilva @yellowcap

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -93,6 +93,25 @@ Fetches multiple geometries in a single request for efficient loading.
 
 **Authentication:** Both endpoints require WRI API token authentication.
 
+### AOI ID Field Mappings
+
+The `src_id` field used for geometry lookups is derived from different source columns depending on the data source:
+
+| Source | ID Field (`src_id`) | Description | Example Values |
+|--------|-------------------|-------------|----------------|
+| `gadm` | GID_X (based on subtype) | GADM Geographic Identifier | `"USA"`, `"USA.1_1"`, `"IND.26.20_1"` |
+| `kba` | `sitrecid` | KBA Site Record ID | `3261`, `15089` |
+| `landmark` | `gfw_fid` | GFW Feature ID | `12345` |
+| `wdpa` | `wdpa_pid` | WDPA Protected Area ID | `67890` |
+
+**GADM ID Mapping by Subtype:**
+- `country`: Uses `GID_0` (e.g., `"USA"`, `"IND"`)
+- `state-province`: Uses `GID_1` (e.g., `"USA.1_1"`, `"IND.26_1"`)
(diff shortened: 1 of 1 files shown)


danielfdsilva:
@srmsoumya can you provide an example of a chat response without the geometry?

srmsoumya:
@danielfdsilva Yes, this change is documented in  `docs/README.md` under the "Example pick aoi output" and schema sections.

Here's an example of what an AOI chat response looks like now without geometry:
```json
{
    "aoi": {
        "GID_0": "IND",
        "COUNTRY": "India",
        "GID_1": "IND.26_1", 
        "NAME_1": "Odisha",
        "GID_2": "IND.26.20_1",
        "NAME_2": "Koraput",
        "name": "Koraput, Odisha, India",
        "source": "gadm",
        "src_id": "IND.26.20_1",
        "subtype": "district-county"
        // Note: geometry excluded - fetch via /api/geometry/gadm/IND.26.20_1
    },
    "subregion_aois": null,
    "subregion": null,
    "aoi_name": "Koraput, Odisha, India",
    "subtype": "district-county",
    "message": "Selected AOI: Koraput, Odisha, India, type: district-county"
}
```
The key changes are:
- No geometry field in the response
- Added `source` and `src_id` fields for geometry lookups
- Frontend can fetch geometry separately via `/api/geometry/gadm/IND.26.20_1` when needed for map rendering

geohacker:
@srmsoumya unfortunately, this is not a good idea. Let's not add all of GADM geometries into our application database — that will make the database unnecessarily big and we have to keep writing migrations to update the data when it changes.

Like discussed in https://github.com/wri/project-zeno/issues/226 and https://github.com/wri/project-zeno-next/issues/19 we will:
* Move the current geoparquet search into a separate service
* The frontend will rely on GFW geo API for the time being to fetch geometries.

PR #231, opened 2025-07-23
feat: Move separate geocoding service and integrate with AOI seclection tool

sunu:
Addresses https://github.com/wri/project-zeno/issues/226

@geohacker if this looks alright, I'll clean up and add a separate image build etc. We can also reuse the same API image but the API image is quite large (>6GB) and probably better if we can have a smaller image for this on its own.


sunu:
we decided to switch from DuckDB to PostGIS for the geometry data. This should fix the memory issues and we won't need to download parquet files on every API replica startup anymore.

We'll add tables for geometries_gadm, geometries_kba, geometries_wdpa and geometries_custom in PostGIS, then implement the /api/geometry/{source}/{src_id} endpoint like in #230 to serve from there.

I'll open a separate PR for that.

PR #238, opened 2025-07-24
SQLAlchemy Async

### LINKED ISSUE #235: Make database connection and queries async
We are using uvicorn to serving the API, all ops should be async. Otherwise, it will block the main thread and further requests will have to wait. 

For now there are two main blocking ops:
1. Duckdb for geocoding
2. Postgresql ops

We are removing duckdb to move to postgresql. So let's make postgresql connections async. This should be possible by using the asyncio extension for sqlalchemy https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html

cc @batpad @srmsoumya @sunu

geohacker:
Fix #235 

cc @batpad @srmsoumya @sunu @willemarcel


batpad:
Closing in favour of #282

PR #241, opened 2025-07-24
WIP: geocoding through postgis

sunu:
Probably lots of stuff to fix still (especially source id mapping etc) but this seems to be working with the agent locally eg I can ask for deforestation in Odisha or for kbas in Odisha etc.

Haven't tried the wdpa ingest yet because it's a bit too big.

cc @geohacker @srmsoumya

### DIFF
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -1,6 +1,7 @@
 services:
   db:
-    image: postgres
+    image: postgis/postgis:17-3.5
+    platform:  linux/amd64
     restart: always
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U postgres"]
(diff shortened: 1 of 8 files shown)


sunu:
@srmsoumya can you take a quick look? If this looks all good, I'll manually do the one time data ingest for now and then we can deploy this to staging.

We still need to do few more things like generate some fixtures with GADM geometries, KBAs and indigenous land with those samples etc for local dev setup and testing; and also write some tests for this. But I propose we do that in a separate PR just to unblock others sooner. cc @batpad

srmsoumya:
@sunu overall, this lgtm 🚀 🙌 

Notes: 
- We will have to re-run tests for pick-aoi tool to validate the postgres similarity search performs similar to what we had with duckdb (from our vibe tests yesterday, that seemed to be the case)
- Remove geometries from tool response in a separate PR
cc @yellowcap

srmsoumya (approved):

PR #245, opened 2025-07-24
Set upadate node_type based on message type of messages in update

leothomas:
I noticed that all message appear to have the same type for any single checkpoint. Since we yield one update per checkpoint, we can set the `node_type` for that update based on the message type from the messages in the update. 

If message type == "tool" , `node_type` is set to `tools`
If message type == "human", `node_type` is set to `human`
if message type == "ai" or there are more than 1 message types in the messages, `node_type` is set to `agent`

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -142,7 +142,15 @@ def replay_chat(thread_id):
 
                 update[key] = value
 
-            yield pack({"node": "agent", "update": dumps(update)})
+            mtypes = set(m.type for m in update["messages"])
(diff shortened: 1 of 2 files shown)


leothomas:
Timestamp now available in thread history: 

image


(Note: we only have access to 1 timestamp per checkpoint, but I think that's reasonable, since each checkpoint is one node in the graph's execution)

leothomas:
This is a very self-contained PR - merging

Issue #254, opened 2025-07-25
DIST alerts data pull failing


geohacker:
DIST alerts aren't working at the moment. We see errors like:

```
{"event": "Failed to pull DIST-ALERT data: 'status'", "session_id": null, "user_id": "5da73b530a97330012c5e6ba", "thread_id": "24a4877f-2258-44d4-b61a-2836420b3153", "query": "Analyze DIST alert time series data from Odisha in 2020", "request_id": "6689b337504c4a309ac4fbd8690e96dd", "logger": "src.tools.data_handlers.dist_alerts_handler", "level": "error", "timestamp": "2025-07-23T17:46:39.447224Z", "exception": "Traceback (most recent call last):\n  File \"/app/src/tools/data_handlers/dist_alerts_handler.py\", line 66, in pull_data\n    if result[\"status\"] == \"success\":\n       ~~~~~~^^^^^^^^^^\nKeyError: 'status'"}
```

cc @batpad @leothomas @srmsoumya

PR #258, opened 2025-07-28
fix: improve context layer type annotations

sunu:
fixes https://github.com/wri/project-zeno/issues/254

cc @srmsoumya


sunu:
We fixed the data in `data/zeno_data_clean.csv` to so that it reflects the context layer name expected by the DIST alert api. Closing this PR as it's not required anymore.

PR #259, opened 2025-07-28
add API endpoint for users to add ratings to traces

batpad:
Opening this as a Draft PR - will need some rebasing and probably handling of migration conflicts.

The way this is meant to work:

 - With https://github.com/wri/project-zeno/pull/243 the backend will now send a `trace_id` to the frontend at the end of every "interaction" - i.e. series of responses to a single user prompt.
 - The frontend can then make a POST request to `/api/ratings` with `thread_id`, `rating`, and `trace_id` (rating is either `1` or `-1`.

TODO on this PR:

 - Fix any conflicts
 - Add `comments` field to allow users to add free text comments to the rating (optional TEXT field)
 - Make sure tests pass and cover all endpoints / cases

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -17,6 +17,7 @@ from fastapi.security import HTTPBearer
 from itsdangerous import BadSignature, TimestampSigner
 from langchain_core.load import dumps
 from langchain_core.messages import HumanMessage
+from langfuse import Langfuse
 from langfuse.langchain import CallbackHandler
 from pydantic import BaseModel, Field
 from sqlalchemy import select, text
@@ -145,6 +146,7 @@ async def logging_middleware(request: Request, call_next) -> Response:
 
 
 langfuse_handler = CallbackHandler()
+langfuse_client = Langfuse()
 
 # HTTP Middleware to asign/verify anonymous session IDs
 signer = TimestampSigner(os.environ["COOKIE_SIGNER_SECRET_KEY"])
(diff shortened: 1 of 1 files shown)


willemarcel (approved):
It's ready to merge now. My main changes were:
- Move the serialization schemas to `schemas.py`, and keep only the Orm objects in the `data_models.py`
- Use fixtures to create Threads
- I've also discovered that some of the fixtures I created previously were a bit weak, so I improved it and also improved the authentication override function

I would prefer the thread id in the ratings' endpoint URL, instead of in the payload.

batpad:
Thanks so much @willemarcel !

> I would prefer the thread id in the ratings' endpoint URL, instead of in the payload.

I agree. We also have a `trace_id` though - so do you think it should be like:

`/api/threads/<thread_id>/rating` and we accept `trace_id`, `rating`, and `comment` in the payload, or does it make sense to also put the `trace_id` in the URL?

I think I feel okay about `/api/threads/<thread_id>/rating` . Am happy to go ahead and make this change and get another review from you.

batpad:
@willemarcel made the change to the URL structure. Could you please review again whenever you get a chance? Thanks 🙏

willemarcel:
Nice! It's all good!

batpad:
@willemarcel am trying to see if I can get this posting the trace ratings to Langfuse as well - if this is easy, I'll add to this PR, or else we can do that as a separate PR.

yellowcap (approved):
I did not run this locally but looks good overall. My only comment is about the rating.

The designs look like its thumps-up / thumbs-down, so how are we thinking about the integer rating for this case? I think we can do `0`/`1` for the start, but not sure.

batpad:
@yellowcap that's a very good question. In the database, we are storing this now as `-1` and `1` for thumbs-down / thumbs-up. And I believe sending the same to Langfuse.

Happy to change this up and go with any other recommendation.

PR #268, opened 2025-08-01
Refactor auth and allowed domain handling

sunu:
- Uses HTTPPBearer so we can easily authenticate in the Swagger UI Openapi docs too and auth header parsing is handled by FastAPI

- Fixes security issue where allowed domain list could be overridden through query params and adds a test case for this issue

### DIFF
--- a/tests/api/test_anonymous_users.py
+++ b/tests/api/test_anonymous_users.py
@@ -1,4 +1,5 @@
 """Tests for anonymous user access and quota enforcement."""
+
 import pytest
 from unittest.mock import patch
 from src.utils.config import APISettings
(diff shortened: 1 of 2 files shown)


leothomas (approved):
great catch! So the issue is that FastAPI passes query parameters down to dependencies, even though those parameters aren't defined in the RequestModel? That's kind of crazy that fastapi would blindly pass through any parameters

sunu:
@leothomas The parameter was being passed through because we had `fetch_user_from_rw_api` as a dependency. [Dependency injection](https://fastapi.tiangolo.com/tutorial/dependencies/) in FastAPI basically allows us to add parameters to views as necessary but we have to be careful what we add 😅

PR #269, opened 2025-08-01
Add migration to remove unique key constraint on daily usage table

leothomas:
The daily usage table uses a composite primary key: `user_id + date`. 

Since these two fields are defined a composite primary key, uniqueness of their combination will be automatically enforced. There is currently an uniqueness constrain on `user_id` which blocks new values being added, whereas we write a new record per user/date, so I'm removing the `user_id` uniqueness constraint

### DIFF
--- a/db/alembic/versions/d1d3e7357b26_remove_unique_key_constraint_on_daily_usage.py
+++ b/db/alembic/versions/d1d3e7357b26_remove_unique_key_constraint_on_daily_usage.py
@@ -17,8 +17,25 @@ depends_on = None
 
 
 def upgrade():
-    # Drop the unique constraint on 'id' column
-    op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")
+    # Drop the unique constraint on 'id' column if it exists
+    # op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")
+
+    op.execute(
+        """
+    DO $$
+    BEGIN
+        IF EXISTS (
+            SELECT 1
+            FROM pg_constraint
+            WHERE conname = 'daily_usage_id_key'
+              AND conrelid = 'daily_usage'::regclass
+        ) THEN
+            ALTER TABLE daily_usage DROP CONSTRAINT daily_usage_id_key;
+        END IF;
+    END
+    $$;
+    """
+    )
 
 
 def downgrade():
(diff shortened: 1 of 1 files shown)


batpad:
@leothomas do you know if this will fail if the uniqueness constraint does not exist? Is there a good way to make it so the uniqueness constraint is dropped only if it exists? Am just worried since we manually removed the constraint on staging that this might fail when deployed.

cc @sunu

leothomas:
@batpad I've updated the PR to use raw SQL to execute the migration conditionally if the index exists: 

```sql
DO $$
    BEGIN
        IF EXISTS (
            SELECT 1
            FROM pg_constraint
            WHERE conname = 'daily_usage_id_key'
              AND conrelid = 'daily_usage'::regclass
        ) THEN
            ALTER TABLE daily_usage DROP CONSTRAINT daily_usage_id_key;
        END IF;
    END
    $$;
```

Here's how I tested it: 

1. Assert current migration fails if the uniqueness constrain does *not* exist (we expect the second call to fail) ✅  : 
```python 
# 
op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")
op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")
```

2. Assert the new migration successfully removes the constraint (we expect the second `op.drop_constraint()` call to fail because the constrain has been removed) ✅ : 
```python 
op.execute(
  """
DO $$
BEGIN
  IF EXISTS (
      SELECT 1
      FROM pg_constraint
      WHERE conname = 'daily_usage_id_key'
        AND conrelid = 'daily_usage'::regclass
  ) THEN
      ALTER TABLE daily_usage DROP CONSTRAINT daily_usage_id_key;
  END IF;
END
$$;
"""
op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")
    )
```

3. Assert the new migration handles the case where the constraint does *not* exist ✅ : 
```python 
    op.drop_constraint("daily_usage_id_key", "daily_usage", type_="unique")

    op.execute(
        """
    DO $$
    BEGIN
        IF EXISTS (
            SELECT 1
            FROM pg_constraint
            WHERE conname = 'daily_usage_id_key'
              AND conrelid = 'daily_usage'::regclass
        ) THEN
            ALTER TABLE daily_usage DROP CONSTRAINT daily_usage_id_key;
        END IF;
    END
    $$;
    """
    )
```

batpad:
Thanks @leothomas !

batpad (approved):

PR #270, opened 2025-08-04
Drop geometry from pick-aoi tool response; pass geometry by id

sunu:
Follow up to https://github.com/wri/project-zeno/pull/241. Drops geometry from pick-aoi tool response now that geometries can be passed around by id only.

Other enhancements:
- Update streamlit example to pass geometries by id
- Update metadata endpoint to return more geometry related mappings for subtypes etc

Also fixes a few other minor issues:
- If the geometry is passed by id and the query doesn't mention the geometry name, the generated insight often doesn't mention the AOI name in the title. Fixed this by passing AOI name explicitly to the insight generation prompt.
- Fixed a bug where the LLM was appending a part of the query to the user picked AOI name by adding delimiters to user actions.

cc @srmsoumya


src/api/app.py
```diff
@@ -289,21 +293,21 @@ def stream_chat(
         for action_type, action_data in ui_context.items():
             match action_type:
                 case "aoi_selected":
-                    content = f"User selected AOI in UI: {action_data['aoi_name']}"
+                    content = f"<user-action>User selected AOI in UI: {action_data['aoi_name']}</user-action>"
```
srmsoumya, line 296: @sunu can we avoid using tags here? this might work for `sonnet` & `claude` but disrupt other models.

sunu: @srmsoumya should we do new lines after the message then?

This is the problem I'm trying to fix for context:
> format user action messages for clarity
otherwise the llm often gets confused and some times appends the first portion of the query to the aoi name.
>
> For example, for a sequence of messages like this:
User selected AOI in UI: Koraput
tree cover loss in last decade
>
> the llm picks "Koraputtree" as the AOI name

sunu: something like this:
image

sunu: Updated the PR to use new lines instead of tags -- that seems to work as well.

srmsoumya: ah weird, this was not happening last I checked, yes a new line is better. Ideally what we want is:
`I see you have selected Koraput as AOI and are interested in tee cover loss in last decade...`

sunu: Yes, with the fix in place this works as expected:

image

batpad:
@sunu @kamicut how shall we handle merging this and coordinating with frontend update? Should we wait for the frontend PR to be merged? Or merge this in and then make changes to frontend? Let's figure how best to coordinate.

sunu:
@batpad I think we should deploy the backend pr first and then deploy https://github.com/wri/project-zeno-next/pull/94. The frontend wont have aoi geometries on the map until https://github.com/wri/project-zeno-next/pull/94 but chat and insights generation etc will work as usual.

I think we can time the deploys to be one after the other; so it should be fine.

yellowcap:
ok resolved conflict & merging

PR #239, opened 2025-07-24
Make Tools Async

srmsoumya:
Ref #234 

Convert blocking HTTP requests to async for better uvicorn performance.

## TODO

- [x] `src/tools/data_handlers/dist_alerts_handler.py` - Replace `requests` with `httpx`
- [x] `src/tools/pull_data.py` - Make tool function async
- [x] `src/tools/pick_aoi.py` - Make tool function async
- [x] `src/tools/pick_dataset.py` - Make tool function async  
- [x] `src/tools/generate_insights.py` - Make tool function async

## Notes
- [x] May be skip `gfw_sql_handler.py` for now (WRI moving to new analytics endpoint, this will be depricated)
- [x] `pick_aoi.py` will move to geometry queries to a separate service, so we don't have to handle duckdb queries on the application side

cc @batpad @geohacker @leothomas @yellowcap

PR #278, opened 2025-08-05
poll for success for DIST Alert API calls

sunu:
DIST Alert API returns "pending" status on the first request for a particular AOI. It takes a bit of time (less than a second in most cases) to compute the result. So in this PR, we have added a wait and retry mechanism when calling the DIST Alert API for analytics.

This is doing synchronous sleep now that will block but once we move to async tools in https://github.com/wri/project-zeno/pull/239 we should use asyncio.sleep to wait without blocking.

cc @yellowcap @batpad @geohacker


yellowcap (approved):
LGTM

PR #283, opened 2025-08-06
Enable new analytics api layers

yellowcap:
This replaces the dist alert specific pull data handler with one that can handle all currently supported combinations in the new analytics api. A few comments:

- I would propose to disactivate the GFW api handler because the new analytics api also supports tree cover loss by driver. So we might not need the GFW endpoints anymore.

- I also think this will require updates on the data lookp table on the dataset picker, because we support more layers now that are not in the csv currently.

- I added unittests for all combinations of AOI types and data analytics layers. All test pass, except for some combinations where the API times out. Here is a summary of my local test run.

image


yellowcap:
Ok got this to work on streamlit, so it should be good to go. @srmsoumya and evtl @sunu please test this locally to see if it works for you too.

NOTE: I have not yet synced the RAG database, but I did sync the updated clean_csv file. So to test locally:

The latest csv reference file currently is

```bash
aws s3 cp s3://zeno-static-data/zeno_data_clean_v2.csv data/
```

then run

```bash
python src/ingest/embed_datasets.py
```

This will update the local database at `data/zeno-docs-openai-index`.

This should be the only change in the underlying data files.

I created tests for the pick dataset as well, so to test if things work run the two new tests:

```bash
uv run pytest tests/tools/test_pull_data.py
uv run pytest tests/tools/test_pick_dataset.py
```

And ask some questions in streamlit too 🤓

Issue #281, opened 2025-08-05
Make the geometry api and the llm tools handle custom geometries


sunu:
Right now the llm tools and the geometry api can handle `gadm`, `kba`, `landmark` etc as geometry types. We should also add user uploaded / drawn custom aoi geometries to that list of geometry types that the llm tools and the geometry api can handle.

## Using custom geometries with the analytics api
For pulling analytics for custom geometries from the analytics api, we need to pass one or more custom geometries as Features in a FeatureCollection as defined in the `DistAlertsAnalyticsIn` schema in the [openapi docs of the analytics api](http://analytics-416617519.us-east-1.elb.amazonaws.com/docs).

Since each custom geometry can contain one or more geometries of its own, we should map those geometries into a GeometryCollection while querying the analytics endpoint. The metadata of a custom geometry like its id and name can be passed in the properties of the individual Feature representing the custom geometry. Here's an example of what the `aoi` will look like when querying the analytics api for multiple custom geometries:
```
{
 "type": "FeatureCollection",
 "features": [
   {
     "type": "Feature",
     "geometry": {
       "type": "GeometryCollection",
       "geometries": [
         {
           "type": "Polygon",
           "coordinates": [...]
         },
         {
           "type": "Polygon",
           "coordinates": [...]
         }
       ]
     },
     "properties": {
       "name": "My Custom AOI - North District",
       "id": "550e8400-e29b-41d4-a716-446655440001"
     }
   },
   {
     "type": "Feature",
     "geometry": {
       "type": "GeometryCollection",
       "geometries": [
         {
           "type": "Polygon",
           "coordinates": [...]
         },
         {
           "type": "Polygon",
           "coordinates": [...]
         }
       ]
     },
     "properties": {
       "name": "My Custom AOI - South Zone",
       "id": "f47ac10b-58cc-4372-a567-0e02b2c3d479"
     }
   }
 ]
}
```

Since each Feature / custom area will have its own id, the returned results can be correlated based on those ids.

See this Slack thread for more context on this topic: https://developmentseed.slack.com/archives/C07SBN17G59/p1754586981731869

cc @kamicut @yellowcap @srmsoumya @vgeorge

PR #297, opened 2025-08-11
Enable use of custom geometries in chat context and llm tools

sunu:
With this in place, it's possible to use custom geometries in the chat by their name and pass the custom geometry to the analytics api to get relevant data. (But currently requests to the analytics api seem to get stuck in pending state forever. here's an example: http://analytics-416617519.us-east-1.elb.amazonaws.com/v0/land_change/dist_alerts/analytics/c3be9b37-5a88-5311-b1d0-226421aa4709) (tracked by https://github.com/wri/project-zeno/issues/298)

Fixes https://github.com/wri/project-zeno/issues/281.

image


src/tools/data_handlers/analytics_handler.py
```diff
@@ -71,29 +70,61 @@ def _get_aoi_type(self, aoi: Dict) -> str:
             return "indigenous_land"
         elif aoi["subtype"] == "protected-area":
             return "protected_area"
+        # custom_area instead of custom-area because
+        # that's what the existing
```
sunu, line 74: please ignore this comment. we should use `custom-area` and update the frontend code to do the same as well

sunu: that's done in https://github.com/wri/project-zeno-next/pull/115

src/tools/data_handlers/analytics_handler.py
```diff
@@ -149,13 +185,13 @@ async def _poll_for_completion(
...
-        max_retries: int = 3,
-        poll_interval: float = 0.5,
+        max_retries: int = 5,
+        poll_interval: float = 1,
     ) -> Dict | str:
         """Poll the API until the request is completed or max retries exceeded."""
         for attempt in range(max_retries):
             logger.info(f"Polling attempt {attempt + 1}/{max_retries}")
-            await asyncio.sleep(poll_interval)
+            await asyncio.sleep(poll_interval * (attempt + 1))
```
yellowcap, line 194: Since this is progressive now, can we do poll inverval 0.5? This will increase by 0.5 secons in each of the retries right?

src/tools/data_handlers/analytics_handler.py
```diff
@@ -313,7 +351,7 @@ async def pull_data(
                     "Analytics request is pending, will retry with polling..."
                 )
                 result = await self._poll_for_completion(
-                    endpoint_url, payload, max_retries=3, poll_interval=0.5
+                    endpoint_url, payload, max_retries=5, poll_interval=1
```
yellowcap, line 354: same here, since the poll interval increases with each try, lets use 0.5?

yellowcap (commented):
Looks good to me overall, but have not tested this locally. I just have a small comment on the timeouts otherwise its ok for me.

PR #309, opened 2025-08-12
Fix tests for when test-db already has users

yellowcap:
Not sure if this makes sense, to me it looks like the test DB is stateful so I am getting errors when trying to load the test users in my test DB.

My guess is that the test DB is stateful because of the geodata that needs to be pre-loaded in our database anyway?

But maybe I am getting this wrong. Feel free to close if this does not make sense.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,22 +1,22 @@
 """Test configuration and fixtures."""
+
 import os
 import uuid
+from collections.abc import AsyncGenerator
+
 import pytest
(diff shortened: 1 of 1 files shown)


batpad:
@yellowcap can you say what errors you are running into? The database tables should all be truncated at the end of each test run, so this shouldn't be needed.

cc @willemarcel

yellowcap:
Closing this in favor of #311

PR #310, opened 2025-08-12
Name threads via LLM

### LINKED ISSUE #248: Name threads via LLM
Ideal case: name threads via the names of areas in the thread, queries and topic of conversation, and automatically set the thread name once user has engaged with n number of prompts.

cc @srmsoumya @leothomas

geohacker:
Fixes #248 

I didn't make this an endpoint. The API will return the thread name after the first query, which the frontend is already using to set the name.

cc @kamicut @LanesGood

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -603,6 +603,24 @@ async def check_quota(
     return {}
 
 
+async def generate_thread_name(query: str) -> str:
+    """
+    Generate a descriptive name for a chat thread based on the user's query.
+
+    Args:
+        query: The user's initial query in the thread
+
+    Returns:
+        A concise, descriptive name for the thread
+    """
+    try:
+        prompt = f"Generate a concise, descriptive title (max 50 chars) for a chat conversation that starts with this query:\n{query}\nReturn strictly the title only, no quotes or explanation."
+        response = HAIKU.invoke(prompt)
+        return response.content[:50]  # Ensure we don't exceed 50 chars
+    except Exception as e:
+        logger.exception("Error generating thread name: %s", e)
+        return "Unnamed Thread"  # Fallback to default name
+
 @app.post("/api/chat")
 async def chat(
     request: ChatRequest,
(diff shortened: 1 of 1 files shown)


geohacker:
The thread name is based on the first query:

image

src/api/app.py
```diff
@@ -603,6 +603,24 @@ async def check_quota(
...
+
+    Args:
+        query: The user's initial query in the thread
+
+    Returns:
+        A concise, descriptive name for the thread
+    """
+    try:
+        prompt = f"Generate a concise, descriptive title (max 50 chars) for a chat conversation that starts with this query:\n{query}\nReturn strictly the title only, no quotes or explanation."
+        response = HAIKU.invoke(prompt)
```
sunu, line 618: We should switch this `.invoke()` call and all others in the code base to `.ainvoke()` to make the call async.

geohacker: Oh! good catch. Let me fix

src/api/app.py
```diff
@@ -603,6 +603,24 @@ async def check_quota(
...
+    Generate a descriptive name for a chat thread based on the user's query.
+
+    Args:
+        query: The user's initial query in the thread
+
+    Returns:
+        A concise, descriptive name for the thread
+    """
+    try:
+        prompt = f"Generate a concise, descriptive title (max 50 chars) for a chat conversation that starts with this query:\n{query}\nReturn strictly the title only, no quotes or explanation."
```
srmsoumya, line 617: This looks good! for better reliability we can use langchain structured outputs, here how we could implement this:
```python
from pydantic import BaseModel, Field

class ChatTitle(BaseModel):
    title: str = Field(max_length=50, description="Concise chat title")

prompt = f"Generate a concise, descriptive title for a chat conversation that starts with this query:\n{query}"

response = await HAIKU.with_structured_output(ChatTitle).ainvoke(prompt)
return response.title
```

srmsoumya (approved):

PR #312, opened 2025-08-12
Add lint tasks and lint files

yellowcap:
Run linters through entire codebase.


batpad (approved):
Minor: The Github action to lint currently runs twice for pushes to open pull requests. I think we can do either on `push` or `pull_request`, I don't think we need to have both.

Issue #292, opened 2025-08-07
Spike injecting dataset specific instructions into insight prompt


01painadam:
## Goal

The current [insight generation prompt](https://github.com/wri/project-zeno/blob/8fdfd719c6fbe14d3c0106de6163b81cbfb7815a/src/tools/generate_insights.py#L55) is dataset agnostic and could lead to incorrect usage of data provided by the data api, or misleading interpretations of the data.

This spike will investigate injecting additional, dataset specific instructions, to the insight generation prompt in order to improve the efficacy and quality of generated insights.

## Hints

- We can use the `function/usage` and `caution` information available [here](https://onewri.sharepoint.com/:x:/r/sites/LandandCarbonWatch/_layouts/15/doc2.aspx?sourcedoc=%7Bc75a6529-76bb-4061-9e86-e6700e083170%7D&action=edit&wdinitialsession=3c994ef2-ce1d-0b3b-d764-3d79d252ae38&wdrldsc=5&wdrldc=1&wdrldr=EditModeCannotAcquireLockUnrecoverable&wdenableroaming=1&wdlcid=en-GB&wdorigin=AuthPrompt&wdredirectionreason=Force_SingleStepBoot).
- Work with the dataset providers to write specific insight prompts
- Work with the data api team to ensure the returned data is used correctly

01painadam:
A first pass at some dataset-specific instructions can be found [here](https://gfw.atlassian.net/wiki/spaces/LCL/pages/2132344843/Dataset+specific+insight+instructions).

PR #314, opened 2025-08-12
Better integrate rag setup

### LINKED ISSUE #287: Rethink S3 assets and distribution
Currently, we have an init container that [syncs assets from S3](https://github.com/wri/project-zeno-deploy/blob/develop/helm/zeno/templates/deployment.yaml#L23-L25) this includes geo data and LLM indexes. We should revisit this:

* Only download necessary things. Given GADM and geo data are now part of the database, we can skip this?
* Some of the data csvs are small enough to be packages as part of the code
* Some of indexes need to be updated regularly, they can also be part of the code

cc @yellowcap @batpad @srmsoumya @sunu @leothomas

### LINKED ISSUE #307: Add more context to the insight
We are already tracking a lot of metadata about the current AOI and datasets in the state. The insight could surface these as fields (non-generative), and then be used to display on insights. Examples are

- Description field for "info" button
- Cautions field for "cautions" tab
- Links to further reading
- Links to dataset docs

The intention here is to purely send more metadata that is not LLM driven for adding transparency.

yellowcap:
This refactors how we manage the dataset list.

### This PR is mainly to solve #287 

- Move dataset info int yaml file
- Use yaml to load dataset info into python as dict
- Use this dict in the `ingest/embed_dataset.py` script
- Run the same script in the Dockerfile
- Simplify and centralize handling of datasets throughout pick dataset and pull data

Closes #287

### Side effects

- Inject instructions from dataset file into create insights tool, adressing #292 
- Increases the amount of information returned by the pick dataset tool, addressing #307

Closes #307 and #292

### For the reviewers

@sunu and @batpad I added a line to the Dockerfile but that might be the wrong way to do it. Please advise

@srmsoumya please review the refactoring of the dataset handling. Its more elegant now and more centralized, but please run locally and revise.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,3 @@ WORKDIR /app
 
 # Install the dependencies
 RUN uv sync --frozen
-
-# Command to build the RAG database
-RUN uv run ingest/embed_datasets.py


Dockerfile
```diff
@@ -28,4 +28,5 @@ WORKDIR /app
 # Install the dependencies
 RUN uv sync --frozen
 
-# Command to run the application.
+# Command to build the RAG database
```
sunu, line 31: This won't really work. The embed datasets script will run at build time and the output will be written to the data directory where the build process runs (eg the github actions runner vm).

As we don't include the updated data directory in the docker image, the updated index wont really be saved in the built image or in the s3 bucket.

I would propose we run the embed datasets script in a github action watching the `analytics_datasets.yml` file so that the action runs automatically when the yaml file changes and then uploads the generated index to the s3 bucket. And we should keep current flow of pulling the index from the s3 bucket on startup. This will ensure changes to the datasets are tracked and the index is auto updated upon any change.

The other options are to bake the index into the docker image or include it in the git repo. But I feel the earlier option is much cleaner.

yellowcap: ok so I propose that we merge this and manually upload the latest version of the `zeno-docs-openai-index` file to the S3 bucket.

We can then tackle removing the S3 dependency in #319 .

sunu: Sounds good!

We should remove the `RUN uv run ingest/embed_datasets.py` line from the dockerfile before merging otherwise the build will fail.

src/tools/pick_dataset.py
```diff
@@ -109,7 +146,7 @@ async def select_best_dataset(query: str, candidate_datasets: pd.DataFrame):
                     "dataset_id",
                     "dataset_name",
                     "description",
-                    "date",
+                    "content_date",
                     "context_layers",
```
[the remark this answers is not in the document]
yellowcap: Yes agreed, I actually think we can re-surface some of this in the insights as well as additional context to bake it into the insights directly. Like this the insights will become more self-sufficient.

yellowcap:
🎉 moving this along!

PR #316, opened 2025-08-12
Feature/quota info in all responses

leothomas:
This PR adds: 
- 2 headers to the response of all `/chat` requests: `X-Daily-Prompts-Used` and `X-Daily-Prompts-Quota` which contain the number of prompts used and total number of available prompts per 24hr period
- Add these same values to the output of the `/auth/me` endpoint 

Example of response headers for the `/chat` endpoint requests: 

```
{'date': 'Tue, 12 Aug 2025 18:36:59 GMT', 'server': 'uvicorn', 'x-daily-prompts-used': '1', 'x-daily-prompts-quota': '25', 'content-type': 'application/x-ndjson', 'set-cookie': 'anonymous_id=d40e0613-44df-49e9-8252-219657dc73ae.aJuJzA.sQBH7TtCfL-2flhSZQ2sHDv4iaQ; HttpOnly; Max-Age=31536000; Path=/; SameSite=Lax', 'transfer-encoding': 'chunked'}
```

Example of quota info in `/auth/me` endpoint response: 

```json
{"id":"67c64511bb39c598b557123f","name":"Leo Thomas","email":"leo@developmentseed.org","createdAt":"2025-03-04T00:10:58","updatedAt":"2025-04-21T20:13:59","threads":[],"userType":"regular","dailyPromptsUsed":1,"dailyPromptQuota":25}
```


batpad:
@leothomas this looks great!

Minor things:

 - Can we change from `daily-prompts` to just `prompts` - so like `x-prompts-used` and `x-prompts-quota` and similarly in the `/auth/me` endpoint - I think that just keep things more flexible in case we want to switch to weekly / monthly quotas or anything. IF there is some place in the frontend that needs to know when the quota will be reset we could add that as a separate value in the `/auth/me` endpoint like `nextQuotaReset` or something, but I wouldn't worry about that for now.
 - Do you think you could add tests for the functionality? (Am also happy to do that in either this or a separate PR if you don't want to get your head into the tests, but we should have tests to cover this).

If you make the change to remove `daily`, this is fine to merge - am happy to tackle doing the tests separately.

PR #335, opened 2025-08-19
first pass at adding new user_profile fields, migrations and endpoints

batpad:
Creating a draft PR here to continue working through.

cc @leothomas

### DIFF
--- a/src/user_profile_configs/countries.py
+++ b/src/user_profile_configs/countries.py
@@ -252,5 +252,5 @@ COUNTRIES = {
     "ZA": "South Africa",
     "ZM": "Zambia",
     "ZW": "Zimbabwe",
-    "OTHER": "Other"
+    "OTHER": "Other",
 }


tests/api/test_user_profile.py
```diff
@@ -0,0 +1,396 @@
...
+        assert data["lastName"] == "Doe"
+        assert (
+            data["profileDescription"]
+            == "Forest researcher interested in conservation"
+        )
+        assert data["id"] == user.id
+        assert data["name"] == user.name  # Original fields preserved
+
+    @pytest.mark.asyncio
+    async def test_update_profile_success_detailed_fields(
```
willemarcel, line 94: the `test_profile_fields_roundtrip_auth_me` is doing the same as this test function, but also checking the response of auth/me. I would delete this function and keep only `test_profile_fields_roundtrip_auth_me`.

willemarcel (approved):
It looks good. I've done 2 improvements suggestions

PR #339, opened 2025-08-20
Fix created_at and updated_at timestamps on models

### LINKED ISSUE #318: Thread timestamps are incorrect
A thread I just created on 8-13 states that it was created two days ago...

```
agent_id: "UniGuana"
created_at: "2025-08-11T08:57:11.510689"
id: "5aaeb0ae-ea7d-4f06-8b90-8843d7b9bf35"
name: "Queensland Wildfires"
updated_at: "2025-08-11T08:57:11.510734"
user_id: "5da73b530a97330012c5e6ba"
```

None of my threads seem to have accurate timestamps for `created_at` or `updated_at`. cc @leothomas @srmsoumya

batpad:
Fixes #318 

We were using `default` and `onupdate` a little bit wrong - they needed to be callables so the datetime.now() function is called every time while saving - the way we had it, they were called only when the app instantiated and then that datetime value was cached, leading to all sorts of weirdness with timestamps.

This should fix the issue and includes a test for the thread timestamps.

cc @leothomas

### DIFF
--- a/src/api/data_models.py
+++ b/src/api/data_models.py
@@ -32,8 +32,8 @@ class UserOrm(Base):
     id = Column(String, primary_key=True, unique=True, nullable=False)
     name = Column(String, nullable=False)
     email = Column(String, unique=True, nullable=False)
-    created_at = Column(DateTime, nullable=False, default=datetime.now())
-    updated_at = Column(DateTime, nullable=False, default=datetime.now())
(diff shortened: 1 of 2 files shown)


batpad:
@sunu can I get a review from you here, please?

batpad:
@willemarcel would you be able to review? Thanks.

willemarcel (approved):

PR #346, opened 2025-08-21
Fix pick aoi when custom table not present

yellowcap:
The `user_id` is needed for this function, but it is currently only set if custom table is present.

I moved the user id retrieval out of that if clause.

I noticed because I dont have this table in my localhost, so there was an error in the agent test.


sunu (approved):
LGTM!

Issue #300, opened 2025-08-11
Support multiple area queries


yellowcap:
We want to support queries like "Dist alerts by kba for para brazil" 

**Describe the solution you'd like**

This is already enabled in the aoi picker, and the values are passed to the pull_data tool, but are not used there. So the change to enable this should be not too big and can focus on the pull data tool.

yellowcap:
@LanesGood we thought this was wired up and we were waiting for the analytics api to support the queries. Once the api was up, we realized we had to do a little more work to support it.

With #353 you can now as "county with highest deforestation in india".

For the frontend, this will still select only one region as the AOI on the map (india), but the insights will provide info about the regions in india. This is what @danielfdsilva meant with multi aoi is not supported I think, the frontend does not show groups of areas. We can add that later, for now we keep it like this because selecting multiple aois on the frontend implies state changes that need to be reflected on the backend too. So we keep that for later.

PR #353, opened 2025-08-25
Support subregion AOI in analytics payload

srmsoumya:
Addreses #300 

Adds proper subregion handling in analytics requests, enriches results with readable names and improves AOI tool messaging.

**Notes**
Backward compatible; subregion enrichment only applied when subregion selection is present.

### DIFF
--- a/src/tools/data_handlers/analytics_handler.py
+++ b/src/tools/data_handlers/analytics_handler.py
@@ -404,7 +404,6 @@ class AnalyticsHandler(DataSourceHandler):
                     data_points_count,
                     message_detail,
                 ) = await self._process_response_data(result)
-                logger.debug(f"RAW_DATA: {raw_data}")
 
                 # Enrich raw_data with names
                 if subregion:


src/tools/data_handlers/analytics_handler.py
```diff
@@ -380,10 +399,29 @@ async def pull_data(
...
-                raw_data, data_points_count, message_detail = (
-                    self._process_response_data(result)
-                )
-
+                (
+                    raw_data,
+                    data_points_count,
+                    message_detail,
+                ) = await self._process_response_data(result)
+                logger.debug(f"RAW_DATA: {raw_data}")
```
yellowcap, line 407: Seems like a very long log. Maybe truncate this

```suggestion
                logger.debug(f"RAW_DATA: {raw_data}"[:300])
```

src/tools/data_handlers/analytics_handler.py
```diff
@@ -380,10 +399,29 @@ async def pull_data(
...
+
+                # Enrich raw_data with names
+                if subregion:
+                    subregion_aois_id_to_name = {
+                        format_id(item["src_id"]): item["name"].split(",")[0]
+                        for item in subregion_aois
+                    }
+                    # For tree cover loss, the API returns "id" as the key for the AOI IDs, for other datasets it returns "aoi_id"
```
yellowcap, line 415: Seems like something to cahnge in the api design. We should report this to Justin and team.

yellowcap (approved):
Looks good, just one comment on a log statement.

src/tools/data_handlers/analytics_handler.py
srmsoumya: yes, I have reported this to him over slack

PR #366, opened 2025-08-27
Add anonymous id extraction from headers with API key verification and IP address usage tracking

leothomas:


db/alembic/versions/831ad1e71a80_add_ipaddress_column_to_daily_usage_.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "831ad1e71a80"
+down_revision: Union[str, None] = "32753a3e09e0"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
```
batpad, line 19: There's nothing in the upgrade?

leothomas: No clue how that happened! Just pushed an updated version of the file

batpad:
@leothomas am definitely a bit worried about this one, but if tests pass, I'd say YOLO lets go and we can work through any issues as the frontend works to integrate.

PR #371, opened 2025-08-28
  Add E2E testing for agent evaluation

srmsoumya:
Adds modular end-to-end testing framework supporting:
- Local agent testing (direct agent invocation)
- API testing (via client.py)
- CSV dataset input (experiments/e2e_test_dataset.csv)
- Langfuse dataset integration
- Result export to CSV files and Langfuse traces

**How to run tests**

> Basic usage (1 test, local mode):
```bash
pytest tests/agent/test_e2e.py -v -s
```

> Configuration via environment variables:
```bash
# Test mode: "local" (default) or "api"
export TEST_MODE=local

# Sample size: 1 (default, CI-friendly), -1 (all rows)
export SAMPLE_SIZE=5

# API mode (requires running server)
export TEST_MODE=api
export API_BASE_URL=http://localhost:8000
export API_TOKEN=your_token_here

# Langfuse dataset mode
export LANGFUSE_DATASET=your_dataset_name
```

Test results are saved in local mode to: `data/tests/simple_e2e_TIMESTAMP_summary.csv`

cc @Quantisan @sunu @batpad


sunu (commented):
I was able to run this locally with the list of env vars documented by @srm.

@batpad if we want to run this in the ci, we need a way to provision a api key to use for running tests etc against the api. Is there an existing way to do that?

Additionally, we probably also need to store the resulting csv results in a bucket for future reference but that should be straight-forward.

batpad:
> @batpad if we want to run this in the ci, we need a way to provision a api key to use for running tests etc against the api. Is there an existing way to do that?

This would be like some key that can be used instead of the WRI auth bearer token? Currently, the API does not support something like this. Happy to include whatever you think is best - in general, I don't think its a bad idea to allow some idea of a "machine user" - it will probably be handy for a bunch of things. If you can outline exactly how the behaviour should be, can look into how best to implement this.

batpad:
@srmsoumya @sunu lets merge this maybe and worry about CI later?

PR #386, opened 2025-09-03
CLI tool to manage machine users and access keys

sunu:
Adds the following:
- machine user system for programmatic API access with access token
- CLI tool for creating/managing machine users and API keys with rotation/revocation
- higher quota (1000 queries / day) for machine users
- tests for machine user and access key functionalities

### DIFF
--- a/db/alembic/versions/08eca0bda924_add_machine_users_and_api_keys.py
+++ b/db/alembic/versions/08eca0bda924_add_machine_users_and_api_keys.py
@@ -1,7 +1,7 @@
 """add_machine_users_and_api_keys
 
 Revision ID: 08eca0bda924
-Revises: 831ad1e71a80
+Revises: 95d9d8ca3bf1
 Create Date: 2025-09-01 09:06:34.832320
 
 """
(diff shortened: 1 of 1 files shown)


sunu:
@batpad could you give this a try locally and do a review please?

src/api/app.py
```diff
@@ -514,6 +515,10 @@ async def fetch_user_from_rw_api(
 
     token = authorization.credentials
 
+    # Handle machine user tokens with zeno-key prefix
+    if token and token.startswith(f"{MACHINE_USER_PREFIX}_"):
```
batpad, line 519: Minor, can we use `:` as the separator instead of `_` - this would be consistent with how we're handling tokens from anonymous users, where we prefix with ANONYMOUS_USER_PREFIX (see code below in same file).

db/alembic/versions/08eca0bda924_add_machine_users_and_api_keys.py
```diff
@@ -0,0 +1,82 @@
...
+down_revision: Union[str, None] = "831ad1e71a80"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Add machine user fields to users table
+    op.add_column(
+        "users", sa.Column("is_machine_user", sa.Boolean(), default=False)
```
batpad, line 26: @sunu can you explain why we need to have this boolean field in addition to having a separate MACHINE_USER type?

sunu: @batpad thanks for the catch! I don't think we need a separate field. I've removed it.

PR #389, opened 2025-09-03
Small improvements on comparison

yellowcap:
Main thing is a new test for comparison with subregions.

Also some instructions to show subregions in graph when comparing.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1953,19 +1953,6 @@ wheels = [
 ]
 
 [[package]]
-name = "langchain-ollama"
-version = "0.3.6"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "langchain-core" },
-    { name = "ollama" },
-]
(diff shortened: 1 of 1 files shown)


src/tools/generate_insights.py
```diff
@@ -92,9 +92,9 @@ class InsightResponse(BaseModel):
 - For stacked-bar: [{{"category": "2020", "metric1": 100, "metric2": 50}}] + set series_fields
 - For grouped-bar: [{{"year": "2020", "type": "metric1", "value": 100}}] + set group_field
 - If dates are present, order those in chronological order (not alphabetically)
+- By default report the data as-is. Only do aggregation if asked for by the user. For example, if the user asks to compare kba's in two regions, report the data at the kba level, not the region level.
 
 User query: {user_query}
-Area of interest: {aoi_name}
```
srmsoumya, line 97: @yellowcap afaik we need the `aoi_name` here, when the user selects an AOI from the UI, the LLM is not aware of the name, so it is useful to have it here.

srmsoumya (changes requested):

PR #391, opened 2025-09-03
try fixing postgres connection pooling issues

batpad:
Attempting to fix Postgres connection pooling issues.

Pasting AI summary of changes:

## Changes Made

  Priority 1: AsyncConnectionPool for Checkpointer ✅

  File: src/agents/agents.py
  - Added psycopg_pool.AsyncConnectionPool import
  - Created global connection pool _checkpointer_pool with proper configuration:
    - min_size=5, max_size=20 connections
    - Proper connection parameters (autocommit=True, prepare_threshold=0)
  - Added get_checkpointer_pool() function to initialize pool on first use
  - Added close_checkpointer_pool() function for cleanup
  - Fixed the critical issue: fetch_checkpointer() now uses the connection pool instead of creating individual connections

 ### Priority 2: Proper Lifespan Management ✅

  File: src/api/app.py
  - Enhanced the lifespan() function to:
    - Initialize the checkpointer pool on startup
    - Properly dispose SQLAlchemy engine on shutdown
    - Close checkpointer pool on shutdown

 ### Key Improvements

  1. Eliminates connection leaks: No more individual AsyncConnection.connect() calls that never get closed
  2. Proper resource management: Both connection pools are initialized on startup and cleaned up on shutdown
  3. Connection reuse: The checkpointer now uses pooled connections instead of creating new ones
  4. Scalable: Pool sizes can handle higher concurrent load (5-20 connections for checkpointer)

 ###  Expected Impact

  - Dramatically reduced connection usage: Instead of creating new connections per request, connections are reused from pools
  - Better performance: Connection pool eliminates connection setup/teardown overhead
  - Prevents "connection slots exhausted" errors: Proper connection lifecycle management
  - Load test ready: Should handle much higher concurrent requests without exhausting connections

  The changes maintain backward compatibility while fixing the core connection management issues. You can now deploy and run your load tests - the connection exhaustion problems should be resolved!

cc @sunu @geohacker @yellowcap

### DIFF
--- a/src/agents/agents.py
+++ b/src/agents/agents.py
@@ -4,8 +4,8 @@ from datetime import datetime
 from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
 from langgraph.graph.state import CompiledStateGraph
 from langgraph.prebuilt import create_react_agent
-from psycopg import AsyncConnection
 from psycopg.rows import dict_row
+from psycopg_pool import AsyncConnectionPool
 
 from src.graph import AgentState
 from src.tools import (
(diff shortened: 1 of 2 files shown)


sunu (commented):
Looks safe to try. I think I have 2 questions:
- why not a common connection pool for both the api and the checkpointer since they're using the same database?
- And related to the same point, are the kwargs passed to the checkpointer connection pool different from what we would use for a connection pool for the api?

batpad:
@sunu both good questions.

I think my thinking here was really to get something out quickly to test to see if this does fix / help with the errors we were seeing - just to ensure we are on the right track / be able to distinguish between application-level problems and needing more hardware.

We should definitely revisit your questions and do a bit further tuning of managing the pg connections as well as double-check RDS settings.

I will go ahead and merge this though, as I do want to confirm that better pooling greatly helps when load testing - and then we can come back to tweaking / making it even better.

PR #398, opened 2025-09-04
Add dynamic model configuration via environment variables

srmsoumya:
**Update**:
- Add MODEL environment variable to config with default "sonnet"
- Update llms.py to use dynamic model selection from config
- Add model information to /api/metadata endpoint showing current and available models
- Update .env.example with MODEL configuration options

**Available models**: `sonnet, haiku, gemini, gemini-flash, gpt, phi4, smollm2`

**Usage**: Set `MODEL=gemini` in `.env` to change the default model for deployment.

Note: `generate-insights` tool will always use `sonnet` by default for now.

cc @yellowcap


yellowcap (approved):
Looks good

PR #399, opened 2025-09-04
Add metadata to langfuse traces

leothomas:
I've re-enabled the Langfuse backend, as part of the local API setup in order to test this feature. 

One important note: 

I've added the user information to the trace's metadata field, which accepts arbitrary, JSON encodable values. The metadata fields can be used for filtering in the _custom dashboards_ but not in the basic Traces interface. Only the values stores in `Tags` can be used to filter from the Traces interface. It's recommended to use tags for low cardinality values (eg: user type) so we should identify any metadata values that are most important for sorting/filtering if we want to include those as tags.

Screenshot: 

image


src/api/app.py
```diff
@@ -980,7 +992,40 @@ async def chat(
...
+            "role_code",
+            "job_title",
+            "company_organization",
+            "country_code",
+            "preferred_language_code",
+            "gis_expertise_level",
+            "areas_of_interest",
+            "has_profile",
+        ]:
+            langfuse_metadata[attr] = getattr(user, attr, None)
```
batpad, line 1028: i dont think we need to add all this user information for all the langfuse traces? I believe just user id should be enough?

leothomas: For sure! I was just being a data hoarder - but after today's discussion on privacy and PII I've removed it!

src/api/app.py
```diff
@@ -186,7 +186,10 @@ async def logging_middleware(request: Request, call_next) -> Response:
     return response
 
 
-langfuse_handler = CallbackHandler()
+os.environ["LANGFUSE_TRACING_ENVIRONMENT"] = os.getenv("STAGE", "production")
```
batpad, line 189: am still a bit concerned by the extra lines / changes here - but if you think it's okay, happy to merge

PR #401, opened 2025-09-05
Add `get_capabilities` tool and improve follow-up generation

### LINKED ISSUE #357: Hallucinated Capabilities
### Description

Zeno can sometimes hallucinate its own capabilities. There are two locations this is happening:

**1. Self-description**
|Suggested queries | Capabilities |
|-|-|
|Image|Image|

When asking zeno queries like `"Tell me what you can do"` Zeno will sometimes suggest it has access to datasets it does (e.g. atmospheric temp, air quality) and suggest analysis using these datasets or locations it cannot directly analyse (e.g. Amazon Rainforest)

**2. Insight follow-up suggestions**
See: https://github.com/wri/project-zeno/blob/3f0c8a9da6f86611df98e08cb00d91286e6723dc/src/tools/generate_insights.py#L102-L104
Again, zeno sometimes suggests prompts outside of scope.

### Hints
- Self-description could be a fixed response
- Remove insight follow-up

srmsoumya:
Closes #357 

**Changes**

- Added `get_capabilities` tool for factual capability information
- Added capability context to generate_insights.py to prevent hallucination in follow-up suggestions

**Frontend Impact**
- No breaking changes - only adds messages to chat stream
- Better UX - accurate responses to "What can you do?" questions and more realistic follow-up suggestions
- No new UI needed - treat as regular tool message, preferably drop the `tool_message` as its for the agent to interpret & provide a response

Eliminates hallucination when users ask about agent capabilities and improves follow-up suggestion quality.

cc @batpad @kamicut


src/tools/generate_insights.py
```diff
@@ -102,9 +102,11 @@ class InsightResponse(BaseModel):
 1. One chart insight with appropriate chart type, Recharts-compatible data, and clear axis fields
 2. 1 or 2 simple follow ups to the user query based on the actual data available
 
+Your capabilities: You can analyze data for any area of interest, pull data from datasets like DIST-ALERT, Global Land Cover, Tree Cover Loss, and Grasslands for different time periods, and create charts/insights. Base follow-ups on what's actually possible with available data and tools.
```
yellowcap, line 105: This should be dynamic if possible, otherwise it will get out of date as we add more datasets.

src/tools/get_capabilities.py
```diff
@@ -0,0 +1,58 @@
...
+    Use this tool when users ask about what you can do, what data is available, what's possible or about you.
+    """
+    return """
+ABOUT ME:
+Hi there! I'm Global Nature Watch's Geospatial Agent - your friendly AI assistant for exploring our planet's environmental data. I was built with love by the amazing humans and AI at World Resources Institute and Land & Carbon Lab. Think of me as your nature-loving data detective, ready to help you uncover insights about forests, land use, and environmental changes anywhere on Earth!
+
+CORE CAPABILITIES:
+I can find and analyze geospatial data for any area of interest, create interactive maps and charts, generate compact insights, and compare places and trends over time.
+
+AVAILABLE DATASETS:
```
yellowcap, line 17: Also here, this shoudl be dynamic and come from the yaml.

srmsoumya:
Good point 💯

PR #403, opened 2025-09-08
Add user context to agent prompts and improve personalization

srmsoumya:
Enhanced agent prompts to include user profile information: 
- ~name~
- ~expertise level~
- location
- interests
- preferred language
- sector & role

This makes the agent response more personalized.

cc @sunu @batpad

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -308,6 +308,7 @@ async def stream_chat(
     session_id: Optional[str] = None,
     user_id: Optional[str] = None,
     tags: Optional[list] = None,
+    user: Optional[dict] = None,
 ):
     # Populate langfuse metadata
     if metadata:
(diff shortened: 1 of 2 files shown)


src/api/app.py
```diff
@@ -988,6 +989,26 @@ async def chat(
...
+                "first_name": user.first_name,
+                "last_name": user.last_name,
+                "job_title": user.job_title,
+                "company_organization": user.company_organization,
+                "country_code": user.country_code,
+                "preferred_language_code": user.preferred_language_code,
+                "gis_expertise_level": user.gis_expertise_level,
+                "areas_of_interest": user.areas_of_interest,
+                "profile_description": user.profile_description,
+                "sector_code": user.sector_code,
```
batpad, line 1008: Note, the sector_code here will be the shortname that we store in the database - you probably want the sector label to pass to the LLM - see here: https://github.com/wri/project-zeno/blob/main/src/user_profile_configs/sectors.py#L8 - this is where the codes get mapped to labels.

src/api/app.py
```diff
@@ -988,6 +989,26 @@ async def chat(
...
+                "role_code": user.role_code,
```
batpad, line 1009: Same for `role_code`

yellowcap:
Ok after getting input from @01painadam we decided to remove name and job title from the prompt. Let's only put in the interests fields but leave away the PII.

PR #433, opened 2025-09-15
Add unit field to ChartInsight for y-axis display

### LINKED ISSUE #417: Provide units for insight charts
While insight charts do often have reference to the unit and/or include a unit as the yaxis label, we don't explicitly send units in the API. Adding a `unit` string value would ensure that the charts are always rendered with the correct units and are more explicitly understood by end users.

srmsoumya:
Closes #417 

- Added unit field to `ChartInsight` model with examples (hectares, alerts, %, km²)
- Updated chart data generation to include unit in frontend payload
- Enhanced LLM prompt to provide appropriate units for y-axis values

cc @LanesGood @kamicut

### DIFF
--- a/src/tools/generate_insights.py
+++ b/src/tools/generate_insights.py
@@ -61,6 +61,10 @@ class ChartInsight(BaseModel):
     y_axis: str = Field(
         description="Name of the field to use for Y-axis (for applicable chart types)"
     )
+    unit: str = Field(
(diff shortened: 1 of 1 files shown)


yellowcap:
This can be closed because we are working on a refactor of the insights generation. @srmsoumya maybe ensure in the new version that we track units.

PR #437, opened 2025-09-15
Wdpa ingest simplify geom

yellowcap:
This simplifies geometries with an adaptive threshold. The resulting table locally was 1GB @sunu , shall we reduce this more or is that acceptable?

```
zeno-data=# SELECT pg_size_pretty(pg_total_relation_size('geometries_wdpa')) AS "Table Size";
 Table Size 
------------
 1055 MB
(1 row)
```


sunu (approved):
lgtm!

PR #446, opened 2025-09-16
Fix AOI context persistence bug

srmsoumya:
- Updates `AOI` selection to properly clear previous context when new AOI is selected
- Adds custom state reducer to handle AOI updates and ensures UI actions update `aoi_options` state correctly


yellowcap (approved):
🦆 looking good

PR #452, opened 2025-09-16
Remove user context from the prompt

srmsoumya:


yellowcap (approved):
Cool thanks! Slightly sad about this but I think its more efficient and more concise / netural that way.

PR #454, opened 2025-09-16
Fix Issue with daterange mismatch between thread name & analysis

srmsoumya:
- Add current date context to thread name generation prompt to improve relative time query handling (e.g., "past 3 months", "last week")
- Add 50-character truncation with ellipsis for long names


yellowcap (approved):
lgtm

PR #462, opened 2025-09-22
Allow admins to render any thread

yellowcap:
This change allows admin users to see any tread in read-only mode. 

This would help the WRI admin team a lot to understand bug reports and feature requests. The rendered thread is much easier to read than the langfuse trace.


srmsoumya (approved):
lgtm.

PR #466, opened 2025-10-03
Update chart instructions

01painadam:
A couple of instruction changes to address bugs reported by internal researchers. These should _hopefully_ ensure that the LLM doesn't:
- Show emissions and loss on the same bar chart
- Calculate net forest change using TCL and Gain
- use confusing terms for net carbon flux


srmsoumya (approved):
LGTM.

PR #472, opened 2025-10-15
Fix analytics handler for custom aois

yellowcap:
Fix data pull for custom aoi handling.

Also adds a test that calls the analytics api with the custom aoi.

Closes https://github.com/wri/project-zeno-next/issues/346


srmsoumya (approved):
Looks good to merge, tested things on API side.

PR #480, opened 2025-10-29
CodeAct pattern for insight generation with sandboxed code execution

srmsoumya:
This PR refactors the insight generation system to use a CodeAct pattern with Docker-based sandboxed code execution, replacing the previous approach of passing all data directly to the LLM.

### Key Changes

**Architecture**
  - Implement `PersistentPythonSandbox` class for managing stateful Docker sandbox sessions
  - Replace monolithic LLM insight generation with multi-step agentic workflow:
        a. Prepare datasets as CSV files in isolated temp directory
        b. Copy files to sandbox container
        c. Agent executes Python code for data analysis
        d. Retrieve chart data from sandbox
        e. Separate LLM call generates chart schema

**Testing**
  - Add scripts/run_generate_insights_tests.sh to run tests individually (workaround for event loop issues)

**Next Steps**
  - Add guardrails on insight wording and follow-up suggestions
  - Improve CodeAct prompting for better analysis
  - Future PR: Explore native code-execution tools from Gemini/Claude/Groq APIs (current limitation: stateless execution)

cc @sunu @yellowcap

### DIFF
--- a/src/tools/generate_insights.py
+++ b/src/tools/generate_insights.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Annotated, Dict, List
 
 import pandas as pd
-import yaml
 from langchain_core.messages import ToolMessage
 from langchain_core.tools import tool
 from langchain_core.tools.base import InjectedToolCallId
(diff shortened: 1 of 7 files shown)


src/tools/generate_insights.py
```diff
@@ -1,25 +1,184 @@
...
+from src.utils.llms import GEMINI_FLASH
 from src.utils.logging_config import get_logger
 
 logger = get_logger(__name__)
-encoder = tiktoken.get_encoding(
-    "o200k_base"
-)  # tiktoken encoding for OpenAI's GPT-4o, used for approximate token counting
+
+
+class PersistentPythonSandbox:
```
yellowcap: This will require a docker deamon in the k8s deployment. @batpad and @sunu `llm-sandbox` can be pointed to a `DOCKER_HOST` . After a quick lookup it seems like that this should not be too hard. My llm friends sugested something like

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: zeno-api
spec:
  template:
    spec:
      containers:
      - name: api
        image: your-image
        env:
        - name: DOCKER_HOST
          value: "unix:///var/run/docker.sock"  # Points to sidecar socket
        volumeMounts:
        - name: docker-sock
          mountPath: /var/run/docker.sock
      
      - name: docker-daemon
        image: docker:24-dind
        securityContext:
          privileged: true  # Required for DinD
        env:
        - name: DOCKER_TLS_CERTDIR
          value: ""  # Disable TLS for simplicity
        volumeMounts:
        - name: docker-sock
          mountPath: /var/run/docker.sock
      
      volumes:
      - name: docker-sock
        emptyDir: {}  # Shared socket between containers
```

src/utils/llms.py
```diff
@@ -1,6 +1,8 @@
 from dotenv import load_dotenv
 from langchain_anthropic import ChatAnthropic
 from langchain_google_genai import ChatGoogleGenerativeAI
+from langchain_groq import ChatGroq
```
yellowcap: I propose to skip this. We are not likely going to use it, and it gave me an error because i did not have an api key. Change also in review PR.

yellowcap:
I am documenting our decision making process in this [ADR on confluence](https://gfw.atlassian.net/wiki/spaces/~7120207547230a62834f7f988fb2eb3bb65b93/pages/2297528334/CodeAct+ADR)

PR #484, opened 2025-10-30
Gemini temperature 0, fixed thinking budget

yellowcap:
@srmsoumya if these are better lets use them now


srmsoumya:
yes, not sure if they are any better 😄

Issue #487, opened 2025-10-31
Implement Insights CodeAct with gemini code execution


yellowcap:
We have a first  implementaiton for a code-act pattern for insights using llm-sandbox #480.

### Don't build more infrastructure just yet

However, wihile working on this we realized that this would require additional infrastructure in  deployment on Kubernetes . Running docker containers is entirely possible, but requires setting up new services.

### Focus on gemini code executer next

This is to propose that we use the gemini native code execution so that we can test this on staging without any infrastructure or frontend changes. 

### Keep llm-sandbox around for later

We can hopefully put that into the code as an option alongside the already existing llm sandbox implementation.

Please chime in on this discussion @srmsoumya @batpad @sunu @geohacker.

batpad:
@yellowcap thanks for ticketing - this is exciting work!

It makes sense to me to use the gemini native code execution to test things out on staging. I do think this is something that's useful to figure out doing properly on the Kubernetes setup so we're not dependent on external services and their limitations.

I do have a few concerns about how to do this "properly" on production - this will likely require @sunu's time to dig into a bit.

yellowcap:
I think using native code executers per model provide is acutally a proper implementationas well, its just more coding work to adapt to new providers. So if we do the gemini one well, we might just stick to it even in production. We can then add support for other providers as well. The "reusability" tradeoff here is having to choose a supported vendor vs building our own code execution infra.

srmsoumya:
@yellowcap thank you for putting together the [ADR](https://gfw.atlassian.net/wiki/spaces/LCL/pages/2297528334/CodeAct+for+Insights+ADR) on this.

Let's move forward with implementing the CodeAct pattern using Gemini for now - this will help WRI make GNW publicly available soon. In parallel, we can explore alternative options like Pyodide and container sandboxes on the side - this will be useful for other projects as well, as data analysis is central to what we are doing with agents.

@sunu I noticed that [llm-sandbox also has a K8s backend option](https://vndee.github.io/llm-sandbox/backends/#kubernetes-backend) - would appreciate your thoughts on it when you get a chance (no rush on this).

PR #491, opened 2025-11-05
Implement CodeAct pattern with Gemini native code execution

srmsoumya:
Invalidates #480 Addresses #487

Replaces prompt-based LLM insight generation with CodeAct using Gemini's native code executor.  

### Changes
- New `GeminiCodeExecutor` for stateless code execution with inline CSV data
- Agent explores datasets, writes Python code to analyze data, and generates chart outputs
- Updated graph state to track code execution details
- Added frontend UI support in streamlit to display code blocks and execution outputs 

All tests for generate-insights are passing & I have tested a few queries on streamlit. This will need a bit of prompt engineering to handle all edge-cases before we can deploy to production.

cc @sunu @batpad @geohacker

### DIFF
--- a/src/tools/generate_insights.py
+++ b/src/tools/generate_insights.py
@@ -149,7 +149,6 @@ Now prepare the data for visualization in Recharts.js:
 - Base this on the actual numbers and patterns you found
 """
 
-    logger.info(f"Analysis prompt:\n{prompt}")
     return prompt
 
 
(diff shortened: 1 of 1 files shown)


src/tools/generate_insights.py
```diff
@@ -228,143 +179,158 @@ async def generate_insights(
...
+Generate ALL content (insights, titles, follow-ups) in the SAME LANGUAGE as the user query.
+Dataset specific guidelines: {dataset_guidelines}
+Dataset specific cautions: {dataset_cautions}
 
-        # Add follow-up suggestions
-        message_parts.append("Follow-up suggestions:")
-        for i, suggestion in enumerate(follow_ups, 1):
-            message_parts.append(f"{i}. {suggestion}")
+#### Capability Context
+You can analyze data for any area of interest, pull data from datasets like {available_datasets} for different time periods, and create various charts/insights. Base follow-ups on what's actually possible with available data and tools.
```
yellowcap: this does not make sense to me at this point. We are not pulling data here at all.

yellowcap (commented):
A few comments from initial read, I'll test this locally now.

PR #493, opened 2025-11-06
Add class grouping instructions

01painadam:
Add common Driver Class groupings to analysis instructions. Should enable Zeno to correctly group drivers of deforestation (vs. Tree Cover Loss), human drivers vs. natural.


yellowcap (changes requested):
Looks good except for the variable threshold.

PR #495, opened 2025-11-10
Feature/pro user type

batpad:
Towards implementing a Pro user type with higher prompt limit, but no "admin" privileges.


batpad:
> user_type is a full text field in the database so no new migrations are required

Haa, realized that after I generated migrations and saw they do nothing :D - removed them.

> cli.py operates directly on the database so actions are protected by access to IAM credentials for accessing database secrets (just need to be careful not to commit or leak those otherwise)

Yup - this currently requires connecting to the kubernetes cluster to run, so all those creds are managed separately and securely.

Thanks for the review!

PR #500, opened 2025-11-19
Gemini instructions to be proactive and use cautious words.

### LINKED ISSUE #490: Make Gemini more proactive/greedy in conversation
### What?

Update instructions so that Gemini will make its best attempt to generate analysis based on a user prompt. Where it does so, it should inform the user what choices it made and why. 

### Why?

Gemini often requests user clarification or consent before proceeding. This ultimately uses up a lot of user prompts, and also makes it more time consuming to arrive at a result.

#### Include more caveats

Also, Gemini is using the correct words but does not warn or advise users about the difference about the difference.

> Did some testing with the in staging after the new updates with Gemini and unfortunately it is no longer providing the caveats up front with 'deforestation' vs. 'tcl' like it previously was during these tests. Flagging this as I wonder if switching to Gemini requires any tweaks or adjustments to pull in the cautions as an explicit step. Let me know if you want me to report this as a bug.

We need to ensure that caveats/limitations are being clearly surfaced.

#### Use more cautious language

Based on feedback from researchers, we should include instructions on using more cautioius wording.

yellowcap:
Added instructions for being more proactive and improve wording.

This worked in quick tests, but requires further experimentation to ensure it works in more cases.

Closes #490

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -464,8 +464,6 @@ class AgentState(TypedDict):
     follow_up_suggestions: list
     # Chart data for frontend rendering
     charts_data: list
-    # Always 1 (single insight)
-    insight_count: int
 ```
 
 #### Advanced Chart Fields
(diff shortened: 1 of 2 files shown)


yellowcap:
Closing this in favor of #503

Issue #501, opened 2025-11-19
Test Gemini 3


01painadam:
Gemini 3 is out - can we get it to work with new system?

Output: either get it to work and document required changes or, ticket out steps to implement it

PR #503, opened 2025-11-19
Support gemini 3 and improve wording

### LINKED ISSUE #490: Make Gemini more proactive/greedy in conversation

yellowcap:
## What I am changing

- This adds support for gemini-3-pro.
  - To support gemini-3, the `langchain-google-genai` library needed upgrade.
  - That triggered a chain of updates on the langchain side. 
- Removed a stale state variable `insight_count`, it was not in use
- Improved prompts by adds instructions for more cautious wording.

I did a good amount of testing, both in unit tests and through the nextjs app and a localhost api.

## How to test it

Run unit tests, and spin up streamlit and / or nextjs locally with local api.

Closes #490 and #501

### DIFF
--- a/src/agents/agents.py
+++ b/src/agents/agents.py
@@ -8,6 +8,7 @@ from langgraph.prebuilt import create_react_agent
 from psycopg.rows import dict_row
 from psycopg_pool import AsyncConnectionPool
 
+from src.agents.prompts import WORDING_INSTRUCTIONS
 from src.graph import AgentState
 from src.tools import (
     generate_insights,
@@ -30,6 +31,7 @@ CRITICAL INSTRUCTIONS:
 - Be PROACTIVE in tool calling, do not ask for clarification or user input unless you absolutely need it.
   For instance, if dates, places, or datasets dont match exactly, warn the user but move forward with the analysis.,
 - You MUST call tools sequentially, never in parallel. No parallel tool calling allowed.
+- Provide intermediate messages between tool calls to the user to keep them updated on the progress of the analysis.
 
 TOOLS:
 - pick_aoi: Pick the best area of interest (AOI) based on a place name and user's question.
@@ -99,12 +101,7 @@ GENERAL NOTES:
 - Use markdown formatting for giving structure and increase readability of your response. Include empty lines between sections and paragraphs to improve readability.
 - Never include json data or code blocks in your response. The data is rendered from the state updates directly, separately from your own response.
 
-NOTES ABOUT WORDING:
-- Proactively use the information on cautions and limitations in your response, but keep explanations short and concise. For instance, if a user asks about deforestation, clarify the difference between deforestation and tree cover loss.
-- Avoid using strong statements in your anwswers.
-    - Avoid words like: overwheliming, severe, exceptional, critical, concerning, highly, substantial, considerable, notable, remarkable, important, major, crucial, key, strong, robust, dramatic, meaningful (vague unless defined), alarming, worrying, problematic, challenging, unfavorable, promising, encouraging, favorable
-    - Use neutral, measurement-first words: decline, decrease, increase, remain stable, fluctuate,
-- Other words that need scientific justification and actual tests when used: trend (when trend wasn't actually calculated), significant (when not tied to statistical significance), validated (when not actually measured), accurate (without comparison or error bars)
+{WORDING_INSTRUCTIONS}
 """
 
 
--- a/src/tools/generate_insights.py
+++ b/src/tools/generate_insights.py
@@ -8,6 +8,7 @@ from langgraph.prebuilt import InjectedState
 from langgraph.types import Command
 from pydantic import BaseModel, Field
 
+from src.agents.prompts import WORDING_INSTRUCTIONS
 from src.tools.code_executors import GeminiCodeExecutor
 from src.tools.datasets_config import DATASETS
 from src.utils.llms import GEMINI_FLASH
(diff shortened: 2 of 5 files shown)


src/agents/agents.py
```diff
@@ -25,8 +25,10 @@ def get_prompt(user: Optional[dict] = None) -> str:
     """Generate the prompt with current date. (Ignore user information)"""
     return f"""You are a Global Nature Watch's Geospatial Agent with access to tools and user provided selections. Think step-by-step to help answer user queries.
 
-CRITICAL INSTRUCTION: You ALWAYS need AOI + dataset + date range to perform analysis. If ANY are missing or unclear, ask for clarification.
-CRITICAL INSTRUCTION: You MUST call tools sequentially, never in parallel. No parallel tool calling allowed.
+CRITICAL INSTRUCTIONS:
+- You ALWAYS need AOI + dataset + date range to perform analysis. If ANY are missing or unclear, ask for clarification.
+- You MUST call tools sequentially, never in parallel. No parallel tool calling allowed.
+- Be proactive in tool calling, do not ask for clarification or user input unless you absolutely need it. For instance, if dates, places, or datasets dont match exactly, warn the user but move forward with the analysis.,
```
srmsoumya: This goes against the first instruction - so may be reword the first instruction to something around:
- You ALWAYS need AOI + dataset + date range for analysis: if any are missing or unclear, ask for clarification; but if they are present with minor inconsistencies, proceed proactively and include a warning about the assumptions made.

src/utils/llms.py
```diff
@@ -21,16 +21,16 @@
...
 # Google
 GEMINI = ChatGoogleGenerativeAI(
-    model="gemini-2.5-pro",
-    temperature=0.0,
-    max_tokens=None,  # max_tokens=None means no limit
-    include_thoughts=False,
+    model="gemini-3-pro-preview",
+    temperature=1.0,
+    max_tokens=None,
+    include_thoughts=True,
```
srmsoumya: We are not using thoughts in FE - keeping this False, keeps the API consistent (although, I think we have filter set in FE for thinking tokens)

src/utils/llms.py
```diff
@@ -21,16 +21,16 @@
 
 # Google
 GEMINI = ChatGoogleGenerativeAI(
-    model="gemini-2.5-pro",
-    temperature=0.0,
-    max_tokens=None,  # max_tokens=None means no limit
-    include_thoughts=False,
+    model="gemini-3-pro-preview",
+    temperature=1.0,
+    max_tokens=None,
```
srmsoumya: maybe keep the comment?

srmsoumya (approved):
LGTM - added a few comments.

src/agents/agents.py
```diff
@@ -92,6 +94,16 @@ def get_prompt(user: Optional[dict] = None) -> str:
     - "What place in Eastern Europe has the most ecosystem disturbance alerts?"
 - Always reply in the same language that the user is using in their query.
 - Current date is {datetime.now().strftime("%Y-%m-%d")}. Use this for relative time queries like "past 3 months", "last week", etc.
+- If insigths provide them, include follow-up suggestions for further exploration.
```
01painadam: insights

src/tools/generate_insights.py
```diff
@@ -351,6 +351,14 @@ async def generate_insights(
...
 5. **Examples for follow-up suggestions**: "Show trend over different period", "Compare with nearby area", "Identify top performers", "Break down by category"
+
+Notes about wording:
+- Proactively use the information on cautions in your response, but keep explanations short and concise. For instance, if a user asks about deforestation, clarify the difference between deforestation and tree cover loss.
+- Avoid using strong statements in your anwswers.
+    - Avoid words like: overwheliming, severe, exceptional, critical, concerning, highly, substantial, considerable, notable, remarkable, important, major, crucial, key, strong, robust, dramatic, meaningful (vague unless defined), alarming, worrying, problematic, challenging, unfavorable, promising, encouraging, favorable
+    - Use neutral, measurement-first words: decline, decrease, increase, remain stable, fluctuate,
+- Other words that need scientific justification and actual tests when used: trend (when trend wasn't actually calculated), significant (when not tied to statistical significance), validated (when not actually measured), accurate (without comparison or error bars)
```
01painadam: indent - avoid these words

src/agents/agents.py
yellowcap: @srmsoumya, I updated the prompt. I am changing it to say ask to specify if something is _missing_. Otherwise be proactive.

@01painadam goot idea, but that is not possible at the moment, because dates is an input to dataset picker. So the agent has not seen the yaml before that. We can revise this as a refinement separately.

PR #504, opened 2025-11-20
Improve machine user validation

yellowcap:
Better validation of topics list in machine user.

I had a case where the content was str I think and it was breaking the validator. This made it work for me.


yellowcap:
Not needed at the moment. Can revisit if we think its still useful later.

Issue #558, opened 2026-02-13
Fix date range handling in pull data


yellowcap:
The pull data functionality does not correctly check for available data ranges. This leads to confusing and inconsistent date ranges on insights.

https://gfw.atlassian.net/browse/PZB-614

yellowcap:
Done in #559

Issue #447, opened 2025-09-16
Bug: Incorrect date range in thread names


srmsoumya:
For relative time based queries, like past 5 years, thread names have a different date range compared to date range used in actual analysis.

<img width="581" height="669" alt="Image" src="https://github.com/user-attachments/assets/d0e8d514-c2f1-4e14-9f93-cbcd58d6a081" />

PR #559, opened 2026-02-13
Fix date range handling

yellowcap:
## What I am fixing

1. The pull data functionality did not correctly check for available data ranges. This lead to confusing and inconsistent date ranges reported on insights. Even if the data was correct.
2. The thread naming was using the user query to set a date, but that can also have a mismatch with the available data.

Closes
-  #447
- #558

## How I did it

1. I added a check to "clamp" date ranges to the available date range, and inform the user about it.
2. Instructed the thread naming to not use any dates, this is the only way we can ensure there is not a mismatch

I also added a small improvement by moving a class declaration out of a function. That is an unrelated cleanup.

## How to test it

There is a regression test for the use case of TCL, and unit tests for the new function.

### DIFF
--- a/tests/tools/test_pull_data.py
+++ b/tests/tools/test_pull_data.py
@@ -1,9 +1,11 @@
 import uuid
+from datetime import datetime, timedelta
 
 import pytest
 import structlog
 from sqlalchemy import select
 
+from src.agent.tools.datasets_config import DATASETS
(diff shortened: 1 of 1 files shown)


yellowcap:
The failing agent test passes for me locally. Not sure what is going on in the CI. Merging to move this forward.

PR #581, opened 2026-03-24
Remove analysis limit

01painadam:
In prep for the next release with multiarea selection - remove the limit on analysable GAMD subareas.

Enables prompts like: "Which US state has the highest cropland+cultivated grassland : built-up area ratio?, which should select 51 areas (50 states + DC)


jterry64:
I think we also decided we should probably put some basic limit up on the analytics side, just in case it goes crazy and sends all admin 2s or something.

PR #583, opened 2026-03-24
Replace context_layers with config-driven params aligned to analytics API

yellowcap:
## Summary

Replaces the `context_layers` abstraction with a generic, config-driven `params` system that mirrors the actual filter parameters accepted by the [analytics API](https://analytics.globalnaturewatch.org/openapi.json). Previously, each dataset had bespoke handling in Python code -- hardcoded dataset ID constants, per-dataset `if/elif` chains for payload construction, and a `context_layer` field that flattened fundamentally different API parameters (intersections, canopy cover, forest filters, crop types, gas types) into a single string. This PR moves all of that into declarative `analytics_config` blocks in each dataset's YAML file, and makes the handler logic fully data-driven.

### What changed

- **Dataset YAML configs** now declare `analytics_config` with:
  - `date_params` -- how date ranges map to API fields (`date`, `year`, `year_5yr_interval`, or `none`)
  - `params` -- typed parameter definitions with allowed values, defaults, and descriptions (e.g. `canopy_cover`, `forest_filter`, `intersections`, `crop_types`, `gas_types`)
  - `aoi_restrictions` -- allowed AOI subtypes per dataset (replaces hardcoded sLUC GADM check)
  - `alternate_endpoints` -- e.g. land cover composition endpoint for change-over-time queries
  - `tile_url_config` -- how tile URLs are constructed (`append_date_range`, `format_year`, `append_year_range`, or `none`)

- **`analytics_handler.py`**: removed ~200 lines of hardcoded dataset ID constants and per-dataset `if/elif` branches. `can_handle()` now checks for the presence of `analytics_api_endpoint` instead of matching against a list of IDs. Payload construction iterates over config-defined params.

- **`pick_dataset.py`**: the LLM now selects `params: dict` instead of `context_layer: str`. A new `_resolve_params()` function validates LLM-selected values against allowed values and falls back to defaults for invalid/hallucinated selections. The selection prompt shows `available_params` per candidate dataset so the LLM can make informed choices.

- **`generate_insights.py`**: passes `active_params` dict instead of a single `context_layer` string to the code executor prompt.

- **`get_capabilities.py`** / **`embed_datasets.py`**: updated to read from `analytics_config.params` instead of `context_layers`.

- **Frontend**: `context_layer` references replaced with `params` dict in the evaluation page, map renderer, and sidebar fixtures.

- **Docs**: updated state schema examples in `docs/README.md`.

### Known limitation

Tree cover loss (dataset 4) and tree cover loss by dominant driver (dataset 8) hit the same API endpoint (`/v0/land_change/tree_cover_loss/analytics`) with different filter configurations. Ideally these would be separate API endpoints since they have quite different semantics (e.g. time-slice coverage). This works correctly with the current API, but is a candidate for a future API review.

### Smaller changes

- **Gemini executor**: added null check for `code_execution_result.output` to avoid passing empty output parts.
- **`datasets_config.py`**: minor lint fix (`!=` instead of `not len(...) ==`).

## Separate: CI workflow for tools tests

The tools-tests workflow can now be triggered on PRs by adding the `run-tools-tests` label (in addition to the existing `workflow_dispatch` trigger). Includes concurrency control to cancel in-progress runs.

### DIFF
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -46,15 +46,6 @@ jobs:
       - name: Install dependencies
         run: uv sync --frozen
 
-      - name: Generate dataset embeddings
-        env:
-          GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
-        run: |
-          mkdir -p data
-          uv run python src/ingest/embed_datasets.py
-
       - name: Run tools tests
         env:
           DATABASE_URL: postgresql+asyncpg://postgres:postgres@localhost:5432/zeno-data
@@ -82,4 +73,6 @@ jobs:
           COOKIE_SIGNER_SECRET_KEY: fake-cookie-key
           NEXTJS_API_KEY: test-nextjs-api-key
         run: |
+          mkdir -p data
+          uv run python src/ingest/embed_datasets.py
           uv run pytest tests/tools -v
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -2,9 +2,18 @@ name: Tools Tests
 
 on:
   workflow_dispatch:
+  pull_request:
+    types: [labeled]
+
+concurrency:
+  group: tools-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   test:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'run-tools-tests'))
     runs-on: ubuntu-latest
 
     services:
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -49,6 +49,8 @@ jobs:
       - name: Generate dataset embeddings
         env:
           GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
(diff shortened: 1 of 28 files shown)


jterry64 (changes requested):
I think this is overall looks good, just some minor comments on the actual code. I would say going forward, I'd be a bit worried about the YAML getting out of sync with the API. If we take this approach of having the LLM choose all the parameters, once we move the analytics API into here, we might want to just based it off of common pydantic models and automatically generate these analytics_info configs from those.

Alternatively, I was thinking about if it could also make sense to go in the opposite direction. Since we're already using CodeAct/pandas at the end, instead of making the LLM choose the exact stats it needs up front, could we just give the LLM excess data and have it decide with pandas how to best answer the user question? E.g. does it need to select a specific tree cover density threshold, or could we just return the results grouped by each threshold, and the LLM applies a threshold filter if necessary to answer the user question in pandas? Similarly, if a user ask about primary deforestation from fires, it simply needs to select a list like "tree cover loss, fires, primary forest" and the tool can just deterministically pull all data that might be relevant to those topics in the AOI. E.g.  instead of just providing tree cover loss filtered by primary forest and fire, it might pull additional data like VIIRS fire alerts, TCL drivers from wildfires (that's a separate dataset), tree cover extent, total TCL. This could also allow for more interesting insights like "X% of TCL was from fires from any cause, and X% were specifically from wildfires. An uptick in recent fire alerts indicates this is still an active problem". 

This would simplify what the LLM needs to ask for when pulling data, give it more flexibility to answer questions, and possibly allow it to further drill into data it already has on follow up questions. But then I guess we're just moving all the complexity from pulling data to generating insights.

I'll be curious to see how good the LLM is at selecting from all these parameters, but if it can do that pretty well, then this seems like the correct approach.

yellowcap:
Not needed @jterry64 added parameters in #631

PR #597, opened 2026-03-30
Dynamic canopy cover threshold selection  

01painadam:
## Summary
                                                                                                 
The agent previously used a hardcoded 30% GFW default canopy cover threshold for all tree cover  
analyses, regardless of country context. This PR makes threshold selection dynamic and proactive: the agent
infers the correct national forest definition from context, applies it to the data fetch, and    
cites the authoritative source with a link in all outputs.

## What changed                                                                                       
Users asking about tree cover in a specific country now get analysis using that country's official forest definition without having to specify the threshold themselves. These are states in the agent instructiosn (see citations below)/.
                                                                                                   
  > "Show tree cover loss in India from 2015 to 2022" → fetches and analyses at 10% (FSI national  
  definition), not 30%
> "Tree cover loss in Australia" → 20% (ABARES)                                                  
> "Forest loss in Chile" → 25% (CONAF)                                                           
> "Tree cover loss in Colombia" → 30% (IDEAM/UNFCCC REDD+)
                                                                                                   
- Added a canopy_cover parameter to the pull_data tool (LLM-visible), threaded through           
DataPullOrchestrator → AnalyticsHandler → _build_payload()                                       
- Added a TREE COVER THRESHOLD SELECTION block to the system prompt with a country-to-threshold  
lookup table covering ~40 countries/regions               
- Forest GHG net flux remains hardcoded at 30% (API constraint)                                                                                
 
## Citations                                                                                        
Every threshold statement in the response and chart output now includes a named agency and a     
markdown link to the official definition:
                                                                                               
> 10% — India's national forest definition per the Forest Survey of India (FSI)                  
> 30% — Colombia's national definition per IDEAM / UNFCCC REDD+ submission
> 30% — GFW default; no country-specific definition applied                                      
                                                                                                 
## Tests                                                                                            
- 33 unit tests in test_canopy_cover_threshold.py covering payload construction, default  propagation, and prompt override injection                                                       
- Evals in test_threshold_citations.py and test_threshold_presentation.py checking LLM output
quality for threshold selection and citation formatting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -194,3 +194,4 @@ logs/
 
 # Notebooks
 nbs/
+tests/evals/results/
--- a/docs/adr-canopy-cover-threshold-resolution.md
+++ b/docs/adr-canopy-cover-threshold-resolution.md
@@ -1,0 +1,110 @@
+# ADR: Deterministic Canopy Cover Threshold Resolution
+
+**Date:** 2026-04-06
+**Status:** Accepted
+**Branch:** feat/canopy-cover-threshold
+**PR:** wri/project-zeno#597
+
+---
+
+## Context
+
+Tree cover loss datasets on Global Forest Watch support a `canopy_cover` threshold
+parameter (valid values: 10, 15, 20, 25, 30, 50, 75%) that filters which pixels
+are counted as "forest". Different countries define forest using different minimum
+canopy densities in their national legislation and UNFCCC submissions.
+
+The previous implementation encoded the country→threshold mapping as a large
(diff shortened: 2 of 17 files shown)


srmsoumya (approved):
This looks great overall! One thing worth considering - adding dataset or task-specific constraints directly into the base system prompt could gradually clutter the agent's memory. It also has a cost implication, since those extra tokens are sent with every query.

It might be worth exploring whether we could move task and data-specific logic outside of the base configuration, keeping the GNW agent's core simple and focused. Skills might be a good fit for handling that kind of specialization.

jterry64 (changes requested):

PR #598, opened 2026-03-30
Compact console logs, full details in file

srmsoumya:
Restructured `structlog` to use per-handler `ProcessorFormatter`. Console now strips `query, request_id, session_id, thread_id, user_id` and shortens logger names when `LOG_VERBOSE=false`. File handler keeps everything.


sunu (commented):
@srmsoumya I've added a couple of minor comments. But otherwise this looks good to me! 

I like that this doesn't require any deployment changes to keep the existing logging behavior for deployed instances.

PR #600, opened 2026-03-30
[PZB-682] Change AOI limits to 1000 for admin subregions, 50 for all other

jterry64:
- Change limits on AOIs to match Analytics API limit: 1000 for admin areas, 50 for all other AOI types
- Add unit tests without dependencies to check limits are applied. I separated these out to their own files, following FastAPI guidance to make your test file structure match your app file structure, and our internal standard of creating folders for each part of the test triangle.
- I made sure the unit tests were called in our test CI github action.
- I removed the evals you added in your PR @01painadam, which is why I just made a separate branch/PR so I don't delete that code. It seemed to get into an infinite loop for me, where it'd keep getting a failure on the generate insights and then retry. For now, I think this is sufficient; I paired with @gtempus, and we agreed we should see how far we can go with unit testing our own code with no dependencies, and hopefully only require a final eval to test the LLM. What I'm going to try here is merging this PR, going to staging, add a single eval on gnw-evals to test this scenario, and then seeing what it catches. Then I'll circle back, see if those issues could've been caught through unit tests, or if it required a test using the LLM. Then we can shake out what's really testable without dependencies, and where we need evals using dependencies.

### DIFF
--- a/tests/unit/agent/tools/test_pick_aoi.py
+++ b/tests/unit/agent/tools/test_pick_aoi.py
@@ -22,14 +22,14 @@ def _make_aois(source: str, n: int) -> list[dict]:
     ]
 
 
-async def test_check_aoi_selection_gadm_within_limit():
+async def test_check_aoi_selection_returns_none_when_gadm_within_limit():
     result = await check_aoi_selection(
         _make_aois("gadm", SUBREGION_LIMIT_ADMIN)
     )
     assert result is None
 
 
-async def test_check_aoi_selection_gadm_exceeds_limit():
+async def test_check_aoi_selection_returns_error_message_when_gadm_exceeds_limit():
     aois = _make_aois("gadm", SUBREGION_LIMIT_ADMIN + 1)
     result = await check_aoi_selection(aois)
     assert result is not None


src/agent/tools/pick_aoi.py
```diff
@@ -397,29 +397,23 @@ async def check_multiple_matches(
...
+        subregion_limit = SUBREGION_LIMIT_ADMIN
 
     if len(aois) > subregion_limit:
         return (
             f"Found {len(aois)} subregions, which is too many to process efficiently. "
             "Please narrow down your search by either:\n"
             "1. Being more specific with the AOI selection (choose a smaller area)\n"
             "2. Being more specific with the subregion query (e.g., 'kbas' instead of 'areas')\n"
-            "For optimal performance, please limit results to under 25 subregions for KBA, WDPA, and Indigenous Lands, or under 50 for other area types."
+            f"For optimal performance, please limit results to under {SUBREGION_LIMIT} subregions for KBA, WDPA, and Indigenous Lands, or under {SUBREGION_LIMIT_ADMIN} for other area types."
```
gtempus: Thanks for converting to f-string. 👍

tests/unit/agent/tools/test_pick_aoi.py
```diff
@@ -0,0 +1,62 @@
...
+            "source": source,
+            "src_id": f"{source}_{i}",
+            "name": f"Area {i}",
+            "subtype": "state-province",
+        }
+        for i in range(n)
+    ]
+
+
+async def test_check_aoi_selection_gadm_within_limit():
```
gtempus: I would add some additional info to the test name. Make it easy to know what the problem is without having to read the test code, if at all possible.
```suggestion
async def test_check_aoi_selection_returns_none_when_gadm_within_limit():
```

PR #608, opened 2026-04-04
Refactor pick dataset

jterry64:
Before doing further changes to pick dataset, doing some refactoring and adding unit/integration tests so there are more targeted tests than full tool tests:
- Moved out of the functions to get datasets candidate from RAG and selecting the best dataset to their own classes so they can be mocked and individually tested. Made integration tests just for these LLM flows.
- Moved out of model classes to their own files and made unit tests for them.
- Made unit tests for pick_dataset to just test logic without LLM.


srmsoumya (commented):
LGTM, this is a solid refactor - clean separation of concerns and much better testability.                                                                                                                           
                                                                                                                                                                                                                    Small thing - test case for ("Tree cover loss in primary forest", 4, "primary_forest") in `test_pick_dataset.py` expects `primary_forest` as a context layer, but tree_cover_loss.yml has `context_layers: null` for now - thats causing test to fail.

01painadam (changes requested):
`dataset["context_layer"]` is raising a KeyError locally when `context_layer` is absent from the dataset dict. Requires a small change.

I'm also unable to get Zeno to select primary forest since the yaml still uses `variables` (not `context_layer`) - this might be intentionally left out of this PR, tho - just flagging

yellowcap (changes requested):
Thanks @jterry64 I like that you are giving the tool more structure and move things into submodules, that makes sense to me.

The actual business logic changes look quite simple and are probably ok. I would like to ask for a few changes and then would run the tools tests.

As you are adding a couple of new folders and modules, lets put all the dataset_picker logic under one subfolder and structure things there. We are currently mixing it a little bit, but with this additional structure I would say its worth packing tools into their subfolder, i.e. put all dataset picker stuff in `src/agent/tools/pick_datset/...`

A comment on testing: looks like you created a lot of new tests that increase coverage but are super granular and maybe llm generated? I am not a huge fan of these because a lot of them test pretty obvious stuff that is covered already by slightly higher level tests.

Two other small general comments:

- I am not a fan of the folder name `sub_llm_handlers`, its not self explanatory, that should be more verbose or modules could live at the root.
- Remove all the empty `__init__.py` files, they are not needed in python3
- Add the new filters to the existing test fixtures too [async def test_query_with_context_layer(](https://github.com/wri/project-zeno/blob/43686470f159af32014219bc0392e77608b8aecc/tests/tools/test_pick_dataset.py#L408)

jterry64:
@yellowcap I'm going to leave this open for the retro, I'm noticing from this there's pretty different engineering styles between DevSeed and the WRI team. This PR is a good example, let's discuss our different engineering standards and what we move towards now that the WRI team will be more involved. cc @gtempus

gtempus:
Good discussion @yellowcap and @jterry64. I agree that we'll be needing a coding standard and other general engineering discussions. I've just skimmed over this PR and also #609. There are enough items just in those two PRs to keep us well stocked with discussion.

There is a stark contrast between our repos, and we'll need to get to a more unified approach. We can't do it all at once, but we shouldn't let the discrepancies continue to impose excessive cognitive load on the team. We'll just wind up slowing down or retreating to silos if we do.

PR #610, opened 2026-04-07
Enable global queries

yellowcap:
This change adds support for global or world-wide queries (e.g., "globally", "all countries"). It handles that as a special case.

@kamicut and I discussed the option to generalize this more, but decided to solve this one first with minimal changes and then expand. On the frontend, showing many geometries at once is a challenge, as currently it loads AOIs as geojson one by one. So a generalized solution should be thought through more carefully. See https://github.com/wri/project-zeno-next/issues/448

## Summary of changes:

- pick_aoi now detects global synonyms (global, world, worldwide, earth, globe) early and passes such queries to a dedicated handler—no spatial DB lookup or synthetic row required.
- Global queries only support subregion="country": all countries are returned using a simple WHERE subtype = 'country' query, bypassing any use of ST_CoveredBy.
- All global query logic is moved to a new submodule: pick_aoi/global_queries.py (containing is_global_request, handle_global_request, and _query_all_countries). The main tool function delegates to handle_global_request as needed.
- pick_aoi.py and global_queries.py are reorganized into the pick_aoi/ package (including __init__.py, tool.py, global_queries.py, and selection_name_util.py). The public import interface remains unchanged.


srmsoumya (approved):

src/agent/tools/pick_aoi/global_queries.py
```diff
@@ -0,0 +1,97 @@
...
+                ToolMessage(
+                    "Selected all countries in the world",
+                    tool_call_id=tool_call_id,
+                )
+            ],
+        },
+    )
+
+
+async def _query_all_countries() -> pd.DataFrame:
```
jterry64, line 86: @yellowcap doesn't need to happen now, but in general, we should clean up these direct DB/SQL calls and try to keep our DB dependencies in one place.

src/agent/graph.py
```diff
@@ -74,6 +74,8 @@ def get_prompt(user: Optional[dict] = None) -> str:
 - landmark: Indigenous and community lands (tribal territories, community forests)
 
 Examples of when to USE subregion:
+- "Which countries have the most deforestation globally?" → place="Global World", subregion="country"
+- "Compare forest loss across all countries in the world" → place="Global World", subregion="country"
```
sunu, line 78: this is telling the llm to pass place="Gobal World" but "Global World" is not in `GLOBAL_TRIGGER_WORDS` (which contains "global", "world", etc. individually). So if the llm sets place="Global World", the resulting place will fail the `is_global_request` check.

tests/tools/test_pick_aoi.py
```diff
@@ -251,3 +251,116 @@ async def fake_query_subregion_database(
...
+            },
+            "id": str(uuid.uuid4()),
+            "type": "tool_call",
+        }
+    )
+
+    aois = command.update.get("aoi_selection", {}).get("aois")
+    assert aois is not None
+    assert len(aois) == 3
+    assert all(aoi["subtype"] == "country" for aoi in aois)
```
sunu, line 312: don't think this test is calling `handle_global_request` at all. They're silently going through the "normal" aoi picking flow, not the global one because place name "Global World" doesn't trigger `is_global_request`.

tests/tools/test_pick_aoi.py
```diff
@@ -251,3 +251,116 @@ async def fake_query_subregion_database(
...
+            "args": {
+                "question": "What is the deforestation rate in the world?",
+                "places": ["Global World"],
+            },
+            "id": str(uuid.uuid4()),
+            "type": "tool_call",
+        }
+    )
+
+    assert "aoi_selection" not in command.update
```
sunu, line 336: Same problem as above. This assert fails:
```
FAILED tests/tools/test_pick_aoi.py::test_global_query_without_subregion_is_rejected - AssertionError: assert 'aoi_selection' not in {'aoi': {'gadm_id': 'GLB', 'name': 'Global World', 'source': 'gadm', 'src_id': 'GLB', ...}, 'aoi_selection': {'aois': ...e(content='Selected AOIs:\n- ...
```

src/agent/tools/pick_aoi/global_queries.py
```diff
@@ -0,0 +1,97 @@
...
+    for aoi in final_aois:
+        aoi[SOURCE_ID_MAPPING[aoi["source"]]["id_column"]] = aoi["src_id"]
+
+    return Command(
+        update={
+            "aoi_selection": {
+                "name": "All countries in the world",
+                "aois": final_aois,
+            },
+            "aoi": final_aois[0],
```
sunu, line 74: Does this need a guard in case `final_aois` is empty for some reason?

src/agent/tools/pick_aoi/tool.py
```diff
@@ -280,7 +284,7 @@ async def query_subregion_database(
     SELECT t.name, t.subtype, t.{src_id_field}, '{subregion_source}' as source, t.{src_id_field} as src_id
     FROM {table_name} AS t, aoi
     WHERE t.subtype = :subtype
-    AND ST_Within(t.geometry, aoi.geom)
+    AND ST_CoveredBy(t.geometry, aoi.geom)
```
sunu, line 294: 👍🏽

sunu (changes requested):
@yellowcap This looks great overall! 

There is a phrasing inconsistency for "global" place that will prevent this from working reliably and also breaks integration tests. Other than that everything looks good to me!

jterry64:
@yellowcap everything seems oriented towards ranking/comparing countries. FYI another common thing people will want is just global totals (e.g. "total global tree cover loss in 2025"). Will this work out of the box with these changes, or does `generate_insights` also need an update to understand "global" means "aggregate all countries"? This can also be a follow up PR.

yellowcap:
thanks @sunu for the review great coments will integrate. 

@jterry64 this is just to select the coutnries, so it will then pull data for all of these, and it would be up to the code-act insights part to sum to totals. So yes this is covered in principle, but something to make sure works for sure!

src/agent/graph.py
yellowcap: I fixed that by doing the string matching the other way around, i.e. look for `GLOBAL_TRIGGER_WORDS` in place names, and not the other way around.

src/agent/tools/pick_aoi/global_queries.py
yellowcap: in this case we are forcing it to select all countries, so its fine here. The `aoi` key is deprecated anyway, I think we can remove it in a future update. Ticketed here https://github.com/wri/project-zeno/issues/615

tests/tools/test_pick_aoi.py
yellowcap: Good catch. That is what you get for letting agents write your tests. 😊

PR #612, opened 2026-04-07
Allow triggering tools tests with label

yellowcap:
Allow triggering tools unit tests in CI through adding the `run-tools-tests` on PRs.


jterry64 (approved):
Awesome, could you also add some documentation in the repo on when which different tests run in our CI workflow?

PR #613, opened 2026-04-07
Simplify function structure

yellowcap:
A few suggestions on the PR


yellowcap:
Upstream PR https://github.com/wri/project-zeno/pull/608 has been closed but branch is still around. I am moving this into the branch in case its getting picked up later.

PR #614, opened 2026-04-07
Enable primary forest in the tropics on tree cover-related layers

jterry64:
This change adds support the primary forest filter for supported datasets: tree cover, tree cover loss and tree cover gain. The primary forest layer extent is the tropics, so the filter should only be applied for AOIs in the tropics (from -30 to 30 latitude).

# Summary of changes:

- Enable primary forest as contextual_layer on relevant dataset YAMLs
- Fix analytics_handler to correctly pass forest_filter if primary_forest is contextual layer
- Start using AOI info when choosing datasets. Add bounding box to AOIs so LLM can judge extent match.
- Update tests


jterry64:
@yellowcap @gtempus if we want to run the tool tests regularly, it seems like there's a couple flaky or consistently failing ones in pull_data, generate_insights and pick_dataset we should make sure passing on main.

yellowcap (changes requested):
I like the approach but it will enforce that an aoi is picked before, which prevents layers to be added just for visualization. So I would suggest that the additional instructions are only added to the pick dataset llm call if the aoi is available, skip otherwise.

Like this the user can just ask about some datasets and get them on screen without having to commit to a place. cc @01painadam

yellowcap:
> @yellowcap @gtempus if we want to run the tool tests regularly, it seems like there's a couple flaky or consistently failing ones in pull_data, generate_insights and pick_dataset we should make sure passing on main.

Yes the tests are indeed flaky, and they also call a lot of external services, and rely on llms. In other projects, we sometimes have to re-run tests for them to pass. Though in this test suite there are a few that are expected to fail at the moment (like some of the pull data tests).

jterry64:
@yellowcap oof forgot they could be called in any order. I can make it optional, but this is still a little tricky, because the user could choose an AOI where the layer is invalid. Need to think about it how to deal with that. 

Ideally the LLM could figure this out automatically and inform the user, but that'd require another pick_dataset call... maybe for now the API can just throw a good error message about extent, and that can get propagated to the user to re-select the dataset?

jterry64:
@yellowcap ok made it optional. I was trying to test prompts with dataset but no AOI, but it does seem like currently it's pretty insistent on getting a location if you ask for a dataset (even in prod). So we may need to tweak the system prompt in a future PR if we want it to do that.

jterry64:
These evals now pass consistently here against my local: https://docs.google.com/spreadsheets/d/1_G1aq2fSCPqhT6w55_Od6VU7sov76t1lHQTBeZZxbdM/edit?gid=563440160#gid=563440160

query | expected_dataset | expected_dataset_id | expected_start_date | expected_end_date | expected_context_layer | expected_clarification | expected_answer
-- | -- | -- | -- | -- | -- | -- | --
Primary forest only: show tree cover gain in Brazil since 2000. | tree cover gain | 5 | 2000 | 2000, 2020 | primary_forest |   | The answer should be clear that it's showing the tree cover gain in primary forest.
Primary forest only: show tree cover loss in Canada since 2000. | tree cover loss | 4 | 2000 | 2000, 2024 | none | FALSE | The answer should explain that primary forest data is unavailable for Canada because the layer is pan-tropical only, and should provide total annual tree cover loss for Canada instead.
Deforestation in natural forests in Brazil since 2000. | tree cover loss | 4 | 2000 | 2000, 2024 | primary_forest |   |  
Primary forest loss in Canada, Brazil, and Mexico since 2000. | tree cover loss | 4 | 2000 | 2000, 2024 |   | FALSE | The answer should explain that primary forest data is pan-tropical only, and because not all the countries are tropical, it's analyze tree cover loss without primary forest instead.

src/agent/tools/pick_dataset.py
```diff
@@ -202,6 +211,12 @@ async def select_best_dataset(
         DATASET_SELECTION_PROMPT
         | SMALL_MODEL.with_structured_output(DatasetOption)
     )
+
+    if aoi_selection is not None:
```
yellowcap: standard here would be the `if aoi_selection is None` csae first and then `else` instead of `is not None`. Slightly lighter to read.

yellowcap (approved):
Great addition thanks @jterry64 !

src/agent/tools/pick_dataset.py
jterry64: good point, done!

PR #616, opened 2026-04-10
Vector tile endpoints for aoi selection

yellowcap:
Opening only for visibility

This draft PR adds vector tile endpoints and filters for rendering on the frontend after AOI seletction.

For many AOIs our current frontend solution does not work, it downloads geojsons one by one for each aoi selected.

The approach here uses the GFW static tiles api and generates filters that reduce the layer to the selected aois

We should consider adding vector tiles endpoints for custom areas, so that the frontend can always rely on vector tiles and we don't duplicate the layer mechanism in geojson + vectortiles.

### DIFF
--- a/inspect_gfw_pbf_tiles.py
+++ b/inspect_gfw_pbf_tiles.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python3
+"""Download and inspect sample GFW vector PBF tiles.
+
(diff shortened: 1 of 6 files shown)


yellowcap:
This makes sense in the long term, but the global queries were solved differently in #610 . So we can close this for now and re-visit when we want to make aoi rendering more efficient.

PR #617, opened 2026-04-10
Hotfix: logic to switch from change to composition

yellowcap:
The flag for composition vs time was switched. This PR fixes the use of the flag.


batpad (approved):

01painadam (commented):
LGTM - tests pass

PR #618, opened 2026-04-13
PZB-679: Filter for contextual layer extent (primary forest)

jterry64:
While testing in staging, I found the agent was pretty inconsistent at applying the prompted extent rules for primary forest. t seemed to have its own internal knowledge of things like primary forest, ecozones, extents, etc. and would use that info to override our business logic. For example, in Chad, which has some primary forest loss in the bottom half, it would often refuse to use the primary forest layer because it knows overall it doesn't have much humid tropical forest. It also often decided pantropical extent was -23.5 to 23.5 instead of the defined -30 to 30.

Rather than fiddling around with the prompt instructions more, I decided to try a different approach and just only present the agent with options in the given extent. If we always want it to follow spatial/temporal rules, it seemed easier to apply this beforehand rather than make the agent do it nondeterministically. This also ensures it'll be consistent across vendor models. 

I do provide the filtered reasons with a reason why they were filtered, so the LLM can explain to the user why a layer wasn't chosen even if directly asked for.

This also includes a small adjustment to the TCL YAML to make primary forest the default for forest loss or deforestation queries.


jterry64:
Will look into tomorrow why it's failing test_graph here, it passes locally, but is a nondeterministic test

src/agent/tools/pick_dataset.py
```diff
@@ -213,9 +212,17 @@ async def select_best_dataset(
...
+        ]
+        removed_df = None
     else:
-        aois = pd.DataFrame(aoi_selection["aois"]).to_csv(index=False)
+        filtered_layers, removed_layers = get_filtered_contextual_layers(
+            candidate_datasets["context_layers"], aoi_selection
+        )
+
+        candidate_datasets["filtered_context_layers"] = filtered_layers
+        removed_df = removed_layers.to_csv(index=False)
```
yellowcap, line 221: What is the advantage of that? We might not need this, as the model will only see the filtered context layers so it won't choose datasets it does not see at all.

yellowcap (commented):
Looks good to me, except for one comment about if its needed to list excluded datasets.

src/agent/tools/pick_dataset.py
jterry64: @yellowcap it's probably not super important right away, but I was thinking of when users ask fairly directly for datasets ("get my tree cover loss in primary forest in Canada"). This way, the `reason` it generates can explain why a layer wasn't used, even if it normally would be for the query. 

In this case, I noticed the LLM is actually aware from its world knowledge there's a primary forest layer associated with GFW, so it does actually explain it can't use it without this. So glad to remove it for now if you think it'll cause issues.

jterry64: @yellowcap I think we do need it. I have some evals that check whether the answer is clear that it's not using primary forest (despite being requested) due to it being out of extent. These start failing if I don't explicitly list that here, it looks like we're just ignoring that part of the user question.

yellowcap: ok understood and makes sense.

PR #620, opened 2026-04-14
PZB-752: Include tile URL of contextual layers in agent state output

jterry64:
Add a new field `context_layers` to the `dataset` state, containing a ContextLayer object that has additional layer metadata like the tile URL. The frontend should be able to consume this to load context layers on the map along with the main dataset. Created as a new field to not break existing interface, but will delete old `context_layer` property once it's no longer in use.

Example new message from chat API:
```
{
      "type":"tool",
      "name":"pick_dataset",
      "content":"# About the selection\n    Selected dataset name: Tree cover loss\n    Selected context layer: primary_forest\n    Reasoning for selection: This dataset provides annual tree cover loss data and includes a specific context layer for primary forests, which directly addresses the user's query about deforestation in primary forest.\n\n    # Additional dataset information\n\n    ## Description\n\n    Tree Cover Loss (Hansen/UMD/GLAD) maps annual global forest loss from 2001 to 2024 at 30-meter resolution using Landsat satellite imagery. It detects stand-replacement disturbances in vegetation over 5 meters tall, including natural forests and plantations. The dataset supports monitoring annual tree cover loss and deforestation trends, fire impacts, and forestry practices, and is widely used for conservation, land-use planning, and environmental policy analysis. This dataset also includes annual greenhouse gas emissions due to tree cover loss, measure in MgCO2e (tonnes of CO2 equivalent).\n\n\n    ## Function usage notes:\n\n    Identifies areas of gross tree cover loss\n\n\n    ## Usage cautions\n\n    In this data set, \"tree cover\" is defined as all vegetation greater than 5 meters in height, and may take the form of natural forests or plantations across a range of canopy densities. \"Loss\" indicates the removal or mortality of tree cover and can be due to a variety of factors, including mechanical harvesting, fire, disease, or storm damage. As such, \"loss\" does not equate to deforestation.\nDue to variation in research methodology and date of content, tree cover, loss, and gain data sets on GFW cannot be compared accurately against each other. Accordingly, \"net\" loss cannot be calculated by subtracting figures for tree cover gain from tree cover loss, and current (post-2000) tree cover cannot be determined by subtracting figures for annual tree cover loss from year 2000 tree cover.\nThe 2011-2024 data was produced using an updated methodology. Comparisons between the original 2001-2010 data and the 2011-2024 update should be performed with caution.\nIn the original publication, the authors evaluated the overall prevalence of false positives (commission errors) in this data at 13%, and the prevalence of false negatives (omission errors) at 12%, though the accuracy varies by biome and thus may be higher or lower in any particular location. The model often misses disturbances in smallholder landscapes, resulting in lower accuracy of the data in sub-Saharan Africa, where this type of disturbance is more common. Largely because of a delay between an actual disturbance and the event being observed by the satellite imagery, the authors are 75% confident that the loss occurred within the stated year, and 97% confident that it occurred within a year before or after. Users of the data can smooth out such uncertainty by examining the average over multiple years.\n\n\n    ## Content date\n\n    2001-2024 annual\n    ",
      "dataset":{
         "dataset_id":4,
         "context_layer": "primary_forest",
         "context_layers": [{
           "name": "primary_forest",
           "tile_url": "https://tiles.globalforestwatch.org/umd_regional_primary_forest_2001/v201901/uint16/{z}/{x}/{y}.png"
         }],
         "reason":"This dataset provides annual tree cover loss data and includes a specific context layer for primary forests, which directly addresses the user's query about deforestation in primary forest.",
         "language":"en",
         "tile_url":"https://tiles.globalforestwatch.org/umd_tree_cover_loss/latest/dynamic/{z}/{x}/{y}.png?tree_cover_density_threshold=30&render_type=true_color&start_year=2001&end_year=2024",
         "dataset_name":"Tree cover loss",
         "analytics_api_endpoint":"/v0/land_change/tree_cover_loss/analytics",
         "description":"Tree Cover Loss (Hansen/UMD/GLAD) maps annual global forest loss from 2001 to 2024 at 30-meter resolution using Landsat satellite imagery. It detects stand-replacement disturbances in vegetation over 5 meters tall, including natural forests and plantations. The dataset supports monitoring annual tree cover loss and deforestation trends, fire impacts, and forestry practices, and is widely used for conservation, land-use planning, and environmental policy analysis. This dataset also includes annual greenhouse gas emissions due to tree cover loss, measure in MgCO2e (tonnes of CO2 equivalent).\n",
         "prompt_instructions":"Reports gross annual loss of tree cover ≥ 5 m height (2001-2024). Show yearly loss with bar charts or cumulative loss with stacked-area charts; plot area_ha and drop zeros. When reporting tree cover loss, always include the GHG emissions associated with that loss, in MgCO2e (tonnes of CO2 equivalent). If users ask for intra-year or seasonal tree cover loss OR emissions, refuse and remind that the data is annual only. The 2011-2024 data was produced using an updated methodology so comparisons between, and trends across the periods 2001-2010 and 2011-2024 should be performed with caution. DO NOT use the term \"deforestation\", use \"tree cover loss\" instead since \"Tree cover\" includes plantations as well as natural forest meaning loss is not automatically deforestation or carbon loss. The product measures canopy loss only; it doesn't track regrowth or permanence. Use 30% threshold. If a user asks for net gain/loss, refuse and explain that the methodologies for the two datasets differ and thus they cannot be combined directly to calculate net change. DO NOT show emissions and loss in the same chart, use separate charts.\n",
         "methodology":"This data set, a collaboration between the GLAD (Global Land Analysis & Discovery) lab at the University of Maryland (UMD), Google, USGS, and NASA, measures areas of tree cover loss across all global land (except Antarctica and other Arctic islands) at approximately 30 x 30 meter resolution. The data were generated using multispectral satellite imagery from the Landsat 5 thematic mapper (TM), the Landsat 7 thematic mapper plus (ETM+), and the Landsat 8 Operational Land Imager (OLI) sensors. Over 1 million satellite images were processed and analyzed, including over 600,000 Landsat 7 images for the 2000-2012 interval, and more than 400,000 Landsat 5, 7, and 8 images for updates for the 2011-2022 interval, and additional images used for 2023 and 2024. The clear land surface observations in the satellite images were assembled and a supervised learning algorithm was applied to identify per pixel tree cover loss.\nIn this data set, \"tree cover\" is defined as all vegetation greater than 5 meters in height, and may take the form of natural forests or plantations across a range of canopy densities. Tree cover loss is defined as \"stand replacement disturbance\" which is considered to be clearing of at least half of tree cover within a 30-meter pixel. The exact threshold is variable both through space and time, and is biome-dependent. Tree cover loss may be the result of human activities, including forestry practices such as timber harvesting or deforestation (the conversion of natural forest to other land uses), as well as natural causes such as disease or storm damage. Fire is another widespread cause of tree cover loss, and can be either natural or human-induced.\nThis data set has been updated five times since its creation, and now includes loss up to 2024 (Version 1.12). The analysis method has been modified in numerous ways, including new data for the target year, re-processed data for previous years, and improved modelling and calibration. These modifications improve change detection for 2011-2024, including better detection of boreal loss due to fire, smallholder rotation agriculture in tropical forests, selective losing, and short cycle plantations. Since the entire historical timeseries was not reprocessed with the updated methodology, time-series assessments should be performed with caution.\nWhen zoomed out (< zoom level 13), pixels of loss are shaded according to the density of loss at the 30 x 30 meter scale. Pixels with darker shading represent areas with a higher concentration of tree cover loss, whereas pixels with lighter shading indicate a lower concentration of tree cover loss. There is no variation in pixel shading when the data is at full resolution (≥ zoom level 13). The tree cover canopy density of the displayed data varies according to the selection - use the legend on the map to change the minimum tree cover canopy density threshold.\n",
         "cautions":"In this data set, \"tree cover\" is defined as all vegetation greater than 5 meters in height, and may take the form of natural forests or plantations across a range of canopy densities. \"Loss\" indicates the removal or mortality of tree cover and can be due to a variety of factors, including mechanical harvesting, fire, disease, or storm damage. As such, \"loss\" does not equate to deforestation.\nDue to variation in research methodology and date of content, tree cover, loss, and gain data sets on GFW cannot be compared accurately against each other. Accordingly, \"net\" loss cannot be calculated by subtracting figures for tree cover gain from tree cover loss, and current (post-2000) tree cover cannot be determined by subtracting figures for annual tree cover loss from year 2000 tree cover.\nThe 2011-2024 data was produced using an updated methodology. Comparisons between the original 2001-2010 data and the 2011-2024 update should be performed with caution.\nIn the original publication, the authors evaluated the overall prevalence of false positives (commission errors) in this data at 13%, and the prevalence of false negatives (omission errors) at 12%, though the accuracy varies by biome and thus may be higher or lower in any particular location. The model often misses disturbances in smallholder landscapes, resulting in lower accuracy of the data in sub-Saharan Africa, where this type of disturbance is more common. Largely because of a delay between an actual disturbance and the event being observed by the satellite imagery, the authors are 75% confident that the loss occurred within the stated year, and 97% confident that it occurred within a year before or after. Users of the data can smooth out such uncertainty by examining the average over multiple years.\n",
         "function_usage_notes":"Identifies areas of gross tree cover loss\n",
         "citation":"Hansen et al., 2013. \"High-Resolution Global Maps of 21st-Century Forest Cover Change.\" Accessed through Global Forest Watch on [date]. [www.globalforestwatch.org](https://www.globalforestwatch.org)\n",
         "content_date":"2001-2024 annual",
         "selection_hints":"Best dataset for annual tree cover loss questions (2001-2024) at 30m resolution. Covers all vegetation >5m height including plantations — NOT limited to natural forest. Annual resolution only; cannot answer intra-year or seasonal questions. For driver attribution (WHY loss happened), use \"Tree cover loss by dominant driver\" instead. For net change, loss and gain datasets use different methodologies and cannot be combined. Includes GHG emissions (MgCO2e) associated with loss. Methodology changed between 2010 and 2011; cross-period trend comparisons need caution.\n",
         "code_instructions":"CHART TYPES: - Yearly loss → bar chart (x=year, y=area_ha) - Cumulative loss → stacked-area chart - Emissions → MUST be a SEPARATE chart from loss area (NEVER combine on same chart) DATA RULES: - Always include GHG emissions (MgCO2e) alongside loss area, but in a SEPARATE chart - Drop rows where area_ha = 0 - Use 30% canopy density threshold - If primary_forest or intact_forest variable is active, filter accordingly DO/DON'T: - DO NOT use the term \"deforestation\" — always use \"tree cover loss\" - If user asks for intra-year or seasonal breakdown, REFUSE — data is annual only - If user asks for net gain/loss, REFUSE — methodologies differ between loss and gain datasets - Warn about 2001-2010 vs 2011-2024 methodology difference when showing trends across both periods\n",
         "presentation_instructions":"Use \"tree cover loss\" not \"deforestation\" — tree cover includes plantations, not just natural forest. This measures canopy loss only; it doesn't track regrowth or permanence. Always state the canopy density threshold used (30%). If showing emissions, clarify units are MgCO2e (tonnes of CO2 equivalent). If trend spans 2001-2024, note the methodology change after 2010.\n"
      },
      "insights":[
         
      ],
      "charts_data":[
         
      ],
      "codeact_parts":[
         
      ],
      "source_urls":[
         
      ],
      "insight_count":0,
      "timestamp":"2026-04-14T16:25:52.104Z"
   }
   ```


yellowcap (commented):
I quite like this. The only question I have is why `context_layers` is a list if at the moment we can always only pick one.

I discussed with @faustoperez and @kamicut today about context layers having tile urls and the current proposal makes sense to us as a first step. 


For context:

We can later on think about how to structure connections between primary and context layers in the interface. In terms of state management this gets complicated quickly, because we might have multiple primary and contextual layers on the explorer view at the same time. So moving towards something where the user can optionally and pro-actively select context to use in the next prompt might be a viable solution.

jterry64:
@yellowcap it's a list just because we anticipate we may want multiple layers soon (e.g. primary forest and canopy cover, since we use both in analysis) and don't want to break the interface later.

That's a good to know, one thing we were debating was if contextual layers should be under the dataset, or we should just have multiple datasets in the agent state? We went with this approach because it was simpler (we didn't have all those other attributes handy for primary forest), but we could go the other way if that'll make state management easier long term.

willian-viana (approved):
The implementation is very straightforward and doesn't seem to implement anything that would break the client we have now. 

A further discussion on how we want to display this type of intersection would be nice to have.

PR #621, opened 2026-04-15
Fix/reasoning language pollution

01painadam:
## Problem                                                       
                                                                                                       
The agent was responding in the local language of the selected AOI rather than the language of the user's query. Querying data for Spain produced Spanish responses; Brazil produced Portuguese etc etc
                                                                                                       
The cause was in pick_dataset: the LLM selection prompt asked for a free-text reason explaining the dataset choice, and instructed it to "use the language of the user query". In practice the model used the AOI's local language instead — inferred from the place names and bounding box passed as context.

This reason string was embedded directly in the ToolMessage, which LangGraph appends to the shared
message history. Every subsequent LLM call in the pipeline inherited the foreign-language context.

## Changes

`src/agent/tools/pick_dataset.py`
  - Removed reason and language from DatasetOption / DatasetSelectionResult
  - Added date_range_warning: Optional[str] a structured, English-only field populated only when the 
  dataset doesn't cover the requested date range (previously this warning lived inside the free-text  
  reason, so it was being silently lost)                                                               
  - Updated the selection prompt to populate date_range_warning in English when applicable, and leave
  it null otherwise                                                                                    
  - ToolMessage now includes the warning section only when set; no free-text from the LLM selection    
  step leaks into message history                                                                  
                                                                                                       
`tests/tools/test_pick_dataset.py`
  - Added failing tests (committed before the fix) covering the language pollution scenario
  - Updated _make_fake_selection helper to use the new date_range_warning field                        
  - Added tests asserting the warning appears in the tool message when set and is absent when null
                                                                                                       
## TODO
Full-pipeline eval that detects the language of the final agent response and asserts it matches the query language in two cases e.g.
- Non-english query + Non-english speaking AOI
- English query + Spain AOI (regression)

### DIFF
--- a/src/agent/tools/pick_dataset.py
+++ b/src/agent/tools/pick_dataset.py
@@ -76,8 +76,13 @@ class DatasetOption(BaseModel):
     reason: str = Field(
         description="Short reason why the dataset is the best match."
     )
-    language: str = Field(
-        description="Language of the user query.",
+    date_range_warning: Optional[str] = Field(
+        None,
+        description=(
+            "If the dataset does not cover the requested date range exactly, "
+            "provide a brief warning in English explaining the mismatch and what "
+            "date range will be used instead. Null if the date range is covered."
+        ),
     )
 
     @field_validator("dataset_id")
@@ -173,9 +178,10 @@ async def select_best_dataset(
     allow difrenciating between different types of data within the same dataset. So if a user asks
     to show something like "show me tree cover loss by driver", you should select a context layer.
 
-    Evaluate if the best dataset is available for the date range requested by the user,
-    if not, pick the closest date range but warn the user that there
-    is not an exact match with the query requested by the user in the reason field.
+    Evaluate if the best dataset is available for the date range requested by the user.
+    If not, pick the closest available date range and populate date_range_warning in English
+    with a brief explanation of the mismatch and what date range will be used instead.
+    Leave date_range_warning null if the requested date range is fully covered.
 
     Context-layer extent is a hard constraint, if provided, not a warning. If the AOI bbox does not intersect the
     context-layer bbox, you MUST return context_layer = null. Do not select the context layer and explain the limitation.
@@ -187,8 +193,6 @@ async def select_best_dataset(
 
     Keep explanations concise. Do not use datset IDs to describe the dataset.
     For instance, instead of saying "Dataset ID: 123", say "Dataset: Tree Cover Loss".
-
-    Use the language of the user query to generate the reason.
 
     AOI bounding box:
 
@@ -236,7 +240,8 @@ async def select_best_dataset(
     logger.debug(
         f"Selected dataset ID: {selection_result.dataset_id}. "
         f"context_layer={selection_result.context_layer!r} (type={type(selection_result.context_layer).__name__}). "
-        f"Reason: {selection_result.reason}"
+        f"Reason: {selection_result.reason}. "
+        f"Date range warning: {selection_result.date_range_warning!r}"
     )
 
     selected_row = candidate_datasets[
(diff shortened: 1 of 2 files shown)


src/agent/tools/pick_dataset.py
```diff
@@ -173,9 +178,10 @@ async def select_best_dataset(
     allow difrenciating between different types of data within the same dataset. So if a user asks
     to show something like "show me tree cover loss by driver", you should select a context layer.
 
-    Evaluate if the best dataset is available for the date range requested by the user,
-    if not, pick the closest date range but warn the user that there
-    is not an exact match with the query requested by the user in the reason field.
+    Evaluate if the best dataset is available for the date range requested by the user.
+    If not, pick the closest available date range and populate date_range_warning in English
```
yellowcap: Specifying something to be done in English here is not the right approach. Change this to instruct the llm to add the date range warning into the pick reason.

```suggestion
    If not, pick the closest available date range and include a warning in the dataset pick reason
```

src/agent/tools/pick_dataset.py
```diff
@@ -76,8 +76,13 @@ class DatasetOption(BaseModel):
     reason: str = Field(
         description="Short reason why the dataset is the best match."
     )
-    language: str = Field(
-        description="Language of the user query.",
+    date_range_warning: Optional[str] = Field(
```
yellowcap: We have a pick reason, this message should be part of the pick reason not an additional output field.

src/agent/tools/pick_dataset.py
```diff
@@ -188,8 +194,6 @@ async def select_best_dataset(
     Keep explanations concise. Do not use datset IDs to describe the dataset.
     For instance, instead of saying "Dataset ID: 123", say "Dataset: Tree Cover Loss".
 
-    Use the language of the user query to generate the reason.
```
yellowcap, line 191: We can remove the output field, but this should maybe still stay? Is the model bad at picking up the user query language?

tests/tools/test_pick_dataset.py
```diff
@@ -657,6 +659,153 @@ async def test_tcl_by_driver_always_gets_driver_context_layer(state):
...
+        ),
+        # Reason in English — should also not appear in tool message
+        (
+            "This dataset provides annual tree cover loss data at 30m resolution",
+            "Indonesia",
+        ),
+    ],
+)
+async def test_tool_message_does_not_contain_reason(reason, aoi_name):
+    """The reason field must not appear in the ToolMessage content.
```
yellowcap: I don't agree, this is important context for the agent.

tests/tools/test_pick_dataset.py
```diff
@@ -657,6 +659,153 @@ async def test_tcl_by_driver_always_gets_driver_context_layer(state):
...
+
+    Reason is generated in the AOI's local language (e.g. Portuguese for Brazil,
```
yellowcap: That is an interesting observation. If that is the reason for the language change, then lets add a line to answer in the query language not in the language of the place mentioned in the query.

yellowcap (changes requested):
I think you might have identified the reason for the language switching problem (the mentioning of a place), but the solution is not the right one in my opinion.

01painadam:
Thanks for the review @yellowcap - made the recommended changes. Tests pass and the results look good in local:

- Que tamaño de area fue desforestada en los Estados Unidos entre 2015 y 2020? --> response in spanish
- Qual a extensão de terras cultiváveis ​​na República da Irlanda? --> portuguese
- What are the trends in grassland area in Spain --> English

yellowcap (approved):
Really nice catch @01painadam ! Kind of crazy that the llm would change language based on places named. But also makes sense.

PR #623, opened 2026-04-15
PZB-719: Simplified TCD threshold for analysis

jterry64:
Add concept of "parameters" to datasets for intersections with specific values. This simplifies a few other PRs and specifically targets canopy cover, which is one of the few adjustable parameters we have available right now. 

Summary of changes: 
- Parameters are added as an optional field on YAMLs
- During pick dataset, agent is prompted to select parameters if relevant to query
- For canopy cover, analytics handler will use selected parameter or 30 by default. In the future, we may choose a default based on country the AOI is in.
- Added some unit tests for analytics handler, because it was difficult to validate the payload with just the tool tests
- Made an adjustment to pull data to merge the YAML dataset with the agent state properly. I ran into a bug where parameters was getting overwritten, this seems like it'll prevent those kinds of issues in the future.


src/agent/tools/datasets/tree_cover_gain.yml
```diff
@@ -1,9 +1,6 @@
 dataset_id: 5
 dataset_name: Tree cover gain
-context_layers:
```
01painadam, line 3: @jterry64 Is this removed because its not available in the analytics API? Do we need the same for Tree Cover dataset too?

src/agent/tools/pick_dataset.py
```diff
@@ -175,11 +184,16 @@ async def select_best_dataset(
     to show something like "show me tree cover loss by driver", you should select a context layer. These are pre-filtered
     to match the spatiotemporal query constraints.
 
+    Select parameters and values if they are relevant or specified in the user query. Parameters allow further filtering
+    the analysis to better answer the query. Select only values listed in the value field for a parameter. For example,
+    if a user says "show me tree cover loss in forests where canopy cover is greater than 50%", you may select the parameter canopy cover
+    and value 50.
+
```
01painadam, line 200: I like this - leaves it open to be extensible to other param types

tests/tools/test_pick_dataset.py
```diff
@@ -422,7 +422,7 @@ async def test_queries_return_expected_dataset(
         ("Vegetation disturbances over grasslands", 0, "grasslands"),
         ("Tree cover loss by driver", 8, "driver"),
         ("Tree cover loss in primary forest", 4, "primary_forest"),
-        ("Tree  cover loss in the past decade", 4, None),
+        ("Tree  cover loss in the past decade in sparse forests", 4, None),
```
01painadam, line 428: `("Tree cover loss in the past decade in sparse forests", 4, None),`

Double space can be removed

jterry64:
No comments so I'm assuming this is ok. Converted to full PR and added tests.

01painadam:
Screenshot 2026-04-17 at 09 39 51

Nice works as expected! This makes me realise we should probably add the params used to the Insight more clearly cc: @yellowcap

01painadam (approved):
Looks good! Tested locally within stub front-end and I'm getting expected results

PR #628, opened 2026-04-17
PZB-776 - Standalone insights: move payloads to DB and store `insight_id` in state

yellowcap:
### Summary
This PR moves `generate_insights` output from LangGraph checkpoint state into a dedicated PostgreSQL `insights` table.  
The tool now writes the full payload to DB and returns only `insight_id` in state; clients fetch full insight data via `GET /api/insights/{insight_id}`.

### Changes
- Added `InsightOrm` + Alembic migration for `insights`
- Refactored `generate_insights` to persist insight rows and update state with `insight_id`
- Updated `AgentState` to replace inline insight fields with `insight_id`
- Added insights API endpoint + response schema + access control
- Updated tests for the new `insight_id` flow (`generate_insights`, graph, eval/judge)

### Backward Compatibility
Old checkpoints with inline `charts_data`/`insight` continue to replay; new checkpoints use `insight_id`.  
Clients should handle both shapes (prefer `insight_id`, fallback to inline fields).

### ADR Notes
- **Hybrid schema decision:** typed columns for stable metadata, JSONB only for dynamic `chart_data`
- **Stream/replay decision:** no server-side hydration; streaming stays key-passthrough and clients fetch insight payloads separately
- **Migration decision:** no backfill of old checkpoints due to high risk/low value; legacy threads remain supported as-is

### DIFF
--- a/tests/evals/judge.py
+++ b/tests/evals/judge.py
@@ -177,6 +177,7 @@ async def run_generate_insights(query: str, state: dict) -> dict:
     Returns dict with: chart_type, chart_data, insight, code, refused
     """
     from sqlalchemy import select
+    from sqlalchemy.orm import selectinload
 
     from src.api.data_models import InsightOrm
     from src.shared.database import get_session_from_pool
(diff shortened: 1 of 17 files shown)


yellowcap:
I was able to render multiple insights from different convos on a quick debug page that just renders all insights from a user in the DB. So this is working nicely!

image

01painadam:
Great work on this! 

Excited that this can also enable exposing the params selected in the analysis to users for transparency (essentially enabling [this](https://www.figma.com/design/fUyMHyhGck5FmGEg0HlNHa/-Global-Nature-Watch--Playground?node-id=642-723&t=Inrd8bU1VT86P5kW-4))

tests/evals/judge.py
```diff
@@ -198,34 +203,44 @@ async def run_generate_insights(query: str, state: dict) -> dict:
...
+                    await session.execute(
+                        select(InsightOrm).where(InsightOrm.id == insight_id)
+                    )
+                )
+                .scalars()
+                .first()
+            )
+
+        if row:
+            result["chart_type"] = row.chart_type
```
solomon-negusse: This should be fetched from `row.charts`, right?

solomon-negusse (approved):
Looks good @yellowcap, just have one comment to check on.

tests/evals/judge.py
yellowcap: updated judge stuff, including added a way to execute the judge flow through pytest. Worked fine locally, but we are not running this in CI at the moment.

PR #629, opened 2026-04-17
set x-environment header by environment

solomon-negusse:
Pass environment flag to analytics api. This is needed for tcl release to test it in analytics staging while it's in embargo.

### DIFF
--- a/src/agent/tools/data_handlers/analytics_handler.py
+++ b/src/agent/tools/data_handlers/analytics_handler.py
@@ -145,7 +145,7 @@ class AnalyticsHandler(DataSourceHandler):
         "Content-Type": "application/json",
         "X-environment": (
             "production"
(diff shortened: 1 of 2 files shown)


src/agent/tools/data_handlers/analytics_handler.py
```diff
@@ -142,6 +143,11 @@ class AnalyticsHandler(DataSourceHandler):
     HEADERS = {
         "Accept": "application/json",
         "Content-Type": "application/json",
+        "X-environment": (
+            "production"
+            if os.getenv("STAGE", "production").strip().lower() == "production"
```
yellowcap:
```suggestion
            if os.getenv("GNW_STAGE", "production").strip().lower() == "production"
```

yellowcap (approved):
Looks good to me, I'd be _slightly_ more verbose on the env variable name, but otherwise this can go in.

PR #630, opened 2026-04-20
Fix/antimeridian bbox

01painadam:
### Summary
pick_aoi returned world-spanning bboxes for countries that straddle the antimeridian (Fiji, Russia, USA...)

For example, Fiji returned [-179.99, -21, 180.00, -12], a bbox that spans the entire globe, because `ST_XMin` / `ST_XMax` on a multipolygon that crosses 180degrees produces extreme values.

The frontend uses this bbox to fly the map to the selected AOI, so Fiji would zoom out to the whole world instead of the Fiji islands.

**WARNING** this PR is blocked by a complimentary FE fix in `project-zeno-next` [fix/antimeridian-bbox](https://github.com/wri/project-zeno-next/pull/466), as this introduces a breaking change

| Before | After |
| --- | --- |
| Screenshot 2026-04-20 at 11 25 02 | Screenshot 2026-04-20 at 11 23 48 |

### Fix
Replaced the naive json_build_array(ST_XMin, ST_YMin, ST_XMax, ST_YMax) with _antimeridian_bbox_sql(), a PostGIS CASE expression that:

1. Detects crossing geometries via ST_XMax - ST_XMin > 180
2. Classifies each sub-polygon by its centroid longitude (centroid > 0 = eastern hemisphere, centroid < 0 = western)
3. Computes a crossing bbox where west > east e.g. Fiji: [177, -21, -179, -12]
4. Falls back to the original naive bbox when no crossing is detected or either hemisphere subquery returns NULL

The centroid approach is necessary because sub-polygons clipped at 180 degrees have coords that don't distinguish hemisphere by longitude range alone.

Applied to both BBOX_SQL (standard geometry tables) and CUSTOM_BBOX_SQL (custom area geometries).

### Tests

9 new tests in tests/unit/agent/tools/test_an timeridian_bbox.py:
- 5 unit tests — SQL string structure (no DB required)
- 4 PostGIS integration tests using synthetic WKT geometries: 
- Fiji-like, Russia-like crossings west > east
- Contiguous USA and France: non-crossing

### DIFF
--- a/src/agent/tools/pick_aoi/tool.py
+++ b/src/agent/tools/pick_aoi/tool.py
@@ -350,30 +350,19 @@ async def query_subregion_database(
         else ""
     )
 
-    # For GADM-to-GADM lookups use hierarchical ID prefix matching instead of
-    # ST_CoveredBy — avoids an expensive spatial join against the full country
-    # geometry (e.g. USA with Alaska/Hawaii, Russia with Chukotka).
-    if table_name == GADM_TABLE and source == "gadm":
-        sql_query = f"""
-        SELECT t.name, t.subtype, t.{src_id_field}, '{subregion_source}' as source, t.{src_id_field} as src_id, {BBOX_SQL}
-        FROM {table_name} AS t
-        WHERE t.subtype = :subtype
-        AND t.gadm_id LIKE :gadm_prefix
-        {gadm_filter}
-        """
-    else:
-        sql_query = f"""
-        WITH aoi AS (
-            SELECT geometry AS geom
-            FROM {source_table}
-            WHERE "{id_column}" = :src_id
-            LIMIT 1
-        )
-        SELECT t.name, t.subtype, t.{src_id_field}, '{subregion_source}' as source, t.{src_id_field} as src_id, {BBOX_SQL}
-        FROM {table_name} AS t, aoi
-        WHERE t.subtype = :subtype
-        AND ST_CoveredBy(t.geometry, aoi.geom)
-        """
+    sql_query = f"""
+    WITH aoi AS (
+        SELECT geometry AS geom
+        FROM {source_table}
+        WHERE "{id_column}" = :src_id
+        LIMIT 1
+    )
+    SELECT t.name, t.subtype, t.{src_id_field}, '{subregion_source}' as source, t.{src_id_field} as src_id, {BBOX_SQL}
+    FROM {table_name} AS t, aoi
+    WHERE t.subtype = :subtype
+    AND ST_CoveredBy(t.geometry, aoi.geom)
+    {gadm_filter}
+    """
     logger.debug(f"Executing subregion query: {sql_query}")
 
     async with get_connection_from_pool() as conn:
--- a/src/agent/tools/pick_aoi/tool.py
+++ b/src/agent/tools/pick_aoi/tool.py
@@ -370,15 +370,16 @@ async def query_subregion_database(
         def _read(sync_conn):
             processed_src_id = src_id
             if source == "kba":
+                # for these sources IDs stored as numeric values
                 try:
                     processed_src_id = int(processed_src_id)
                 except ValueError:
                     pass
-            params: dict = {"src_id": processed_src_id, "subtype": subtype}
-            if table_name == GADM_TABLE and source == "gadm":
-                gid0 = str(processed_src_id).split(".")[0]
-                params["gadm_prefix"] = f"{gid0}.%"
-            return pd.read_sql(text(sql_query), sync_conn, params=params)
+            return pd.read_sql(
+                text(sql_query),
+                sync_conn,
+                params={"src_id": processed_src_id, "subtype": subtype},
+            )
 
         results = await conn.run_sync(_read)
 
--- a/src/agent/tools/pick_aoi/tool.py
+++ b/src/agent/tools/pick_aoi/tool.py
@@ -344,11 +344,19 @@ async def query_subregion_database(
         f"Querying subregion: {subregion_name} in table: {table_name} for source: {source}, src_id: {src_id}"
     )
 
-    gadm_filter = (
-        f"\n    AND t.gadm_id ~ '{GADM_STANDARD_ID_RE}'"
-        if table_name == GADM_TABLE
-        else ""
-    )
+    if table_name == GADM_TABLE:
+        if source == "gadm":
+            if "." in src_id:
+                subregion_filter = ".".join(src_id.split(".")[:-1])
+            else:
+                subregion_filter = src_id
+            gadm_filter = f" AND t.gadm_id LIKE '{subregion_filter}.%'"
+        else:
+            gadm_filter = f" AND t.gadm_id ~ '{GADM_STANDARD_ID_RE}'"
+        spatial_filter = ""
+    else:
+        gadm_filter = ""
+        spatial_filter = " AND ST_Intersects(t.geometry, aoi.geom) AND NOT ST_Touches(t.geometry, aoi.geom)"
 
     sql_query = f"""
     WITH aoi AS (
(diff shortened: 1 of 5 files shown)


tests/unit/agent/tools/test_antimeridian_bbox.py
```diff
@@ -0,0 +1,131 @@
+"""Unit + integration tests for _antimeridian_bbox_sql.
```
jterry64: I don't think we want all these tests actually hitting the DB. I would just confirm the SQL query works manually and then make tests asserting the SQL is well formed without any dependencies.  Then it will be a unit test (no dependencies).

I would also avoid making a test file just for every bug - FastAPI standard is test file path should match path of file it's testing, so it's to match them up. So should be like tests/unit/agent/tools/pick_aoi/test_tool.py.

src/agent/tools/pick_aoi/tool.py
```diff
@@ -39,18 +39,50 @@
...
+            ST_XMin({geom_expr}),
+            ST_YMin({geom_expr}),
+            ST_XMax({geom_expr}),
+            ST_YMax({geom_expr})
+        )
+    END
+    """
+
+
+BBOX_SQL = f"({_antimeridian_bbox_sql('geometry')}) AS bbox"
```
jterry64, line 79: I just added the bbox here very recently, are you sure this is the source of the error? I don't think the frontend would've been using it at all

src/agent/tools/pick_aoi/tool.py
```diff
@@ -39,18 +39,50 @@
...
+def _antimeridian_bbox_sql(geom_expr: str) -> str:
+    """
+    Returns [west, south, east, north] JSON array.
+    For antimeridian-crossing geometries, produces a crossing bbox where west > east
+    (e.g. Fiji: [177, -21, -179, -12]) using centroid-based sub-polygon classification.
+    Falls back to naive ST_XMin/ST_XMax when no antimeridian crossing is detected
+    or when the crossing subqueries return NULL (edge case).
+    """
+    return f"""
+    CASE
```
jterry64, line 53: I think you can simplify this using something like `SELECT ST_Envelope(ST_ShiftLongitude(geom))` or the `geography` type that'll automatically deal with splitting around the meridian

jterry64 (changes requested):
I just added the bbox here very recently, are you sure this is the source of the error? I don't think the frontend would've been using it

src/agent/tools/pick_aoi/tool.py
01painadam: Yeah, good question! The bbox values I was seeing for Fiji/USA/Russia had a bad smell and the issue here was the ST_XMin/ST_XMax for a geometry straddling the antimeridian returns a  ~360 deg  bbox.

You're right that the FE _wasn't_ consuming , and in the interest of others being able to build on top if this API i though it best to fix it here so that it could act as the source of truth.

(Note that I couldn't get a nice FE fix alone)

src/agent/tools/pick_aoi/tool.py
01painadam: Ah, nice!

01painadam: Actually this change broke it :( 

ST_ShiftLongitude breaks for geoms with width >180 degs that _don't_ cross the antimeridian (the IS NOT NULL checks are the safety net that prevents false positives)

01painadam:
Needed a bit of help to resolve this one so woth your eyes on this again @jterry64 

- Replaced the antimeridian bbox SQL with ST_ClipByBox2D (more performant than ST_Dump)
- moved tests to unit-only with a DB-free conftest.
- Replaced ST_CoveredBy with gadm_id LIKE 'RUS.%' prefix matching for GADM subregion queries to improve latency on lookups for big queries e.g. Russia/USA states.

src/agent/tools/pick_aoi/tool.py
```diff
@@ -39,18 +39,50 @@
...
+            SELECT COALESCE(
+                CASE
+                    WHEN west IS NOT NULL AND east IS NOT NULL
+                    THEN json_build_array(west, ST_YMin({geom_expr}), east, ST_YMax({geom_expr}))
+                END,
+                json_build_array(ST_XMin({geom_expr}), ST_YMin({geom_expr}), ST_XMax({geom_expr}), ST_YMax({geom_expr}))
+            )
+            FROM (
+                SELECT
+                    ST_XMin(ST_Envelope(ST_ClipByBox2D({geom_expr}, {east_half}))) AS west,
```
jterry64, line 65: The ST_ClipByBox2D still clips the original geometry, no? Would it work if you reversed this to clip the bbox, like `ST_ClipByBox2D(ST_Envelope({geom_expr}), {east_half}))`?

tests/unit/agent/tools/pick_aoi/test_tool.py
```diff
@@ -0,0 +1,35 @@
+from src.agent.tools.pick_aoi.tool import _antimeridian_bbox_sql
+
+
+def test_sql_contains_crossing_condition():
+    sql = _antimeridian_bbox_sql("geometry")
+    assert "ST_XMax(geometry) - ST_XMin(geometry) > 180" in sql
+
+
+def test_sql_uses_clip_not_dump():
```
jterry64: I don't think you need to specifically test this since it was just an internal change

tests/unit/agent/tools/pick_aoi/test_tool.py
```diff
@@ -0,0 +1,35 @@
+from src.agent.tools.pick_aoi.tool import _antimeridian_bbox_sql
+
+
+def test_sql_contains_crossing_condition():
```
jterry64, line 4: Thank you for removing the dependency! I think this is fine for now, in the future we'd probably want to be testing `query_aoi_database` and see that this string is showing up appropriately based on mocked geometries/params, but we'd need to also remove the dependencies from `query_aoi_database` so this will hold us over

jterry64 (changes requested):

src/agent/tools/pick_aoi/tool.py
01painadam: Should work, yes - I'll test

01painadam: Screenshot 2026-04-21 at 09 51 57

Looks like theyre non-commutative - reversing changes the output

tests/unit/agent/tools/pick_aoi/test_tool.py
01painadam: Removed

jterry64 (approved):

src/agent/tools/pick_aoi/tool.py
jterry64: Oh ok, well if this was working fast, I'd double check though in staging this isn't going to cause perf issues

src/agent/tools/pick_aoi/tool.py
```diff
@@ -317,36 +349,47 @@ async def query_subregion_database(
...
-        LIMIT 1
-    )
-    SELECT t.name, t.subtype, t.{src_id_field}, '{subregion_source}' as source, t.{src_id_field} as src_id, {BBOX_SQL}
-    FROM {table_name} AS t, aoi
-    WHERE t.subtype = :subtype
-    AND ST_CoveredBy(t.geometry, aoi.geom)
-    {gadm_filter}
-    """
+
+    # For GADM-to-GADM lookups use hierarchical ID prefix matching instead of
```
yellowcap: That's mixing concerns. And PostGIS is very good at doing intersecitons, these are indexed too. We should not do this change.

yellowcap (commented):
Looks good to me except, for the additional tweak that was done for finding subareas.

yellowcap:
I'd like to travel over the antimeridian one day!

PR #631, opened 2026-04-20
Add TCD threshold to map

jterry64:
Right now, TCD threshold is always set to 30 for layers with a parameter option, and isn't included as contextual layer. This PR:

- Add TCD as contextual layer for datasets where it's used as a filter
- Sets the appropriate TCD parameter in the tile URL for all appropriate datasets/contextual layers
- Added TCD parameter to tree cover, TCL by driver and carbon flux, since these all should be filtered by TCD to match TCL from the user
- Pull outs all dataset-specific logic for building tile services into one function to make future refactoring easier. @gtempus our discussion made me think at least putting all this logic in one separate place is a good first step before a bigger refactor, lmk if you agree.

### DIFF
--- a/src/agent/tools/data_handlers/analytics_handler.py
+++ b/src/agent/tools/data_handlers/analytics_handler.py
@@ -280,15 +280,9 @@ class AnalyticsHandler(DataSourceHandler):
             TREE_COVER_LOSS_BY_DRIVER_ID,
         ]:
             forest_filter = None
-            canopy_cover = 30
 
             if dataset.get("context_layer") == "primary_forest":
                 forest_filter = "primary_forest"
-
-            if dataset.get("parameters") is not None:
-                for param in dataset.get("parameters"):
-                    if param["name"] == "canopy_cover":
-                        canopy_cover = max(param["values"])
 
             intersections = []
             if dataset.get("dataset_id") == TREE_COVER_LOSS_BY_DRIVER_ID:
@@ -298,7 +292,6 @@ class AnalyticsHandler(DataSourceHandler):
                 **base_payload,
                 "start_year": start_date[:4],
                 "end_year": end_date[:4],
-                "canopy_cover": canopy_cover,
                 "forest_filter": forest_filter,
                 "intersections": intersections,
             }
@@ -322,7 +315,6 @@ class AnalyticsHandler(DataSourceHandler):
         elif dataset.get("dataset_id") == FOREST_CARBON_FLUX_ID:
             payload = {
                 **base_payload,
-                "canopy_cover": 30,
             }
         elif dataset.get("dataset_id") == TREE_COVER_ID:
             forest_filter = None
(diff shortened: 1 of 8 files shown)


gtempus:
Totally agree, @jterry64. There is so much imperative code in this very large file that breaking out declarative functions will help immensely with future understanding. It's a good and humble first step! 😅

gtempus:
Also, @jterry64, we need to disable the Tools Tests that are consistently failing. 
If the build is failing, we don't merge.

src/agent/tools/data_handlers/analytics_handler.py
```diff
@@ -347,6 +338,20 @@ async def _build_payload(
...
+            TREE_COVER_LOSS_BY_DRIVER_ID,
+            FOREST_CARBON_FLUX_ID,
+        ]:
+            canopy_cover = 30
+            if dataset.get("parameters") is not None:
+                for param in dataset.get("parameters"):
+                    if param["name"] == "canopy_cover":
+                        canopy_cover = max(param["values"])
+
+            payload["canopy_cover"] = canopy_cover
```
yellowcap, line 353: shall we add a validation here? To make sure the selected param is a valid one from the DATASETS yaml? This will break if the llm hallucinates a number.

yellowcap (commented):
Looks good to me, but have not tested locally yet. One comment about the value picking, adding some validation might be important.

src/agent/tools/data_handlers/analytics_handler.py
jterry64: Oh didn't think about that, maybe I can add a general validator for params that it's in the values?

jterry64:
@yellowcap added the same for params what you were doing for context layer, just discarding any hallucinations. Although I found in practice it's pretty hard to make it hallucinate the params without forcing it. 

I was able to do local testing and evals and it seems to be working pretty well now across all these datasets. These are the evals I added, which are all passing, so I think this is good to merge (I'll have a gnw-evals PR up soon for a column I added to check params):

test_id | status | query | expected_dataset | expected_dataset_id | expected_start_date | expected_end_date | expected_context_layer | expected_dataset_parameters | expected_clarification | expected_answer
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
3-557 |   | Forests in Brazil in 2000 where canopy cover is 50% | Tree cover | 7 | 2000 | 2000 |   | [{"name":"canopy_cover","values":[50]}] |   |  
3-558 |   | Forests in Brazil in 2000 where canopy cover is 90% | Tree cover | 7 | 2000 | 2000 |   | [{"name":"canopy_cover","values":[75]}] |   | Clarifies 90% canopy cover is not available, so chooses closest one
3-559 |   | Carbon net flux in Brazil where canopy cover is 50% since 2000 | Forest greenhouse gas net flux | 6 | 2000 | 2024 |   | [{"name":"canopy_cover","values":[50]}] |   |  
3-560 |   | Drivers of deforestation in Brazil in dense forests since 2000 | Tree Cover Loss by Dominant Driver | 8 | 2000 | 2024 | driver | [{"name":"canopy_cover","values":[75]}] |   | Clarifies that interprets dense forest as 75% canopy cover

yellowcap:
Amazing thanks @jterry64 !

PR #632, opened 2026-04-22
Type/schema hardening for agent state, pick_aoi cleanup, and CI test reruns

yellowcap:
### What changed
- Strengthened agent state typing in `src/agent/state.py`:
  - Added `StatisticsParameter` typed structure (`name`, `values`)
  - Changed `Statistics.parameters` from `list[dict] | None` to `list[StatisticsParameter] | None`
  - Replaced `insights: list` with explicit fields:
    - `insight: str`
    - `follow_up_suggestions: list[str]`
  - Replaced `codeact_parts: list[CodeActPart]` with serializable `EncodedCodeActPart` typed dict
  - Removed deprecated top-level state fields:
    - `aoi`
    - `subtype`

- Updated `generate_insights` state access in `src/agent/tools/generate_insights.py`:
  - Switched dataset guideline/caution lookup to use the local `dataset` object consistently (instead of repeated `state.get("dataset")` lookups), making access safer and clearer.

- Removed deprecated `pick_aoi` state writes:
  - `src/agent/tools/pick_aoi/tool.py`
  - `src/agent/tools/pick_aoi/global_queries.py`
  - No longer sets deprecated `update["aoi"]` / `update["subtype"]`; relies on `aoi_selection`.

- Expanded AOI fixtures for tests:
  - Added Russia entries to `tests/fixtures/aoi_pick_aoi_v1.json` for country lookup and subregion query coverage.

- Improved CI resilience for flaky model-dependent tests:
  - Added `pytest-rerunfailures==15.0` to dev deps in `pyproject.toml`
  - Updated `uv.lock`
  - Updated tools workflow command to:
    - `uv run pytest tests/tools tests/agent -v --reruns 2 --reruns-delay 5`

### Why
- Tightening state/schema types reduces ambiguity and catches integration mistakes earlier.
- Removing deprecated `aoi`/`subtype` eliminates duplicate state representations and prevents drift from `aoi_selection`.
- Russia fixture expansion supports newer AOI test scenarios.
- CI reruns mitigate transient Gemini/API failures while still failing truly unstable tests (3 consecutive failures required).

### Impact
- **Typing/contracts:** clearer, stricter state definitions for insights/statistics/codeact payloads.
- **Runtime behavior:** no intended functional change to AOI selection flow beyond removing deprecated fields.
- **Tests/CI:** broader AOI fixture coverage and lower flaky-failure rate in tools/agent workflow.
- **Dependencies:** one new dev dependency (`pytest-rerunfailures`).

### DIFF
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -75,4 +75,4 @@ jobs:
         run: |
           mkdir -p data
           uv run python src/ingest/embed_datasets.py
-          uv run pytest tests/tools tests/agent -v
+          uv run pytest tests/tools tests/agent -v --reruns 2 --reruns-delay 5
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,7 @@ dev = [
     "geopandas==1.0.1",
     "lancedb==0.24.1",
     "pytest-asyncio==1.1.0",
+    "pytest-rerunfailures==15.0",
     "locust==2.32.5",
 ]
 frontend = [
--- a/src/agent/state.py
+++ b/src/agent/state.py
@@ -5,8 +5,6 @@ from langchain_core.messages import BaseMessage
 from langgraph.graph import add_messages
 from typing_extensions import TypedDict
 
-from src.agent.tools.code_executors.base import CodeActPart
-
 
 class AOISelection(TypedDict):
     name: str
@@ -29,6 +27,11 @@ class Statistics(TypedDict):
     context_layer: str | None
 
 
+class EncodedCodeActPart(TypedDict):
+    type: str
+    content: str
+
+
 class AgentState(TypedDict):
     messages: Annotated[Sequence[BaseMessage], add_messages]
     user_persona: str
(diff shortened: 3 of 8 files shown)


.github/workflows/tools-tests.yml
```diff
@@ -75,4 +75,4 @@ jobs:
         run: |
           mkdir -p data
           uv run python src/ingest/embed_datasets.py
-          uv run pytest tests/tools tests/agent -v
+          uv run pytest tests/tools tests/agent -v --reruns 2 --reruns-delay 5
```
jterry64, line 78: This is the only change I'm unsure about, because it'll make it take twice as long to finish the run, and I think 10+ tests will just faill both times. 

I do notice it's not totally random, the same 10-15 always seem to fail, so I can go through today and just fix them up or mark them as "flaky" and skip here. Feel free to merge this change or remove it and I'll work after this is merged.

jterry64 (approved):

yellowcap:
Yes we should minimize the number of times the rerun is necessary. So lets skip or fix tests that are consistently failing in a fillow up pr. And the retry will be for the true random failures.

PR #634, opened 2026-04-22
Remove whitelist/anonymous access paths and enforce authenticated-only API

yellowcap:
## Summary

This PR removes legacy access-control paths and fully standardizes the API on authenticated-user access only.

### What changed

- Removed whitelist/signup gating logic from API auth flow:
  - deleted whitelist checks and signup-limit gating from auth dependency flow
  - removed obsolete auth service module (`src/api/services/auth.py`)
- Removed anonymous-user API paths:
  - `/api/chat` and `/api/quota` now require authenticated users
  - removed anonymous token/header handling and anonymous quota identity logic
  - removed anonymous execution branch in chat service
- Cleaned up API config and metadata:
  - removed whitelist/anonymous-related settings and metadata fields
- Removed whitelist operator tooling:
  - deleted `whitelist-email` CLI command and related helper logic
- Database cleanup:
  - removed `WhitelistedUserOrm` from models
  - added migration to drop `whitelisted_users` table
- Security fix:
  - fixed thread delete flow so checkpoint deletion only occurs **after** thread ownership/authorization is verified
  - added regression coverage to ensure unauthorized deletes do not delete checkpoints
- Updated docs/CI/env/tests to match authenticated-only contract.

## Why

- Whitelist and anonymous-user support are no longer needed.
- Simplifies auth logic and reduces maintenance surface.
- Eliminates dead configuration and operational commands.
- Closes a high-priority authz sequencing bug in thread deletion.

## Test plan

- `uv run ruff check src/api tests/api tests/cli tests/tools`
- `uv run pytest tests/api/test_auth.py tests/api/test_anonymous_users.py tests/api/test_deployment_phases.py tests/api/test_quotas.py tests/api/test_metadata.py tests/api/test_auto_naming.py tests/cli/test_machine_user_cli.py -q`
- `uv run pytest tests/api/test_threads.py -q`

## Notable behavior changes

- `/api/chat` now rejects unauthenticated requests.
- `/api/quota` now rejects unauthenticated requests.
- `/api/metadata` no longer returns signup/anonymous flags.
- Whitelist table and model are removed (via migration + model cleanup).

## Follow-up / future improvement flag

- **Token-cache timeout hardening (future):** current auth token cache behavior can be tightened further by shortening TTL and/or using token-expiry-aware cache bounds to reduce stale-auth windows after upstream token revocation.  
  - Suggested follow-up: move from long fixed TTL to short TTL (e.g., 5–15 min) with optional `exp`-bounded caching and negative-cache handling for invalid tokens.


jterry64 (approved):
Thank for your patience! Appreciate the detailed list of changes. This actually all looks good to me, nothing here looks too dangerous and seems well tested. Good work!

Issue #584, opened 2026-03-25
Standalone insights and new agent persona


yellowcap:
# ADR-001: Standalone Insight Artifacts

**Status:** Proposed
**Date:** 2026-03-25

---

## Context

Zeno currently operates as a conversational analyst: the agent pulls data, runs analysis, and returns chart data embedded in the LangGraph conversation state. When the conversation ends or the thread is discarded, the insight is gone. Retrieving it later requires loading the full checkpointer state for that thread.

This creates several problems:

- **Insights are not reusable.** A chart generated in one conversation cannot be referenced or rendered in another without replaying the thread.
- **Insights are not self-contained.** To understand what a chart shows — what data it was built from, what code produced it — you need the surrounding conversation context.
- **Insights cannot be edited.** There is no mechanism for the agent or user to refine a chart after it is produced. A new conversation run is required.
- **The agent is framed as an analyst.** Its outputs are answers, not artifacts. This limits the paradigm to Q&A rather than collaborative content creation.

The desired shift is to treat insights the same way an IDE assistant treats code edits: the agent helps create a persistent artifact that lives independently of the conversation that produced it. The conversation is the working context; the insight is the output.

---

## Decision

Refactor `generate_insights` to persist each insight as a database record (`InsightOrm`) and return only its UUID in agent state. The frontend fetches the insight by ID from a dedicated API endpoint rather than reading it from thread state.

This mirrors the existing pattern for custom areas (`CustomAreaOrm`), which are also created by the agent and fetched independently.

---

### 1. Database: `InsightOrm`

New table in `src/api/data_models.py`:

| Field | Type | Notes |
|---|---|---|
| `id` | UUID | Auto-generated PK |
| `user_id` | String FK, nullable | Nullable to support anonymous sessions |
| `thread_id` | String FK, nullable | For traceability back to the originating conversation |
| `query` | String | The user's original question — makes the insight self-explanatory without the conversation |
| `inputs` | JSONB | `{aoi_names, dataset_name, start_date, end_date, source_urls}` — full data provenance |
| `title` | String | |
| `chart_type` | String | |
| `insight` | String | 2-3 sentence key finding |
| `follow_up_suggestions` | JSONB | Suggested next steps generated alongside the insight |
| `chart_config` | JSONB | `{xAxis, yAxis, seriesFields, groupField, colorField, stackField}` |
| `data` | JSONB | Array of chart data rows |
[...]

PR #635, opened 2026-04-22
Multichart improvement

solomon-negusse:
This addresses issue with TCL charts sometimes not including carbon chart and coming up with wrong titles as a result (see[ pzb-787](https://gfw.atlassian.net/browse/PZB-787))

### DIFF
--- a/src/agent/tools/datasets/tree_cover_loss.yml
+++ b/src/agent/tools/datasets/tree_cover_loss.yml
@@ -24,8 +24,8 @@ keywords:
 - forest loss
 - disturbance
 analytics_api_endpoint: /v0/land_change/tree_cover_loss/analytics
-prompt_instructions: 'Reports gross annual loss of tree cover ≥ 5 m height (2001-2025). Show yearly loss with bar charts or
-  cumulative loss with stacked-area charts; plot area_ha and drop zeros. When reporting tree cover loss, always include the
+prompt_instructions: 'Reports gross annual loss of tree cover ≥ 5 m height (2001-2025). Show yearly loss with bar charts;
+  for multiple regions over time, use stacked bar charts; plot area_ha and drop zeros. When reporting tree cover loss, always include the
   GHG emissions associated with that loss, in MgCO2e (tonnes of CO2 equivalent). If users ask for intra-year or seasonal tree
   cover loss OR emissions, refuse and remind that the data is annual only. The 2011-2025 data was produced using an updated
   methodology so comparisons between, and trends across the periods 2001-2010 and 2011-2025 should be performed with caution.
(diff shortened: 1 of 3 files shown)


src/agent/tools/datasets/tree_cover_loss.yml
```diff
@@ -43,19 +43,35 @@ selection_hints: 'Best dataset for annual tree cover loss questions (2001-2024)
...
 
   '
-code_instructions: 'CHART TYPES: - Yearly loss → bar chart (x=year, y=area_ha) - Cumulative loss → stacked-area chart - Emissions
-  → MUST be a SEPARATE chart from loss area (NEVER combine on same chart) DATA RULES: - Always include GHG emissions (MgCO2e)
-  alongside loss area, but in a SEPARATE chart - Drop rows where area_ha = 0 - Use 30% canopy density threshold - If primary_forest
-  or intact_forest variable is active, filter accordingly DO/DON''T: - DO NOT use the term "deforestation" — always use "tree
-  cover loss" - If user asks for intra-year or seasonal breakdown, REFUSE — data is annual only - If user asks for net gain/loss,
-  REFUSE — methodologies differ between loss and gain datasets - Warn about 2001-2010 vs 2011-2024 methodology difference
-  when showing trends across both periods
+code_instructions: 'MANDATORY OUTPUT: Always produce EXACTLY 2 charts (one for area_ha, one for carbon_emissions_MgCO2e).
```
yellowcap, line 46: Do we really want that? What if I want only area or only co2?

src/agent/tools/datasets/tree_cover_loss.yml
```diff
@@ -43,19 +43,35 @@ selection_hints: 'Best dataset for annual tree cover loss questions (2001-2024)
...
+  NEVER combine area_ha and carbon_emissions_MgCO2e on the same chart. Even if the query does not mention emissions,
+  the carbon_emissions_MgCO2e column is ALWAYS required in the output data.
+
+  CHART TYPE & AXIS SELECTION (contextual based on query):
+
+  Case 1: Single AOI, time series (MOST IMPORTANT) — Bar chart, x=year, y=area_ha (separate chart for emissions with x=year).
```
yellowcap, line 52: Not a huge fan of this kind of prompting, this is a bit like "coding in prompts". This will get messy and unstable over time quite quickly. Especially when added on a per-dataset level in instructions. This might apply to more than one dataset, and we should not add more and more detail instructions everywhere.

yellowcap (commented):
I would not add this kind of instructions. If this is high priority for the TCL 25 release, we can let it pass. But we should revisit this soon.

This is exactly why we need a simpler way to inject stuff that is more systematic, see #584

src/agent/tools/datasets/tree_cover_loss.yml
solomon-negusse: Yes makes sense, I wasn't too happy about this either. I'll do a separate PR with narrower scope that fixes the bug.

src/agent/tools/datasets/tree_cover_loss.yml
solomon-negusse: I explained the reason in the slack message.

yellowcap:
@solomon-negusse is this still relevant? If not can we close this PR?

yellowcap:
Closing as stale. @solomon-negusse please re-open if still relevant.

PR #637, opened 2026-04-24
Fix tools tests for ci

yellowcap:
This PR improves the tools test stability.

- Remove google api patches from agent test (these were making the test fail with async)
- Remove 2 tests for picking datasets that were not well formulated and failed
- Slight improvement on wording for TCL vs TCL by driver for specific dates
- Make pytest naming of the dataset pick tests human readable (names instead of index nrs)


01painadam:
Looks good to me - much cleaner!

01painadam (approved):

PR #638, opened 2026-04-24
PZB-781: Default dataset dates

jterry64:
Right now, the agent must choose start and end dates to call pick dataset and pull data. This leads to some "prompt pressure" to choose a date even if the user didn't specify, and so it'll sporadically choose between asking the user for clarification or selecting random dates.

This PR makes the dates now optional, and it'll re-use the existing function for checking against the dataset's full date range to choose a default if not specified.


yellowcap (approved):
Nice work @jterry64 !

PR #640, opened 2026-04-27
Update CLI command paths to use src/api/cli.py

sunu:


batpad (approved):
I bumped into this as well - thanks much for fixing @sunu !

PR #642, opened 2026-04-27
PZB-895 - Persist pull-data statistics and link insights to source data

yellowcap:
## Summary

- Add a dedicated `statistics` database table for outputs produced by `pull_data`, including dataset metadata, date range, AOIs, parameters, context layer, source URL, and the raw statistics payload.
- Keep the existing inline `state["statistics"]` payload for backwards compatibility, while adding the persisted statistics row ID to each new statistics object returned by `pull_data`.
- Add `statistics_ids` to persisted insights so each generated insight can track which statistics objects were used as source data.
- Preserve backwards compatibility for older thread states and tests where statistics objects do not yet include IDs.
- Expose `statistics_ids` in insight API responses so clients can inspect provenance from fetched insights.
- Add an Alembic migration for the new `statistics` table, indexes on `user_id` and `thread_id`, and the new `insights.statistics_ids` provenance column.
- Add focused test coverage for statistics persistence, insight provenance tracking, and older state compatibility.

## Implementation Details

- Introduces `StatisticsOrm` in the SQLAlchemy data models with UUID primary keys and JSONB columns for raw data, AOI names, and optional parameters.
- Updates `pull_data` to persist successful analytics results through the shared DB session pool using the current structlog context for `user_id` and `thread_id`.
- Updates `generate_insights` to extract statistics IDs from agent state and write them to `InsightOrm.statistics_ids`.
- Adds a helper for extracting statistics IDs so provenance behavior is easy to test independently from live LLM execution.
- Leaves existing state consumers unchanged by keeping `data`, `dataset_name`, `start_date`, `end_date`, `source_url`, `aoi_names`, `parameters`, and `context_layer` in the state update.

## Test Plan
- `uv run ruff check src/api/data_models.py src/api/schemas.py src/api/routers/insights.py src/agent/state.py src/agent/tools/pull_data.py src/agent/tools/generate_insights.py tests/tools/test_pull_data.py tests/tools/test_generate_insights_tiered.py db/alembic/versions/9f8e7d6c5b4a_add_statistics_table.py`
- `uv run pytest tests/tools/test_generate_insights_tiered.py -k 'statistics_provenance or extract_statistics_ids'`
- `uv run pytest tests/tools/test_pull_data.py -k 'persists_statistics'`

Refs https://github.com/wri/project-zeno/pull/628

### DIFF
--- a/db/alembic/versions/9f8e7d6c5b4a_add_statistics_table.py
+++ b/db/alembic/versions/9f8e7d6c5b4a_add_statistics_table.py
@@ -1,0 +1,75 @@
+"""add_statistics_table
+
+Revision ID: 9f8e7d6c5b4a
+Revises: 75841e9932e5
+Create Date: 2026-04-27 10:49:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = "9f8e7d6c5b4a"
+down_revision: Union[str, None] = "75841e9932e5"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "statistics",
+        sa.Column(
+            "id",
+            sa.UUID(),
+            server_default=sa.text("gen_random_uuid()"),
+            nullable=False,
+        ),
+        sa.Column("user_id", sa.String(), nullable=True),
+        sa.Column("thread_id", sa.String(), nullable=True),
+        sa.Column("dataset_name", sa.String(), nullable=False),
+        sa.Column("start_date", sa.String(), nullable=False),
+        sa.Column("end_date", sa.String(), nullable=False),
+        sa.Column("source_url", sa.String(), nullable=True),
+        sa.Column(
+            "data",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=False,
+        ),
+        sa.Column(
+            "aoi_names",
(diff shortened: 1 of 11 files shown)


yellowcap:
## Where to store the data

We discussed where to store the data. The two options are 

1. Permanent links on data-api results
2. On GNW side
  - In PostGres
  - In S3

### Proposal

Store data for each insight in S3 independently on the GNW side.

### Pros

- Long term sustainability is GNW responsability direclty
- Better decoupling from api
- Easier to add more providers later

### Cons

- Duplication of data for similar queries
- Increased complexity on GNW side
- Creates multiple sources of truth

jterry64:
I assume this is like just directly from the analytics API, not any post-processing from codeact? If that's the case, we are already providing forever links. When we change versions of data, it changes the resource ID, so the old link still provides a snapshot of the data when the request was made. Is my understanding correct there, @gtempus?

So I wouldn't duplicate it for now, especially because I think long term the goal would be to integrate the analytics more into the rest of GNW instead of keeping it as a separate API. That seems like more work upfront and will likely be changed later. However, if we also want to store statistics where the LLM has done some kind of post-analysis (aggregate/filter, combine multiple results, etc.) it might be worth storing that in relation to a specific insight.

@gtempus may also have some thoughts.

yellowcap:
ok perfect, I updated the PR to rely on the analytics api url. That is super clean, happy about this!

gtempus:
Thanks, @yellowcap. I think this is a good step. 
If performance or availability concerns arise, we can consider persisting analytics on the agent side. 
No need to do that now.

jterry64 (approved):
I wasn't sure if you were saying this was ready for review, but this all LGTM!

batpad (approved):

PR #644, opened 2026-04-27
Add small instruction to mention context layer

jterry64:
Just tested this through local dev to see that it mentions it in the summary consistently but without overdoing it.


solomon-negusse (approved):

yellowcap (approved):
🛼 let's roll

PR #645, opened 2026-04-27
PZB-780: Consistently select context layers or parameters if they apply

jterry64:
In PZB-780, we discovered context layers are applied inconsistently. E.g. asking for "deforestation in Brazil" leads to primary forest context layer selection, but not "deforestation in Brazil in 2024". 

Following recommendation from a different LLM, I tweaked the prompt to more explicitly choose a context layer even if not mentioned by name, to use the description when deciding, and to do it after selecting the dataset so it didn't confuse instructions.

I also removed the instruction about choosing date range over context layer, and rephrased as prioritizing date range in dataset selection, since I think that's what was trying to be expressed given the example.

@yellowcap I removed a couple tool tests that were flaky, and seemed to me like the answer the test wanted wasn't really clear cut. Like "Show the trend in natural land loss over time in Brazil" is asking for tree cover loss, when it could easily be DIST or land cover change, right? That might be a better question for an eval. "Show deforestation by driver in 2019" also isn't clear to me we should show TCL rather than TCL by driver with a corrected date range.

### DIFF
--- a/src/agent/tools/pick_dataset/tool.py
+++ b/src/agent/tools/pick_dataset/tool.py
@@ -91,19 +91,19 @@ async def select_best_dataset(
     user query and provide reason why it is the best match. Always return at least one dataset.
     Use all information provided to decide which dataset is the best match, especially the selection hints.
 
+    Evaluate if the best dataset is available for the date range requested by the user.
+    If not, pick the closest available date range and include a warning in the dataset pick reason.
+    Pick the most granular dataset. Prioritize matching the date range when candidate datasets are similar.
+
     After selecting the best dataset, inspect its filtered_context_layers. Choose the single context layer whose description
     best fits the user's query, even when the user does not name the layer directly. Context layers differentiate types of
-    data within the same dataset, and their descriptions are guidance for when each layer applies. If no context layer fits the query, return null.
+    data within the same dataset, and their descriptions are guidance for when each layer applies.
+    Pick the most granular context layer that matches the query. If no context layer fits the query, return null.
 
     After selecting the best context layer or null, select parameters and values if they are relevant or specified in the user query. Parameters allow further filtering
     the analysis to better answer the query. Select only values listed in the value field for a parameter. For example,
     if a user says "show me tree cover loss in forests where canopy cover is greater than 50%", you may select the parameter canopy cover
-    and value 50.
-
-    Evaluate if the best dataset is available for the date range requested by the user.
-    If not, pick the closest available date range and include a warning in the dataset pick reason.
-
-    Pick the most granular dataset/contextual layer/parameters that matches the query.
+    and value 50. Pick the most granular parameters that matches the query.
 
     Keep explanations concise. Do not use datset IDs to describe the dataset.
     For instance, instead of saying "Dataset ID: 123", say "Dataset: Tree Cover Loss".
--- a/src/agent/tools/pick_dataset/tool.py
+++ b/src/agent/tools/pick_dataset/tool.py
@@ -94,7 +94,7 @@ async def select_best_dataset(
     Evaluate if the best dataset is available for the date range requested by the user.
     If not, pick the closest available date range and include a warning in the dataset pick reason.
     Pick the most granular dataset. Prioritize matching the date range when candidate datasets are similar.
-    
+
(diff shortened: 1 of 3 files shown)


tests/tools/test_pick_dataset.py
```diff
@@ -194,10 +194,6 @@ def _query_case_id(param):
             "I need to track forest plantations harvesting cycles in northern Europe",
             TREE_COVER_LOSS,
         ),
-        (
-            "Show deforestation by driver in 2019",
```
yellowcap, line 198: We added this as a regression test, we actually want this to select TLC, not TLC by driver. So time specificity wins over context layer specificity. That was a preference from @01painadam. So this one should stay and pass in my opinion.

src/agent/tools/pick_dataset/tool.py
```diff
@@ -91,23 +91,19 @@ async def select_best_dataset(
...
 
-    Pick the most granular dataset/contextual layer/parameters that matches the query.
+    After selecting the best dataset, inspect its filtered_context_layers. Choose exactly one context layer whenever its description
+    or instructions apply to the query, even if the user does not mention it by name. Return null only if no available context layer is relevant.
+    Context layers differentiate types of data within the same dataset, and their descriptions are guidance for when each layer applies.
+    Pick the most granular context layer that matches the query. If no context layer fits the query, return null.
 
-    Give more importance to date matching than context layer matching. For instance, dont select tree cover
-    loss by driver if the user requests a specific time range, pick tree cover loss instead.
+    After selecting the best context layer or null, select parameters and values if they are relevant or specified in the user query. Parameters allow further filtering
```
yellowcap, line 103: That is not a sequence, both should be treated independently right?

```suggestion
    Select parameters and values if they are relevant or specified in the user query. Parameters allow further filtering
```

src/agent/tools/pick_dataset/tool.py
```diff
@@ -91,23 +91,19 @@ async def select_best_dataset(
...
     If not, pick the closest available date range and include a warning in the dataset pick reason.
+    Pick the most granular dataset. Prioritize matching the date range when candidate datasets are similar.
 
-    Pick the most granular dataset/contextual layer/parameters that matches the query.
+    After selecting the best dataset, inspect its filtered_context_layers. Choose exactly one context layer whenever its description
+    or instructions apply to the query, even if the user does not mention it by name. Return null only if no available context layer is relevant.
+    Context layers differentiate types of data within the same dataset, and their descriptions are guidance for when each layer applies.
+    Pick the most granular context layer that matches the query. If no context layer fits the query, return null.
 
-    Give more importance to date matching than context layer matching. For instance, dont select tree cover
```
yellowcap, line 109: This stands in relation with the regression test mentioned below. This is supposed to explain the date vs context param behaviour.

src/agent/tools/pick_dataset/tool.py
```diff
@@ -91,23 +91,19 @@ async def select_best_dataset(
...
-    the analysis to better answer the query. Select only values listed in the value field for a parameter. For example,
-    if a user says "show me tree cover loss in forests where canopy cover is greater than 50%", you may select the parameter canopy cover
-    and value 50.
-
     Evaluate if the best dataset is available for the date range requested by the user.
     If not, pick the closest available date range and include a warning in the dataset pick reason.
+    Pick the most granular dataset. Prioritize matching the date range when candidate datasets are similar.
 
-    Pick the most granular dataset/contextual layer/parameters that matches the query.
+    After selecting the best dataset, inspect its filtered_context_layers. Choose exactly one context layer whenever its description
```
yellowcap, line 98: Also here slightly unsure if we should phrase it as "after selecting dataset" vs "as part of selecting a dataset".

yellowcap (commented):
A couple of comments on this, the main one being not removing a regression test we added previously. @01painadam maybe advise on how important the date vs context layer hierarchy is.

tests/tools/test_pick_dataset.py
jterry64: Oh ok, I was curious about this. I'll try throwing the example back in.

jterry64:
Going to wait on this PR, let's see how it's doing in prod.

PR #646, opened 2026-04-27
PZB-781: Clamp to contextual layer start/end dates

jterry64:
Consider contextual layer start/end dates when clamping the date, and add a start date of 2002 for primary forest.


yellowcap (approved):
Looks good thanks @jterry64

PR #647, opened 2026-04-28
instructions for chart titles

solomon-negusse:
This is potential fix for the issue of tree cover loss chart title including carbon emissions that's not shown on the chart. I haven't been able to repro it after several attempts with many prompt variations including the one in the reported bug. So this can be tested by others once merged.


yellowcap (approved):
Looks good thanks @solomon-negusse

PR #649, opened 2026-04-29
Pr 648

01painadam:
From #648 (fork)

Added description of commodities included, detailing which crops are included in the grouped categories. This will allow GNW to provide values for crops like avocado which is not included on an individual basis, but is part of the tropical fruit categories


yellowcap:
@01painadam can we close this or is it still useful?

PR #650, opened 2026-04-29
Fixing type errors, so mypy is clean

danscales:
Eliminated all the mypy errors when running this command:

cd src; mypy --exclude shared/config.py .

[In my mypy/config, I turn off warning about missing imports, especially library stubs, and about checking untyped functions.]

Among other things:
 - Correct the return types of _get_aoi_type(), _process_response_data(), replace_csv_paths_with_urls, check_duplicate_aois(), and prepare_dataframes().
 - Mark tool_call_id as Optional in generated_insights(), pick_aoi(), pick_dataset(), and pull_data(), but make sure (and make clear) that state is NOT Optional.
 - Fixed arg type of query_subregion_database()
 - Add type for _user_info_cache
 - Fix the type of 'authorization' param of fetch_user_from_rw_api()
 - Create different variable destpath with type Path in download()

All the tests/{api, cli, unit, tools, agent} still pass, except for 3 that were already broken.

### DIFF
--- a/src/agent/tools/code_executors/gemini_executor.py
+++ b/src/agent/tools/code_executors/gemini_executor.py
@@ -3,7 +3,7 @@
 import asyncio
 import io
 from functools import partial
-from typing import Dict, List
+from typing import Dict, List, cast
 
 import pandas as pd
 from google import genai
@@ -120,7 +120,7 @@ class GeminiCodeExecutor:
                         f"retrying in {delay:.1f}s"
                     )
                     await asyncio.sleep(delay)
-        raise last_error
+        raise cast(Exception, last_error)
 
     def _parse_response(self, response) -> ExecutionResult:
         """Parse a generate_content response into ExecutionResult."""
--- a/src/agent/tools/data_handlers/analytics_handler.py
+++ b/src/agent/tools/data_handlers/analytics_handler.py
@@ -166,7 +166,7 @@ class AnalyticsHandler(DataSourceHandler):
             SLUC_EMISSION_FACTORS_ID,
         ]
 
-    def _get_aoi_type(self, aoi: Dict) -> str:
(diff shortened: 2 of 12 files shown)


yellowcap (changes requested):
Amazing this is a huge step forward in code reliability! @danscales  thanks for doing this. Now that mypy passes let's add this also to our github actions CI flow so that it stays that way 😀
What do people think about adding this to pre-commit as well? Pre commit can get in the way for rough draft PRs, but i am up to have it in there, it makes sense to me.

dmannarino (approved):
This is great, thanks Dan! I see @yellowcap suggested putting type checking in CI, but I would suggest that as a follow-up so as not to hold things up. Also, have you tried ty or pyrefly? ty has a benefit of ruff/uv integration, and both ty and pyrefly are loads faster than mypy. FWIW ty agrees that your changes are good and solve a bunch of type errors.

jterry64:
I also like pre-commit, I always forget to run the linter and then embarrassingly need to send a follow up commit... you can set specifics linters to SKIP for draft PRs.

danscales:
> This is great, thanks Dan! I see @yellowcap suggested putting type checking in CI, but I would suggest that as a follow-up so as not to hold things up. Also, have you tried ty or pyrefly? ty has a benefit of ruff/uv integration, and both ty and pyrefly are loads faster than mypy. FWIW ty agrees that your changes are good and solve a bunch of type errors.

I agree with DanielM that it makes sense to check this in now, and then decide on the CI work (especially since DanielM is quite good at github actions, at this point.... :-))

DanielM, I haven't tried ty or pyrefly, so maybe you can decide if it makes sense to use one of those instead of mypy.

yellowcap (approved):
Ok great lets move and add this to CI next.

danscales:
Thanks for the reviews!

PR #652, opened 2026-05-04
PZB-815: build aoi bbox when aoi is selected in the UI

solomon-negusse:
When users select AOI manually on the UI, the bbox is missing from the AOI object in the state which breaks `pick_dataset` tool when getting contextual layers that overlap with the AOI.  This resolves that by injecting the bbox into the AOI object from the `ui_context`.

trace with the issue:

Screenshot 2026-05-05 at 3 49 21 PM

trace with the fix:

Screenshot 2026-05-05 at 3 50 07 PM


jterry64 (approved):
Nice, this looks good to me!

PR #654, opened 2026-05-06
Protype of new agent harness design

srmsoumya:
# Zeno Agent Harness - Concepts

Five ideas drive the design to make the agent independent of the frontend, specific llm providers & custom prompts.

## 1. State holds refs, not raw data

```python
class AgentState(TypedDict):
    aoi_refs:     list[AoiRef]   # {name, source, src_id, subtype}
    dataset_id:   str | None
    data_refs:    list[str]      # stat_ids → data_cache
    artifact_ids: list[str]      # ids → artifacts table
```

Heavy stuff (rows, chart specs, insight text) lives in the backend, keyed
by id. State stays small forever; checkpoints are cheap; the LLM context
never sees raw rows.

## 2. Current state auto-loads every turn

A langchain middleware prepends a session block before each model call:

```
[Session - 2026-05-06]
AOI: Para (gadm:BRA.14_1)
Dataset: tree_cover_loss
Data: stat_3a4f...
Active artifact: @art_b7c2...
```
Context is always there for the agent to make informed decisions based on the current state for user queries.

## 3. Orchestrator = tools + subagents + skills

The orchestrator prompt lists what is available; it does not script a
sequence. Each layer is one job.

### Tools - primitives, no LLM, fast

| Tool | What it does | Example |
|---|---|---|
| `list_datasets(query)` | Search dataset catalog. | `list_datasets("tree cover loss")` |
| `fetch(aoi_refs, dataset_id, dates)` | Pull data into cache, return stat_id. | `fetch(refs, "tree_cover_loss", "2024-01-01", "2024-12-31")` |
| `execute(code, stat_ids)` | Pandas snippet over cached data. | `execute("df['area_ha'].sum()", ["stat_3a4f"])` |
| `get_artifact(id)` | Load an artifact for inspection. | `get_artifact("art_b7c2")` |
| `update_artifact(id, changes)` | Presentation-only patch. | `update_artifact("art_b7c2", {chart_type: "pie"})` |
| `zoom_map(aoi_refs)` | Side-effect: pan the map. | `zoom_map(refs)` |
| `read_skill(name)` | Lazy-load full skill body. | `read_skill("analyze")` |

### Subagents - own reasoning, called as tools

| Subagent | What it does | Example |
|---|---|---|
| `geo_subagent(query)` | Place name → AoiRefs. | `geo_subagent("neighbours of Para")` |
| `analyst_subagent(task, stat_ids, ...)` | Build a chart artifact from cached data. | `analyst_subagent("compare drivers", [stat_id])` |

We will use DSPy to design sub-agents, keeps them small, modular & will help us to run prompt optimization loop if required.

### Skills - multi-step recipes (markdown files)

| Skill | When to use |
|---|---|
| `analyze` | Full pipeline: place → dataset → fetch → chart. |
| `compare` | Multi-AOI or multi-dataset comparison. |
| `explore` | Discover what is available for an area. |

Skill metadata sits in the system prompt; the body loads on demand via
`read_skill`. Drop a new `.md` file in `skills_md/` - it appears
automatically next run.

Our current zeno agent fits into `analyze` skill based on this design.

## 4. Artifacts are addressable (`@art_xxxx`)

Every analyst output is an `Artifact` with a short id (`art_010`).
Edits never mutate; they create a new artifact with `parent_id` pointing
at the original - lineage is intact.

The user can refer to `@art_010` directly. The orchestrator either
trusts the active-artifact field in the session block, or calls
`get_artifact` to inspect the spec before changing it. Title is
whatever the user asked for - not a model-invented headline.

## 5. State for persistence, streaming via LangGraph native

The idea is to keep the agent independent of FE, so we have a CLI, slack
bot, or React FE - each can decide how to render.

**State** persists what survives a refresh: `aoi_refs`, `dataset_id`,
`data_refs`, `artifact_ids`. Tools update state by returning
`Command(update={...})` - LangGraph handles reducer logic and
checkpointing automatically.

**Streaming** uses LangGraph's native `astream(stream_mode=["updates", "custom"])`:

- `updates` stream: state deltas, messages, tool calls/results
- `custom` stream: tool-emitted events via `runtime.stream_writer({...})`

Custom stream events emitted by tools:

| Event type | Meaning | Emitter |
|---|---|---|
| `aoi_resolved` | A place was resolved. | `geo_subagent` |
| `data_fetched` | A dataset slice landed in cache. | `fetch` |
| `artifact` | A new artifact is ready to render. | `analyst_subagent`, `update_artifact` |
| `zoom_map` | Pan the map to AOI. | `zoom_map` |

Any frontend consumes the same stream:

```python
async for event in agent.astream(inputs, stream_mode=["updates", "custom"]):
    mode, payload = event
    if mode == "custom":
        handle_custom_event(payload)  # artifact, zoom, etc.
    elif mode == "updates":
        handle_state_and_messages(payload)
```

- React renders charts when `artifact` custom event arrives.
- `python -m src.agent.harness.cli` prints labeled sections.
- A future Streamlit / WhatsApp / Slack client writes its own renderer.

cc @yellowcap @01painadam @gtempus @jterry64


jterry64:
@srmsoumya awesome, looks super cool. Will try running it locally today and playing around with it. Were you able to take any specific measurements or evaluations on its performance?

jterry64:
This is neat to play with, the CLI is convenient, although it rook me a sec to realize the mock was always returning drivers lol. It's definitely better now for doing things other than full analysis, and I prefer the tools being lightweight and not making other LLM calls. The skills also seem more maintainable and easier to scale to new flows. The only thing I'm unsure on is independence from the frontend - while that's useful, it seems a bit different than the other goals, and would it require reworking our existing frontend?

For following up, if you're trying to make it use more real data, I think it's going to be important we have clear evaluations and metrics for what we're trying to improve over the current implementation. We're working on getting the gold set this sprint to near 100%, so that'd be a good baseline, but it might be more useful to have to 10-20 more challenging tasks to demonstrate how this improves our current limitations. I had started coming up with this set here based on real user traces, might be a good starting point: https://docs.google.com/spreadsheets/d/1_G1aq2fSCPqhT6w55_Od6VU7sov76t1lHQTBeZZxbdM/edit?gid=2033513940#gid=2033513940

For metrics, I'm thinking these are the most important:
- Task completion/helpfulness for more ambiguous or granular tasks 
- Latency, since the tools are lighter weight
- One of your listed goals is independence from LLM provider, so perhaps sensitivity when changing the base model?
Anything else you can think of? Glad to help implement these evals so we can test this properly.

In parallel, we might try to prototype some of these ideas inside the current implementation to better understand how we can evolve the architecture. We're thinking:
- Adding a skills.md on top of existing tools
- Exposing some of the agent state to the agent
- Moving out some of the internal LLM calls

It'd be great if we could agree on what we're evaluating on and then we can compare your next prototype with our prototype.

cc @gtempus

src/agent/harness/tools/list_datasets.py
```diff
@@ -0,0 +1,71 @@
...
+                "end_date": str(doc.get("end_date") or ""),
+            }
+        )
+    return catalog
+
+
+_CATALOG = _load_catalog()
+
+
+def _score(item: dict, query: str) -> int:
```
jterry64, line 36: @srmsoumya this one caught me a bit by surprise, I was curious you avoided the embeddings search. This doesn't seem like it'll perform well for different wordings or languages. Was there a plan on where to go from here?

src/agent/harness/backends/protocol.py
```diff
@@ -0,0 +1,17 @@
+from typing import Protocol
+
+from src.agent.harness.artifact import Artifact
+
+
+class ZenoBackend(Protocol):
```
yellowcap: This is never used anywhere

src/agent/harness/backends/memory.py
```diff
@@ -0,0 +1,26 @@
+from src.agent.harness.artifact import Artifact
+
+
+class InMemoryBackend:
```
yellowcap: This is only used in tests, so let's move it there.

src/agent/harness/session.py
```diff
@@ -0,0 +1,15 @@
+"""Session module - DEPRECATED.
```
yellowcap: Remove this file

src/agent/harness/prompts.py
```diff
@@ -0,0 +1,43 @@
...
+{skills_block}
+
+When a skill matches the user's intent, call read_skill(name) once to load it, then follow its workflow.
+
+# Rules
+
+- Do not paste raw data rows into the conversation. Numbers go through execute or analyst_subagent.
+- For data, AOI, or date-range changes always run a fresh fetch + analyst_subagent. Only use update_artifact for cosmetic changes to an existing chart.
+- Keep replies short. The artifact carries the chart, insights and follow-ups.
+- If the user references @art_xyz, look it up with get_artifact before deciding what to do.
```
yellowcap, line 42: This failed due to prallel tool calls when using `gemini-flash`. In previous agent parallel tool calling would fail due to state updates. Maybe we can loosen this up here? The current ipmlementation does not work though, so I added this to the system prompt for local testing with flash.

```suggestion
- Never do tool calls in parallel. Always call tools one at a time.
```

src/agent/harness/tools/zoom_map.py
```diff
@@ -0,0 +1,13 @@
+from langchain.tools import ToolRuntime, tool
+
+
+@tool
+async def zoom_map(
```
yellowcap, line 5: I vote to remove this for now. I guess this is here to show how frontend updates _could_ be done, but this is not a realistic example (the frontend does this already when picking an aoi). Let's keep things minimal so that we have the most backwards comparability.

src/agent/harness/models.py
```diff
@@ -0,0 +1,24 @@
...
+class ModelRegistry:
+    """Thin wrapper over the existing project-wide model registry. Phase 1
+    only supports LangChain chat models for the orchestrator and any future
+    code paths; DSPy support will land when subagents move off stubs."""
+
+    def __init__(self, mapping: dict[str, str] | None = None) -> None:
+        self._mapping = mapping or {
+            "orchestrator": "sonnet",
+            "geo": "haiku",
+            "analyst": "sonnet",
```
yellowcap, line 15: Let's use a realistic model during development. I understand that this should be more agnostic than the current implementation, but the personality is different so we should still work with what we are most likely going to use. I already found the parallel tool call issue after switching to gemini flash.

```suggestion
            "orchestrator": "gemini-flash",
            "geo": "gemini-flash",
            "analyst": "gemini-flash",
```

src/agent/harness/protocol.py
```diff
@@ -0,0 +1,27 @@
+"""Lightweight data models for the Zeno harness.
```
yellowcap: Remove this file, none of the classes are used anywhere

src/agent/harness/subagents/wrap.py
```diff
@@ -0,0 +1,72 @@
...
+            "messages": [ToolMessage(
+                content=json.dumps(ref_dicts),
+                tool_call_id=runtime.tool_call_id,
+            )],
+        })
+
+    return geo_subagent
+
+
+def make_analyst_tool(analyst: AnalystAgent):
```
yellowcap: I don't see added value in this constructor, it adds unecessary complexity.

src/agent/harness/tools/fetch.py
```diff
@@ -0,0 +1,80 @@
...
+        cur = min(d1, cur + timedelta(days=step))
+    return rows
+
+
+@tool
+async def fetch(
+    aoi_refs: list[dict],
+    dataset_id: str,
+    start_date: str = "2024-01-01",
+    end_date: str = "2024-12-31",
```
yellowcap, line 53: This needs to be more structured, picking a dataset now includes parameters, context_layers, and the dataset itself

src/agent/harness/tools/fetch.py
```diff
@@ -0,0 +1,80 @@
...
+                "confidence": "high",
+            }
+        )
+        cur = min(d1, cur + timedelta(days=step))
+    return rows
+
+
+@tool
+async def fetch(
+    aoi_refs: list[dict],
```
yellowcap, line 50: Here we are asking the agent to inject entire dictionaries for aois. This is not the right approach, we should only pass ids. Similar is true in other tools as well.

yellowcap (commented):
I like the approach a lot, this is more flexible and a big step into the right direction.

My main concern is that it looks like we are not  reading the data from state, but rather directly from tool arguments. This breaks the paradigm of relying on state for data usage.

In the current production implementation we are not even passing the aois as input, they are automagically used from state, the llm does not have to inject anything.

This might be too loose a coupling, so there is probably value in passing in arguments, specially now that the orchestrator sees the state.

So I would say as a middleground, lets let the orchestrator pass along IDs (that will also help to inject stuff into the agent from the outside, or from the UI context).

But then lets _validate_ that the IDs are correct, and make sure we fetch all the values / data from source, wherever we actually use the data for analysis.

src/agent/harness/tools/list_datasets.py
srmsoumya: @jterry64 this was just for protoyping, will replace this with embedding search or rlm when we move this to actual implementation.

yellowcap:
Closing because a lot of this has been implemented and merged in #669 . @srmsoumya lets take from here what is still useful but in a new PR.

Issue #163, opened 2025-05-06
Refactor: use same Dockerfile for db and api


batpad:
Currently, we have a separate Dockerfile definition for the `db/` folder. I would much rather have a single Dockerfile and two separate "commands" to run the db container.

A big reason is so that we only need to copy a single commit hash over to the `deploy` repository, and having a single image tag will eliminate chances of these two images going out of sync in the deploy.

Also, in general, seems easier to manage and cleaner. We will need to install `alembic` and perhaps `postgresql-client` in the container, but I don't think that's a big deal.

@leothomas really sorry I did not catch this in PR review. Let me know if you have any strong objections to this change. Am happy to take it on.

PR #656, opened 2026-05-07
use a single Dockerfile for api and db migrations, refs #163

batpad:
Currently, we build 2 separate Dockerfiles for the API image and for the db image, to run migrations.

This adds unneeded complexity and maintanence burden:

 - Two separate images pushed in CI / Github Actions
 - The big one, needing to remember to change the `db` tag in the deploy values when making changes that include database migrations

Instead, we can build a single image and during the migration job, just run a different `command` to execute migrations.

We've had #163 open for a while, and this is beginning to hurt us more / increase the possibility of making mistakes while updating image tags.

Once we merge this, this will need an analagous PR in the `project-zeno-deploy` directory to make changes to the `values` and the db migration job to go along-with this change. I'll open a separate PR there and reference this before we merge either.

cc @sunu @yellowcap


Dockerfile
```diff
@@ -25,8 +25,11 @@ ENV PATH="/root/.local/bin/:$PATH"
...
 ADD ./src /app/src
 ADD ./frontend /app/frontend
+ADD ./db /app/db
 
 WORKDIR /app
 
 # Install only the main dependencies - no dev deps
 RUN uv sync --frozen --no-dev
+
+RUN chmod +x /app/db/migrate.sh
```
yellowcap: Don't these permissions translate into the docker image? If yes, let's do this once on the repo and remove this line.

yellowcap (commented):
Why is this in draft? Do we have any substantial doubts about this? Happy to help moving this forward if I can be of use.

sunu (approved):
haven't tried locally but this looks good to me overall!

one thing that jumps out a bit is that we're using psycopg2 database url in the docker-compose files and asyncpg url in the DATABASE_URL in the readme. We should make sure `alembic` can work with `asyncpg` database url (it used to not work with asyncpg urls iirc)

batpad:
@sunu ya thanks for double-checking that. These lines in `env.py` should handle this: https://github.com/wri/project-zeno/blob/main/db/alembic/env.py#L33 and I did some testing locally setting an `asyncpg` type URL and it _seemed_ to read correctly.

However, I did run into another issue where `alembic` says some `mergemigrations` is required. Have we merged latest `main` to staging or prod yet? If we have and migrations were fine, then this is possibly some local issue - but if we haven't, I suspect we should fix the `mergemigrations` issue before we deploy, and I'd like to have done a deploy with latest `main` using the current setup first before merging this in, so we don't land up debugging multiple moving parts.

PR #657, opened 2026-05-11
PZB-783 Caution user that TCL driver data is only avail for full time period

danscales:
There was already a ToolMessage warning created in pull_data.py when the available date range for the dataset is different from the requested date range. However, the agent was not usually including the warning from the tool message in the final output. I gave the tool message more explicit instructions that ensures that the agent does display the warning. It handles the case in PZB-783, and should give warnings more reliably in other cases where range_clamped (returned by revise_date_range()) is true.

### DIFF
--- a/src/agent/tools/pull_data.py
+++ b/src/agent/tools/pull_data.py
@@ -145,14 +145,20 @@ async def pull_data(
     raw_data["dataset_name"] = dataset["dataset_name"]
     raw_data["source_url"] = result.analytics_api_url
 
-    if range_clamped:
-        tool_messages.append(
-            f"Date range was adjusted to the dataset's available range: {effective_start} to {effective_end} "
-            f"(requested: {start_date} to {end_date})."
-        )
+    if tool_messages:
+        content = "|".join(tool_messages)
+        if range_clamped:
+            # Ensure that the agent mentions this warning in its final output.
+            content = (
+                f"TOOL_RESULT: {content}\n\n"
+                f"IMPORTANT INSTRUCTION FOR LLM: The user's specified date range ({start_date} to {end_date}) was adjusted to the dataset's available range: {effective_start} to {effective_end}."
+                "Please include a warning in the final response to the user describing the change to the date range"
+            )
+    else:
+        content = "No data pulled"
 
     tool_message = ToolMessage(
-        content="|".join(tool_messages) if tool_messages else "No data pulled",
+        content=content,
         tool_call_id=tool_call_id,
     )
 
--- a/src/agent/tools/pull_data.py
+++ b/src/agent/tools/pull_data.py
@@ -145,20 +145,14 @@ async def pull_data(
     raw_data["dataset_name"] = dataset["dataset_name"]
     raw_data["source_url"] = result.analytics_api_url
 
-    if tool_messages:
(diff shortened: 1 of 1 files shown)


src/agent/tools/pull_data.py
```diff
@@ -145,14 +145,20 @@ async def pull_data(
...
-            f"(requested: {start_date} to {end_date})."
-        )
+    if tool_messages:
+        content = "|".join(tool_messages)
+        if range_clamped:
+            # Ensure that the agent mentions this warning in its final output.
+            content = (
+                f"TOOL_RESULT: {content}\n\n"
+                f"IMPORTANT INSTRUCTION FOR LLM: The user's specified date range ({start_date} to {end_date}) was adjusted to the dataset's available range: {effective_start} to {effective_end}."
+                "Please include a warning in the final response to the user describing the change to the date range"
```
yellowcap: No advantage to ask nicely 😊 

```suggestion
                "Include a warning in the final response to the user describing the change to the date range"
```

yellowcap (changes requested):
@danscales have you tried a more minimal change? Something along the lines of

```python
if range_clamped
        tool_messages.append(
            f"Date range was adjusted to the dataset's available range: {effective_start} to {effective_end} "
            f"(requested: {start_date} to {end_date}). Warn the user about this adjustment."
        )
```

In any case I would try to avoid adding the `important instruction` kind of language, we should find other ways to make sure this is surfaced. One alternative could be adding this to the tool docstring as an instructions about how to use the tool.

danscales:
> @danscales have you tried a more minimal change? Something along the lines of
> 
> ```python
> if range_clamped
>         tool_messages.append(
>             f"Date range was adjusted to the dataset's available range: {effective_start} to {effective_end} "
>             f"(requested: {start_date} to {end_date}). Warn the user about this adjustment."
>         )
> ```
> 
> In any case I would try to avoid adding the `important instruction` kind of language, we should find other ways to make sure this is surfaced. One alternative could be adding this to the tool docstring as an instructions about how to use the tool.

Thanks @yellowcap, just the "Warn the user about this adjustment." that you suggest seems to work for the two times I tried it. I'll go with that and verify that it consistently gives the warning.  I'll also try to write an eval.

danscales:
> > @danscales have you tried a more minimal change? Something along the lines of
> > ```python
> > if range_clamped
> >         tool_messages.append(
> >             f"Date range was adjusted to the dataset's available range: {effective_start} to {effective_end} "
> >             f"(requested: {start_date} to {end_date}). Warn the user about this adjustment."
> >         )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In any case I would try to avoid adding the `important instruction` kind of language, we should find other ways to make sure this is surfaced. One alternative could be adding this to the tool docstring as an instructions about how to use the tool.
> 
> Thanks @yellowcap, just the "Warn the user about this adjustment." that you suggest seems to work for the two times I tried it. I'll go with that and verify that it consistently gives the warning. I'll also try to write an eval.

@yellowcap  I made the changes to the simpler warning that you requested, and also added an eval on the "Date selection" sheet.

yellowcap (approved):
Looks great thanks @danscales !

PR #658, opened 2026-05-11
PZB-877 - Simplify insights generation and fix final AI response

yellowcap:
The tool message from the `generate_insights` tool does not contain enough information for a good AI response. The result was that the final response was sometimes hallucinated, even though the chart itself was correct.

This PR fixes that by adding more information to the tool message.

It also simplifies the insights generation by asking the coding agent to output the insights json data as an artifact. This avoids one additional LLM call and reduces duplication of instructions (between CodeAct and InsightsJson generation).


yellowcap:
Ran evals set locally and looks good

```
SIMPLE E2E TEST SUMMARY
Tests Run (after filters): 65

Agent Answer:              49 /  57 (0.86)

AOI ID Match:              39 /  46 (0.85)
Dataset ID Match:          57 /  58 (0.98)
Dataset Parameter Match:    0 /   0
Context Layer Match:        3 /   3 (1.00)
Data Pull Exists:          57 /  57 (1.00)
Date Match:                34 /  34 (1.00)
Charts Answer:             37 /  51 (0.73)
Expected Text Match:        3 /   4 (0.75)
Clarification Requested:    0 /   0
```

src/agent/tools/code_executors/gemini_executor.py
srmsoumya, line 240: Add `insight=None` or else this will mask the original model error.

src/agent/tools/generate_insights.py
```diff
@@ -431,121 +401,31 @@ async def generate_insights(
...
-    )
+    tool_message = f"Generated {len(result.insight.charts)} chart(s)\n"
+    tool_message += f"Key Finding: {result.insight.primary_insight}\n\n"
 
-    for idx, chart in enumerate(chart_insight_response.charts, 1):
+    for idx, chart in enumerate(result.insight.charts, 1):
         tool_message += f"Chart {idx}: {chart.title}\n"
 
+    tool_message += f"\nAnalysis output:\n{text_output}"
+    tool_message += f"\nChart data CSV:\n{chart_data_df.to_csv(index=False)}"
```
srmsoumya: This might bloat the context, we already have `chart_data` in state - any specific reason why you are passing all the rows as tool message?

srmsoumya (changes requested):

src/agent/tools/generate_insights.py
yellowcap: ok agree, let's put this into the system prompt for now.

yellowcap: system prompt alone did not work, the agent still answers with halucinated data. But limited dumping the data to dataframes with less than 50 rows. This will work for concrete asks like top-x where the answer can still be given directly. For more high level statistical questions the context is not  polluted with big dataframes.

srmsoumya (approved):

PR #659, opened 2026-05-11
Return dates used for dataset selection

yellowcap:
This adds the effective date range to the pick dataset tool. This is in preparation for showing this information on the legend in  [PZB-720](https://gfw.atlassian.net/browse/PZB-720).

- Adds start and end date to `DatasetOption`
- Updates date range in tile url construction (specially TCL 2025)
- Adds tests to check if data is there

cc @kamicut 

[PZB-720]: https://gfw.atlassian.net/browse/PZB-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/agent/tools/pick_dataset/schema.py
+++ b/src/agent/tools/pick_dataset/schema.py
@@ -30,6 +30,14 @@ class DatasetOption(BaseModel):
     parameters: Optional[list[DatasetParameter]] = Field(
         None, description="Dataset specific parameters."
     )
+    start_date: Optional[str] = Field(
+        None,
+        description="Start date of the dataset that best matches the user query.",
+    )
+    end_date: Optional[str] = Field(
+        None,
+        description="End date of the dataset that best matches the user query.",
+    )
     reason: str = Field(
         description="Short reason why the dataset is the best match."
     )
--- a/src/agent/tools/pick_dataset/schema.py
+++ b/src/agent/tools/pick_dataset/schema.py
@@ -32,11 +32,11 @@ class DatasetOption(BaseModel):
     )
     start_date: Optional[str] = Field(
         None,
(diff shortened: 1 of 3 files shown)


tests/tools/test_pick_dataset.py
```diff
@@ -420,9 +421,14 @@ async def test_queries_return_expected_dataset(
 
     command = await pick_dataset.ainvoke(tool_call)
 
-    dataset_id = command.update.get("dataset", {}).get("dataset_id")
+    dataset = command.update.get("dataset", {})
+    dataset_id = dataset.get("dataset_id")
     assert lookup[dataset_id] == expected_dataset
 
+    assert dataset.get("start_date") and dataset.get("end_date")
+    date.fromisoformat(dataset["start_date"])
```
danscales: Maybe add a comment here that date.fromisoformat throws an exception if the start_date/end_date are not valid date strings. I was confused at first, since you are not using the results, so they seemed like no-ops.

src/agent/tools/pick_dataset/schema.py
```diff
@@ -30,6 +30,14 @@ class DatasetOption(BaseModel):
     parameters: Optional[list[DatasetParameter]] = Field(
         None, description="Dataset specific parameters."
     )
+    start_date: Optional[str] = Field(
+        None,
+        description="Start date of the dataset that best matches the user query.",
```
danscales: These are the effective dates, not the original start/end dates of the best matching dataset.  So, maybe add to these two descriptions: "as intersected with any user-specified start/end dates".

danscales (approved):

PR #660, opened 2026-05-11
PZB-867: Correctly query GADM subregions

jterry64:
While looking through golden set failures, we discovered a lot of failures caused by a bug where GADM subregions were queried incorrectly due to some misunderstanding in how to split the IDs. This caused the subregion query to accidentally go one level too high, causing many extra AOIs to be returned. For example, if you ask for all counties in Texas, the query will end up returning all counties in the whole US. This simple change should fix the issue.

### DIFF
--- a/src/agent/tools/pick_aoi/tool.py
+++ b/src/agent/tools/pick_aoi/tool.py
@@ -371,6 +371,8 @@ async def query_subregion_database(
                 subregion_filter = src_id.split("_")[0]
             else:
                 subregion_filter = src_id
+
+            # filter for the next admin level within this admin ID
             gadm_filter = f" AND t.gadm_id LIKE '{subregion_filter}.%'"
         else:
             gadm_filter = f" AND t.gadm_id ~ '{GADM_STANDARD_ID_RE}'"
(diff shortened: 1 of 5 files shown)


solomon-negusse (approved):

src/agent/tools/pick_aoi/tool.py
```diff
@@ -366,8 +366,9 @@ async def query_subregion_database(
...
     if table_name == GADM_TABLE:
         if source == "gadm":
-            if "." in src_id:
-                subregion_filter = ".".join(src_id.split(".")[:-1])
+            # remove _1/_2 GADM suffix
+            if "_" in src_id:
+                subregion_filter = src_id.split("_")[0]
             else:
                 subregion_filter = src_id
             gadm_filter = f" AND t.gadm_id LIKE '{subregion_filter}.%'"
```
danscales, line 376: This is from before, but I'm not sure why we have the period there in the pattern in line 374. That is not a wildcard (underscore is the wildcard for a single character), so that is forcing that the matching GADM id has a period after the subregion_filter, hence requiring that it includes another level of ID.  Is that what we want? If so, maybe you can add a comment here about that, since that is subtle.

danscales (approved):

src/agent/tools/pick_aoi/tool.py
jterry64: @danscales yeah I think in this part of the code, it's specifically trying to search for the next level if that's what you mean

PR #661, opened 2026-05-11
PZB-830: Prompting fixes for primary forest

jterry64:
PZB-830 cover a few minor but related issues:

- There's many rules in the TCL YAML about "deforestation" that don't apply the same when primary forest is a context layer, since that filters to just natural forests and removes plantations. 
- The agent inconsistently mentions the context layer in the title for the chart.
- The agent also has a tendency to use verbose names like "primary forest tree cover loss", when researchers expressed it'd be more readable as "primary forest loss".


yellowcap (approved):
Looks good to me - do we have some evals / unit tests that check this behaviour? If not I'd add one (maybe in the insights text, do this and regex to check the title).

jterry64:
> Looks good to me - do we have some evals / unit tests that check this behaviour? If not I'd add one (maybe in the insights text, do this and regex to check the title).

Thanks for reviewing! I added a new check to the gold set eval, "Show me deforestation in Brazil", that uses an LLM judge to check the text/chart title for: 'References "primary forest loss" in text and chart, not "tree cover loss" or "primary forest tree cover loss". Cautions users about loss vs. deforestation but not that the data includes plantations.'

PR #664, opened 2026-05-12
Deal with case where user's start_date/end_date not included.

danscales:
Deal with simple error that results in an exception during an eval case that I noticed (but the eval recovers and continues).

Seems like this error and exception can happen whenever the user does not specify start date or end date.


jterry64 (approved):
Good catch!

PR #665, opened 2026-05-13
Remove pull data from state but keep backwards compatability

yellowcap:
This avoids writing the `raw_data` from data pulls into state.

Now that we are tracking the pull data requests through urls and store metadata in DB, we can avoid writing the actual data into state. This PR does just that, while keeping backwards compatability.

### DIFF
--- a/src/agent/tools/generate_insights.py
+++ b/src/agent/tools/generate_insights.py
@@ -268,7 +268,8 @@ Now prepare the data for visualization in Recharts.js:
       2. **Field names**: Use clear, lowercase names like 'date', 'value', 'category', 'year', 'count'
       3. **Numeric values**: Always numbers, never strings (e.g., 100 not "100")
       4. **Date ordering**: Chronological order for time series, not alphabetical
-      5. **Data format by chart type**:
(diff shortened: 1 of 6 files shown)


src/agent/tools/generate_insights.py
```diff
@@ -41,10 +53,11 @@ async def prepare_dataframes(
     """
     Prepare DataFrames from raw data for code executor.
 
-    Fetches all source URLs concurrently, then builds DataFrames in order.
+    Fetches ID-backed data by source URL, while keeping older inline-data
+    thread state working.
     """
     raw_results = await asyncio.gather(
-        *[fetch_statistics_from_url(s["source_url"]) for s in statistics]
+        *[_load_statistics_data(s) for s in statistics]
```
srmsoumya, line 64: `pull_data` used to include a `name` column derived from aoi_id. With the new source_url refetch path, `fetch_statistics_from_url()` returns only the raw API result and does not re-add name - can you check this is not affecting the codeact loop

yellowcap: very good catch @srmsoumya . I added a tracking of ID to name in the statistics object and use this to re-inject names into the insight. That is better. https://github.com/wri/project-zeno/pull/665/changes/a3fbe34aa0bb677eac2ac5c9626051daa8006031

srmsoumya (approved):
Lgtm 🚀

PR #666, opened 2026-05-13
PZB-897: Add API endpoints for superusers to manage users / user status

batpad:
Adds a new `superuser` type, in addition to the existing `regular`, `pro` and `admin` types. Superusers have the ability to list existing users and change the user_type of existing users.

Adds a CLI command to promote a user to superuser (needed to make the first superuser).

Adds two endpoints:

 - GET /api/admin/users: Lists existing users, can be filtered with a `?email=` query parameter, available only to superusers, does pagination with `offset` / `limit`, defaults to 50 per page
 - PATCH /api/admin/users/<user-id>/user_type - with a payload like `{ "user_type": "admin" }` to set a user user type for user with id (will return a 403 if user is not superuser). **Question**: Am a bit unsure if this should be `/api/admin/users/<user-id>/user_type` or just ` /api/admin/users/<user-id>` with the only action allowed being this PATCH .. I think the `/user_type/` feels easier to reason about, but feels less "RESTful" 

Adds tests for CLI command and the API endpoints.

For all places on the API where we allow "admins" to do an action, also ensure we allow "superuser" users. This is non-ideal and starts getting a bit messy - we should probably discuss the `UserType` handling and if there's anything we want to change here - this did grow a bit ad-hoc with needs - I _think_ its fine, and better than hard-coding prompt amounts to users, but maybe some things can be consolidated better - definitely checking for ADMIN or SUPERUSER felt ugly.

Would be nice to go ahead with this approach though and save any larger refactor of user handling for later - I'd like to get this basic admin interface into the frontend and then we can easily iterate on both sides.

My process: Spent some time creating the entire plan with Claude, then split it up into smaller steps with tests coming before implementation. Reviewed each step, iterated a bit, and committed each logical step. I found enforcing the smaller commits and granular steps of manual review to be quite useful in catching little things.

cc @yellowcap @gtempus

### DIFF
--- a/src/api/cli.py
+++ b/src/api/cli.py
@@ -262,6 +262,26 @@ async def make_user_pro(session: AsyncSession, email: str) -> UserOrm:
     return user
 
 
+async def make_user_superuser(session: AsyncSession, email: str) -> UserOrm:
+    """Make a user superuser by setting their user_type to superuser"""
+
+    email_lower = email.lower()
+    result = await session.execute(
+        select(UserOrm).where(func.lower(UserOrm.email) == email_lower)
+    )
+    user = result.scalar_one_or_none()
+    if not user:
+        raise ValueError(f"User with email {email} not found")
+
+    user.user_type = UserType.SUPERUSER.value
+    user.updated_at = datetime.now()
+
+    await session.commit()
+    await session.refresh(user)
+
+    return user
+
+
 # CLI Commands
 @click.group()
 def cli():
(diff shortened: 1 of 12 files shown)


gtempus:
Love your process and thank you for sharing, @batpad. 

I agree, this isn't great (from a design/refactor perspective), but it's where we'll start.

src/api/routers/admin.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    stmt = select(UserOrm)
+    if email is not None:
+        stmt = stmt.where(func.lower(UserOrm.email).contains(email.lower()))
+    stmt = stmt.order_by(UserOrm.created_at.desc(), UserOrm.id.asc())
+    stmt = stmt.offset(offset).limit(limit)
+
+    result = await session.execute(stmt)
+    users = result.scalars().all()
+    return [_orm_to_user_model(u) for u in users]
```
yellowcap, line 41: I came across this also on some of the stats / insight persistence and found it not ideal that there are two definitions and we need this wrapper. Shall we put the conversion as a method on the ORM class? not sure if good design pattern but an idea.

src/api/routers/admin.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@router.patch("/api/admin/users/{user_id}/user-type", response_model=UserModel)
```
yellowcap, line 44: I know you commented, so not need to change, but suggestion to put this into url not body


```suggestion
@router.patch("/api/admin/users/{user_id}/user-type/{user_type}")
async def set_user_type(user_id: UUID, user_type: UserType, ...):
```

src/api/cli.py
```diff
@@ -538,6 +558,33 @@ async def _make_pro():
     asyncio.run(_make_pro())
 
 
+@cli.command("make-user-superuser")
+@click.option(
+    "--email", required=True, help="Email of the user to make superuser"
+)
+def make_user_superuser_command(email: str):
```
yellowcap, line 565: Maybe once we have the admin we can get rid of the cli. One day! 🤞🏽

yellowcap (approved):
@batpad looks pretty good to me, happy to move forward. Made 2-3 comments that are non critical. Up to you to adapt or not.

src/api/routers/admin.py
batpad: Hmmm... that's an interesting idea, but for some reason breaks my mental model of "how I would usually do this". Am going to leave this as is in the PR, but happy to do a quick change of the URL structure if there's strong opinions after we implement the frontend.

PR #667, opened 2026-05-13
Gold Set 1-031: Include all classes in SBTN yaml

jterry64:
Fixes dataset selection for gold set 0-131: Which state has the most mangroves in Senegal: Fatick or Ziguinchor?

It was consistently choosing "tree cover" before despite having no mangroves-specific data. Now the eval successfully filters to the mangroves class of natural lands dataset.


danscales (approved):
Looks good. I assume that the class numbers (that were in parentheses) are not used anywhere.

If we find the LLM still has problems, we could make an even more obvious list (i.e. with bullets) to make it clear exactly that each item in the comma-separated list are distinct classes with the specified name.

PR #668, opened 2026-05-14
PZB-894: make relative date range instruction for last x period queries more specific

solomon-negusse:
For queries for data for last x years, the LLM was ignoring the relative date range instruction to use current date for end date and was instead pinning it to 2024.  This PR makes the instruction more specific. This was an extremely difficult to reproduce locally and the only way I was able to was by using the deployed gemini api key locally (will still need to validate that that was what did it)

Screenshot 2026-05-14 at 12 03 38 PM


yellowcap (approved):
Very clear formulation - thanks @solomon-negusse

PR #670, opened 2026-05-18
PZB-897: Add endpoint for superusers to download CSV of users table

batpad:
Continuing on PZB-897 - Adds a new endpoint: `/api/admin/users/export?format=csv`

Site admins, about once a month, need a CSV dump of users for analytics. Currently, this is a manual process requiring `kubectl` access and running a SQL script to get the CSV. With the goal of making these actions self-serve for users with "superuser" privileges, this adds an endpoint to download a CSV of all users.

This mostly sticks to the existing schema we are using when exporting from the database, with minor differences:

 - 21 columns instead of 22 - removes `machine_description` column as it's not needed and would add a bit of complexity.
 - Booleans represented as `true` / `false` instead of `t` / `f` to stick to CSV conventions
 - Datetimes represented as proper ISO strings like `2026-05-10` and not date(str) or so

Since this is a superuser action, we imagine it will not be run often, so to not introduce additional complexity what this does NOT do:

 - Any caching for CSV responses
 - Streaming of CSV response - this would add more complexity to the frontend to reconstruct and build the CSV for download and as this is an occasional superuser action, it's probably fine to build the CSV in memory on the backend and return a single file download to the client.
 - Currently, only offers CSV as the `format` type, but written such that it is trivial to add a JSON download for example if we want.

This uses the existing UserOrm model so should be relatively straightforward.

cc @yellowcap @gtempus

### DIFF
--- a/src/api/routers/admin.py
+++ b/src/api/routers/admin.py
@@ -1,9 +1,11 @@
 """Superuser-only admin endpoints (manage other users)."""
 
-from datetime import datetime
-from typing import Optional
+import csv
+import io
+from datetime import datetime, timezone
+from typing import Any, Optional
 
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Depends, HTTPException, Query, Response
 from sqlalchemy import func, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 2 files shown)


yellowcap (approved):
Yay for more admin actions!

PR #673, opened 2026-05-19
Lint code with mypy and enforce moving forward

yellowcap:
This adds a `pre-commit` hook and a gh action for `mypy`. 

Completes the work in https://github.com/wri/project-zeno/pull/650

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
       - id: ruff-format
         types_or: [ python, pyi ]
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.16.0
+    rev: v2.1.0
     hooks:
       - id: mypy
         args: [--ignore-missing-imports, --explicit-package-bases]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,14 @@ repos:
       # Run the formatter.
       - id: ruff-format
         types_or: [ python, pyi ]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.16.0
+    hooks:
+      - id: mypy
+        args: [--ignore-missing-imports, --explicit-package-bases]
+        pass_filenames: false
+        entry: mypy src/
+        additional_dependencies:
+          - types-PyYAML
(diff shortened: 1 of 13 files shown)


src/agent/tools/generate_insights.py
```diff
@@ -433,6 +433,20 @@ async def generate_insights(
...
+                "messages": [
+                    ToolMessage(
+                        content=f"Failed to generate chart insight. Feedback:{text_output}",
+                        tool_call_id=tool_call_id,
+                        status="error",
+                    )
+                ]
+            }
+        )
+
```
danscales, line 449: I assume this is just another unrelated fix that you're adding into this PR, which is fine, but just noting that it doesn't seem to be related to fixing types, etc.

danscales (approved):
The pull_data() call in pull_data.py:77 still has its arguments in a different order than all the other ones (like in base.py).  Although I'm sure we're always specifying all args with keywords, it would be good to fix it to have the same consistent signuature, to avoid confusion, etc.

Otherwise, looks great!

.pre-commit-config.yaml
```diff
@@ -20,3 +20,14 @@ repos:
       # Run the formatter.
       - id: ruff-format
         types_or: [ python, pyi ]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.16.0
```
photonbit: Just from random curiosity, why mypy 1.16.0 and not 2.1.0?

[v1.16.0](https://github.com/pre-commit/mirrors-mypy/releases/tag/v1.16.0)
 on May 30, 2025

src/agent/tools/generate_insights.py
yellowcap: Its related in the sense that the linter knows that if this can be none and might lead to problems on lines 450 and 451. So its a bug fix thanks to mypy!

.pre-commit-config.yaml
yellowcap: good catch @photonbit ! I guess the 🤖 knowledge cutoff date? fixed in https://github.com/wri/project-zeno/pull/673/commits/ab3546e357ed075f75df31d7e266bfb195c3f292

PR #675, opened 2026-05-19
Switch to automatic CalVer versioning

yellowcap:
Replace the PR-blocking version bump check with a CI workflow that automatically sets the version on every merge to main using CalVer (YYYY.M.D). Same-day merges get an incrementing counter suffix. Adds versioning docs to the README.

Thanks @gtempus for the input on this the CalVer is much easier in this case.


gtempus (approved):
I like this much better, too, @yellowcap! Nice! 💪

PR #676, opened 2026-05-20
Optimizations to thumbnail function:

batpad:
I started digging into implementing a cache for the thumbnail function - in digging with 🤖, found a few spots we should likely optimize before implementing a cache. The current `_fit_overlay` function which can be expensive, was being called in the main event loop, so would be blocking for all operations. This converts it to being called `async` to be non-blocking.

Also makes a couple of other little optimizations, the one that seems to give a big win is just not running the antimeridian fix step at all if we can determine that the geometry does not cross the antimeridian.

 - Run _fit_overlay async to not block main thread
 - Only perform fix_antimeridian if geometry actually crosses antimeridian
 - Optimize number of runs to reduce vertices

All existing tests continue to pass, this should be purely an optimization fix.

@yellowcap I think I'd like to validate these changes before moving on to implementing a cache layer. At the very least, I think we should ensure that this operation is non-blocking before we merge to production.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "0.2.0"
+version = "0.2.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/api/routers/thumbnails.py
+++ b/src/api/routers/thumbnails.py
@@ -1,5 +1,6 @@
 """AOI thumbnail generation via Mapbox Static Images API."""
 
+import asyncio
 import json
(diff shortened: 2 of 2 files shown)


yellowcap (approved):
Looks good async all the way! Thanks @batpad

yellowcap:
Regarding blocking: we can merge to production because the frontend is not hitting this yet. So its not a blocker for deploying, but a blocker for using it in the AOI cards on the frontend.

batpad:
Going to go ahead and merge this in the spirit of having smaller, focussed PRs.

And I think caching might require a bit more discussion on whether to cache final thumbnails or intermediate simplified geometry or none.

PR #678, opened 2026-05-20
Pick dataset decision tree

jterry64:
Inspired by the discussions on make the tools less simpler primitives, it occurred to me that we might be able to remove the RAG check and internal LLM calls entirely from pick_dataset while making the domain logic more clear and cohesive. 

I created an alternative tool to `pick_dataset` called `pick_land_change_dataset` that identifies the correct dataset and context layer by having the agent just extract the right info from the query and passing it as params. These includes things like land cover type, event type and measurement. Once we have the information, it uses a completely deterministic decision tree to select the appropriate dataset and context layer.

I ran the golden set locally using this change, and get a 100% pass rate for dataset ID. I also spot checked against a few harder questions from the dataset ID eval set, but would like to formalize the experiment to compare against the current tool.

I think there's quite a few advantages to this approach:
- Significantly simplifies the tools and makes unit testing much easier
- Latency goes from seconds to milliseconds since it's just a quick if/else tree
- This puts all the business rules in one place instead of scattered across YAMLs, and makes it much clearer how we're choosing datasets
- It allows the LLM to just focus just on user intent instead of having to think many complex dimensions like available datasets, context layers, dates and extents. This will lead to fewer inconsistent responses due to overlapping semantics, weird time intervals and overloaded instructions.

A couple potential cons:
- Our decision tree might get quite large to handle all possible cases. From this initial experiment, we don't need to exhaustively check all combinations, but just the ones that clearly signal certain datasets. This 50 line decision tree could handle the whole golden set; land cover/event params alone can distinguish pretty much all our datasets. 
- We might introduce datasets or parameters that don't fit cleanly into our parameters.

We might consider adding a deployment flag to which tool is used in production so we can easily swap them and see how they perform using an online LLM judge.


01painadam:
I think this would be a good one for discussion - its a design philosophy question, right?

I like the idea in principle _right now_ with ~10 semantically very distinct datasets... but this worries me as the dataset catalogue expands. This will get brittle  and/or unmanageable fast.

The latency gain is nice but I'm unconvinced a user would trade a good _reasoned_ dataset selection for an instant one with no ability to say _why_. Thats the core value of using an LLM call at the pick_dataset step. Latency is part of the value of the LLM, and we're talking seconds, not minutes. I'd also have questions about how this would perform with semantically vague user prompts or prompts that are outright wrong due users not knowing how/what to ask e.g. how would it handle _"is primary forest loss in Russia dying"?_ 

A few things I'd like to surface for the team to discuss:
- Is there a version of this where the deterministic step does the easy routing (broad topic: land change vs carbon vs alerts vs supply chain) and then a narrow LLM call inside the topic disambiguates among a smaller pool of candidate datasets with the YAML context to hand? That probs keeps most of the latency win, keeps the reasoned explanation, and gives us a sensible fallback when the params don't fit.
- Are we comfortable with the policy that adding a new dataset is now a code change rather than a metadata/YAML change?

yellowcap:
@jterry64 thank you for putting this together and showing alternatives to the current setup. I agree that this needs revisiting.

So I defintively agree with the intention behind this PR. There might be better ways than RAG to surface datasets. @srmsoumya has some ideas as well of how we could change this.

But I agree with @01painadam that this is a bigger discussion. It involves more than just the agent tool itself.

For me, part of this discussion is about where the metadata lives and how it is stored and structured. The yamls kind of need a review / refactor. Through that it is also related to evals, where we discussed that the yaml is representing a set of instructions. These are currently distributed across prompt, insight, and description, cautions etc. 

My the first step is to review the yaml, i.e. how can we make that more structured.

jterry64:
@01painadam @yellowcap thanks for the quick feedback, I'm glad it started a discussion at least! I totally see where your concerns are coming from, and I too was suspicious it'd be more brittle when I first thought of it, so wanted to try it out to see. But I was surprised it actually works as well or stronger than the current approach, likely because it provides a more coherent semantic model for the LLM to choose from. It has no problem with questions like your example AJ, because it will just pull out land_cover=primary_forest and event=loss to choose the appropriate dataset; the "dying" part isn't relevant until it generates the insight. 

Likewise, it can handle queries the current one struggles with right now like "Compare grasslands and forest extent in Mozambique" - the current implementation flip flops between natural grasslands, tree cover and land cover datasets, but this approach sees you want to compare across land cover classes so will always choose the correct answer, land cover. It can even handle really ambiguous questions like "Can you tell me why bird populations might be declining in Louisiana?" - I observed the agent itself got no hits on its first query and the rephrased the query to get a good answer based on the error message. Because the latency is so low, the agent can do this freely without slowing down.

If this [this](https://onewri.sharepoint.com/:x:/s/LandandCarbonWatch/IQAfTodz-Vk9SYJW0kYKEXmNAWODllzQSBkdw--6pQwc0zs?e=ThZ3JM) is a reasonable picture of datasets we're adding, there's a lot of semantic overlap. E.g. for fires, we have 3 fire datasets now, it's quite difficult for the LLM to pick correctly just through prompting. So I anticipate we're going to struggle more with similar/overlapping datasets much more than very different datasets. I think this is way more manageable than the YAMLs, which have been blowing up in size with prompt rules to get the behavior we want. The last few things I added, it was very difficult to debug why the agent wasn't following the rules, and seemed most often that it was ignoring rules because it was overloaded.

But I am definitely open to hybrid approach with having some LLM decisions, and at the very least re-assess the YAMLs. My proposal: we should discuss how this relates to our design philosophy, but let's also use evals for empirical evidence. I mentioned trying to make "challenge sets" to test our experimental ideas, let's see what that means for comparing these two tools. Adding TCL by fires next sprint might also be a good challenge for which approach handle multiple fire datasets better.

yellowcap:
Also, part of the idea of making a stronger separation of concern is that we can benchmark each tool much more independently. I do like that we are experimenting with better newer options. So let's dive deeper in to this in the new sprint? Maybe starting with a sync including folks on this thread, @srmsoumya, and anyone else interested.

PR #681, opened 2026-05-21
PZB-910: Fix chart axis configuration for multi-series charts

yellowcap:
## Summary

- Added a `model_validator` to `ChartInsight` that rejects any non-pie/table chart where both `y_axis` and `series_fields` are empty, catching degenerate LLM output at parse time before it reaches state or the DB.
- Tightened executor prompt language to make explicit that `series_fields` must be populated for multi-series charts and `y_axis` must be set for single-series charts.

## Root cause

When the analyst generates two charts from the same dataset (e.g. tree cover loss area and emissions), both charts previously shared the identical `chart_data` array containing all metric columns. `formatChartData` on the frontend — when `yAxis` is empty — treats every non-x column as a series, so each chart would display all four metric columns instead of its intended two.

The paired frontend fix (project-zeno-next) relaxes the axis guard so charts with `yAxis=""` and multi-series data render correctly rather than showing "Chart is missing axis configuration."

### DIFF
--- a/src/agent/subagents/analyst/tool.py
+++ b/src/agent/subagents/analyst/tool.py
@@ -106,6 +106,24 @@ def _extract_statistics_ids(statistics: list[dict]) -> list[str]:
     ]
 
 
+def _filter_chart_data(
+    chart_data: list[dict], x_axis: str, y_axis: str, series_fields: list[str]
+) -> list[dict]:
+    """Return only the columns relevant to a single chart.
+
+    Each chart in a multi-chart insight shares the same underlying CSV but
+    plots different columns. Keeping unrelated columns causes formatChartData
+    (frontend) to auto-discover them as extra series, producing wrong charts.
+    """
+    keep = {x_axis}
+    if y_axis:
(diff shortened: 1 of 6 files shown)


src/agent/subagents/analyst/tool.py
```diff
@@ -106,6 +106,24 @@ def _extract_statistics_ids(statistics: list[dict]) -> list[str]:
     ]
 
 
+def _filter_chart_data(
```
solomon-negusse: since we are passing `seriesFields` to the FE (which it's ignoring right now -  I had a local patch for this), wouldn't it be simpler to just have the existing chart formatter there utilize that to decide what to plot instead of doing the work here?

yellowcap: Yes good thinking @solomon-negusse  I agree. I'll remove that filter.

solomon-negusse (approved):
Looks good, @yellowcap. Nice cleanup of the tests too.

PR #683, opened 2026-05-22
Move version bump to pre-commit

yellowcap:
This adds a pre-commit hook that bumps version. Works similar to ruff formatting.

The current post PR hook on github actions is not working because main branch is protected. So the direct commit fails.

Local flow (same as Ruff): first commit attempt bumps and fails → git add pyproject.toml → commit again → passes.

### Hook

- Bumps pyproject.toml when local version still matches main
- Exits 1 after bumping (same as other pre-commit auto-fix hooks)
- Exits 0 when no bump is needed (local != main, or already at target)

### CI

- Runs uvx pre-commit run bump-calver --all-files
- Passes (0) when the PR already has a bumped version
- Fails (1) if it still has to bump — meaning pre-commit wasn’t run locally before push

### DIFF
--- a/.github/workflows/calver-bump.yml
+++ b/.github/workflows/calver-bump.yml
@@ -1,52 +1,24 @@
-name: CalVer Bump
+name: CalVer Check
 
 on:
-  push:
+  pull_request:
     branches:
       - main
 
 jobs:
-  bump:
+  check:
     runs-on: ubuntu-latest
-    permissions:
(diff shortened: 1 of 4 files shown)


yellowcap:
ok merging this since we already agreed on this as the right versioning strategy and the current bumping is not working. This is not more mental overhead, its all nicely built into pre-commit.

PR #684, opened 2026-05-22
PZB-913: Suggest datasets instead of always choosing one

jterry64:
In anticipation of more conflicting datasets like TCL by fires vs. TCL driver wildfire class vs. DIST from fires, introducing what I think @01painadam was saying of returning suggestions to the user if we don't have an appropriate dataset to match their query. This mainly involves adjusting prompting to return null for dataset_id if there's no good match, but I discovered through some other out-of-date prompts like still some places where we say not to analyze dataset globally.

This PR will be a big user experience change, but largely for the better. Now the agent returns much faster and is much clearer when your query can't be answered by our data or requires further input. It provides good suggestions that help users understand what data is available. This also means our dataset YAMLs can focus on providing clear descriptions of use cases instead of too many instruction prompts on how to choose the right dataset. The only downside is it may eat into user's daily prompts more, but the same happens if we don't provide the data the user was looking for.

@01painadam I'd recommend if you have time to load this up locally and play around a bit to see if it feels alright. If we do want to limit it only to fires, we could temporarily include an instruction under "No Match" that says only use this option for fire-related queries. @yellowcap you mentioned about how we might want to try a more skills-based approach, so feel free to comment here if there was an alternative approach you were thinking.

### DIFF
--- a/src/agent/subagents/pick_dataset/tool.py
+++ b/src/agent/subagents/pick_dataset/tool.py
@@ -111,8 +111,9 @@ async def select_best_dataset(
     if a user says "show me tree cover loss in forests where canopy cover is greater than 50%", you may select the parameter canopy cover
     and value 50.
 
-    Evaluate if the best dataset is available for the date range requested by the user.
-    If not, pick the closest available date range and include a warning in the dataset pick reason.
+    If the user specifies a date range or time granularity (e.g. monthly, a specific year, daily
+    alerts), only select a dataset if it genuinely supports that. If no candidate does, return null
+    and tell the user what dates and granularity are available instead.
 
     Pick the most granular dataset/contextual layer/parameters that matches the query.
 
--- a/src/agent/subagents/pick_dataset/tool.py
+++ b/src/agent/subagents/pick_dataset/tool.py
@@ -90,7 +90,7 @@ async def select_best_dataset(
     start_date: Optional[str] = None,
     end_date: Optional[str] = None,
     aoi_selection=None,
-) -> DatasetSelectionResult:
+) -> DatasetSelectionResult | list[DatasetOption]:
     DATASET_SELECTION_PROMPT = ChatPromptTemplate.from_messages(
         [
             ("system", DATASET_SELECTOR_PROMPT),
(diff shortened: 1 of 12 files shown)


01painadam:
@jterry64 this is a good start, nice job! I think the instructions are balanced enough right now and the trigger and respense _feel_ about right.

It does still seem to confidently pick more often than I'd expect (perhaps not a bad thing):
- "how much of brazil is forest?" confidently selects Tree Cover 2000
- "how much of brazil is forest today?" confidently selects Global Land Cover 2024
- "show me the impact of fires on Brazil" --> TCL Drivers (was expecting fires vs. DIST drivers)
- "how me the total extent of natural fields" --> SBTN Nat lands (was expecting Land Cover vs. Grasslands)

All this said, no red flags - and potentially a good human eval candidate to tune this.

On the topic of eating prompts, yeah - I was wondering if we could get around this with the nudge UI (don't eat a prompt if a selection is made that way)

solomon-negusse (approved):
Looks good!

src/agent/subagents/pick_dataset/tool.py
```diff
@@ -90,14 +90,15 @@ async def select_best_dataset(
     start_date: Optional[str] = None,
     end_date: Optional[str] = None,
     aoi_selection=None,
-) -> DatasetSelectionResult:
+) -> DatasetSelectionResult | DatasetSelectionResponse:
     DATASET_SELECTION_PROMPT = ChatPromptTemplate.from_messages(
```
danscales, line 94: Could you put a comment on the select_best_dataset() function here that indicates that you return a DatasetSelectionResult object if there is a unique best dataset (and filled in with all the dataset details), but return a DatasetSelectionResponse object if there are none or multiple matching datasets.

danscales (approved):

src/agent/subagents/pick_dataset/tool.py
jterry64: Actually I don't quite like this, will fix this

jterry64: Ok made it so it always returns DatasetSelectionResponse and checks later for the selected_dataset

PR #685, opened 2026-05-25
Paginate threads

batpad:
Refs https://github.com/wri/project-zeno-next/issues/440

Backend components for threads pagination -

 - Returns a maximum of 100 threads per request
 - Uses the datetime for the `created_at` field as cursor for cursor pagination
 - Accepts `cursor` and `limit` query parameters to the `threads` endpoint to paginate by given cursor
 - Return a `X-Next-Cursor` header with the datetime to use as cursor for the next page, if there are more results

By putting the pagination information into a header, we avoid changing the shape of the response. Currently, the `threads` API returns an `Array` of threads. The other option would be to change this to an object like:

```
{
  "threads": [],
  "cursor": "2026-01-01"
  ...
}
```

I preferred to keep the shape of the response as is, mostly to avoid any unintended consequences of changing that shape - i.e. existing clients will not need to change anything - they will just only get the first 100 results instead of all results, and will need to use the cursor pagination mechanism to fetch more results. I don't have a strong opinion about this: happy to change to put the pagination information in the response object and not in the header if anyone has a strong opinion.

cc @yellowcap @kamicut

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -47,7 +47,7 @@ app.add_middleware(
     allow_credentials=True,
     allow_methods=["*"],
     allow_headers=["*"],
-    expose_headers=["Content-Disposition"],
+    expose_headers=["Content-Disposition", "X-Next-Cursor"],
 )
 
 
--- a/src/api/routers/threads.py
+++ b/src/api/routers/threads.py
@@ -4,7 +4,7 @@ import uuid
 from datetime import datetime
 from typing import Optional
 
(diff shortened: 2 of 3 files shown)


yellowcap (approved):
Looks good to me if the frontend folks are happy with the x-next-cursor header approach.

batpad:
Going to go ahead and merge this to have an easier time testing wiring up the frontend. If we have strong feelings about not using headers for pagination info, we can revisit and will be easy to change things around.

PR #686, opened 2026-05-25
Skill/fao fra

01painadam:
### Summary 
Adds a new `fao-fra` skill that answers country-reported national forest statistics questions from the FAO Global Forest Resources Assessment 2025 (21 variables across 7 families, all 236 countries).

Screenshot 2026-05-25 at 12 16 01

Routing: orchestrator loads the skill on FAO-shaped queries → `pick_aoi` → `pick_fra_variable` (new subagent) → `query_fra_data` (new primitive tool) → `generate_insights`.

Three composability rules were lifted out of the skill into higher layers during integration so they benefit every chart-producing flow, not just FAO:

1. **Universal reply structure** (orchestrator policy): after
   `generate_insights`, compose reply as intro → chart markers → insight
→ skill caveats → sources → follow-ups.
2. **Universal dataset citation** (orchestrator policy): cite `state["dataset"]["citation"]` as a markdown link automatically. Every dataset YAML already has a `citation` field; FAO's skill body now only adds the FRA-unique extras (Data Explorer, Programme page).
3. **Chart-marker convention** (analyst ToolMessage): the analyst emits a `Chart placement:` instruction with the actual `insight_id` so the orchestrator can place `[Chart <uuid>]` markers for inline FE injection.

Also includes a universal analyst hardening (`NaN → None` sanitization before JSONB persist — fixes `asyncpg.InvalidTextRepresentationError` on sparse FAO time-series).

## Layering

| Concern | Lives in |
|---|---|
| Routing | Orchestrator system prompt (`graph.py`) |
| Universal reply structure & citation | Orchestrator policy |
| Chart-marker convention + UUID | Analyst ToolMessage |
| FRA workflow / capability limits / wording | `fao-fra` skill body |
| FRA-unique citation extras (Data Explorer, Programme) | `fao-fra` skill body |
| Variable-selection reasoning | `pick_fra_variable` subagent |
| FAO API contract | `fao_fra_client` + `query_fra_data` |

## What changed

**New (FAO-specific):**
- `src/agent/skills/skills_md/fao-fra.md` — skill recipe
- `src/agent/subagents/pick_fra_variable/{__init__,tool,prompts,variable_map}.py` — variable-selection subagent (21 vars)
- `src/agent/tools/query_fra_data.py` — primitive tool (no internal LLM)
- `src/agent/datasets/catalog/fao_fra_2025.yml` — dataset config
- `src/agent/datasets/handlers/fao_fra_client.py` — async httpx client

**Modified (universal):**
- `src/agent/graph.py` — register new tool/subagent; one Routing row; universal reply structure + universal citation rule in `# Policy`
- `src/agent/subagents/analyst/tool.py` — NaN sanitization + `Chart placement:` instruction with `insight_id`
- `src/agent/{tools,subagents}/__init__.py` — re-exports
- `tests/unit/agent/tools/skills/test_skills.py` — registry now 4 skills

**Tests (new):**
- `tests/unit/agent/datasets/handlers/test_fao_fra_client.py` (18)
- `tests/unit/agent/tools/fao_fra/test_pick_fra_variable.py` (7)
- `tests/unit/agent/tools/fao_fra/test_query_fra_data.py` (10)
- `tests/evals/test_fao_routing.py` (11 cases, live LLM)

### DIFF
--- a/src/agent/datasets/catalog/fao_fra_2025.yml
+++ b/src/agent/datasets/catalog/fao_fra_2025.yml
@@ -1,0 +1,151 @@
+dataset_id: 10
+dataset_name: FAO Global Forest Resources Assessment (FRA) 2025
+# This dataset is accessed via the `query_fra_data` tool inside the `fao-fra`
+# skill, NOT through the pick_dataset → pull_data pipeline. The flag below is
+# advisory: when `src/ingest/embed_datasets.py` is next rebuilt, it should be
+# updated to honor `pick_dataset_excluded` so FAO does not appear in the
+# pick_dataset RAG candidate set.
+pick_dataset_excluded: true
+context_layers: null
+parameters: null
+variables: []
+tile_url: null
+analytics_api_endpoint: null
+function_usage_notes: ''
+license: CC BY-NC-SA 3.0 IGO
+geographic_coverage: 236 countries and territories (global)
+resolution: National level (country-reported statistics)
+update_frequency: Every 5 years
+content_date: 1990–2025 (reporting years 1990, 2000, 2010, 2015, 2020, 2025)
+start_date: '1990-01-01'
+end_date: '2025-12-31'
+content_date_fixed: true
+keywords:
+  - forest area
+  - deforestation
+  - carbon stock
+  - biomass
+  - growing stock
+  - forest ownership
+  - FAO
+  - FRA
+  - national statistics
+  - country-reported
+
+selection_hints: >-
+  Use for nationally-reported, country-level forest statistics including total
+  forest area, carbon stocks, biomass, growing stock, ownership, designated
+  management, and disturbances. This is FAO-standardised country inventory data,
+  NOT remote-sensing pixel data. It does NOT support sub-national analysis or
+  custom date ranges. Use pull_data / GFW datasets for sub-national or
+  time-series analysis. FRA data is the authoritative source for
+  officially-reported national forest figures, FAO statistics, government
+  inventories, and SDG 15 indicators.
+
+prompt_instructions: >-
+  FRA 2025 is country-reported national inventory data covering 236 countries.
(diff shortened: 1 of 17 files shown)


01painadam:
This was bigger / more complex than intended but @srmsoumya @yellowcap I wanted to have a go at adding a composable skill+tool to try and understand what the best practise are here - would love to hear your thoughts.

Also, adding a non-WRI dataset with high value (many GFW users want comparisons between self-reported vs. remote sensing data)

yellowcap (changes requested):
Nice overall test for extending the agent! 🚀 

The main comment is that we have introduced the [data pull handlers](https://github.com/wri/project-zeno/tree/main/src/agent/datasets/handlers) exactly for this case, where we want to include new data backends like this one without the need of adding new tools. 

Can you try to update this to create a new dataset handler for the FAO dataset, ideally we would not need new tools for this at all (pick FAO, pull with custom handler, make insight).

I.e. make a new handler and add it here
https://github.com/wri/project-zeno/blob/main/src/agent/tools/pull_data.py#L38

Although admittedly, we have made strong assumptions around the pull data output structure that are tailored towards the data api, and might not fit this case.

01painadam:
> Nice overall test for extending the agent! 🚀
> 
> The main comment is that we have introduced the [data pull handlers](https://github.com/wri/project-zeno/tree/main/src/agent/datasets/handlers) exactly for this case, where we want to include new data backends like this one without the need of adding new tools.
> 
> Can you try to update this to create a new dataset handler for the FAO dataset, ideally we would not need new tools for this at all (pick FAO, pull with custom handler, make insight).

Ah nice, I didn't catch that! Ok, I'll give it a go - I'll make a separate PR though, as its fundamentally a different approach

01painadam:
@yellowcap Needed some work to make existing tools more generic (to be compatible across FAO and Analytics apis) but it works - refactor here https://github.com/wri/project-zeno/pull/690

PR #687, opened 2026-05-25
PZB-922: Client disconnect handling

yellowcap:
## What

  When a client disconnects mid-stream (tab closed, user hits stop), the LangGraph graph was continuing to run and
  burning tokens. The StreamingResponse generator kept executing inside the server with no way to abort it.

##   Abort on client disconnect

  A new guarded_stream wrapper (src/api/streaming.py) polls request.is_disconnected() between each yielded chunk. When  the client is gone, it breaks out of the loop and calls aclose() on the upstream generator, which propagates a GeneratorExit into LangGraph and cancels the in-flight LLM call.

##   Why now

  This is a cheap change that has outsized operational impact — disconnect handling prevents wasted LLM spend on  abandoned sessions, and will allow a stop button imlementation that is purely client side.

##   Changes

```

  ┌───────────────────────────────────┬──────────────────────────────────────────────┐
  │               File                │                    Change                    │
  ├───────────────────────────────────┼──────────────────────────────────────────────┤
  │ src/api/streaming.py              │ New — guarded_stream async generator         │
  ├───────────────────────────────────┼──────────────────────────────────────────────┤
  │ src/api/routers/chat.py           │ Wrap stream_chat with guarded_stream         │
    ├───────────────────────────────────┼──────────────────────────────────────────────┤
  │ tests/unit/test_guarded_stream.py │ 6 unit tests for disconnect/aclose behaviour │
    └───────────────────────────────────┴──────────────────────────────────────────────┘

```

##   Testing

```
uv run pytest tests/unit/test_guarded_stream.py tests/unit/test_stream_chat.py -v
```

### DIFF
--- a/src/api/routers/chat.py
+++ b/src/api/routers/chat.py
@@ -11,6 +11,7 @@ from src.api.data_models import ThreadOrm
 from src.api.schemas import ChatRequest, QuotaModel, UserModel
 from src.api.services.chat import generate_thread_name, stream_chat
 from src.api.services.quota import check_quota, enforce_quota
+from src.api.streaming import guarded_stream
 from src.api.user_profile_configs.sectors import SECTOR_ROLES, SECTORS
 from src.shared.database import get_session_from_pool_dependency
 from src.shared.logging_config import bind_request_logging_context, get_logger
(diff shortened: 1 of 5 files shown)


batpad:
@yellowcap this makes a lot of sense and looks good to me!

batpad (approved):
I'd still like to test doing the abort on the client and having that successfully trigger the disconnect event - that should work fine but aborting a fetch can be a bit weird.

I think we should merge this anyways and I can try doing a basic PoC test on the client to see how aborting works.

PR #688, opened 2026-05-25
PZB-913: Improve dataset selection mechanism

yellowcap:
This changes the behaviour slightly:

1. If no match, return a single selection, dataset_id as None, with reason
2. If more than one match, return list with reasons (for providing options after)
3. If one exact match, business as usual

This is better setup for asking the user for picking one afterewards.

### DIFF
--- a/src/agent/subagents/pick_dataset/tool.py
+++ b/src/agent/subagents/pick_dataset/tool.py
@@ -90,7 +90,7 @@ async def select_best_dataset(
     start_date: Optional[str] = None,
     end_date: Optional[str] = None,
     aoi_selection=None,
-) -> DatasetSelectionResult:
+) -> DatasetSelectionResult | list[DatasetOption]:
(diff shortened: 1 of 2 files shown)


src/agent/subagents/pick_dataset/tool.py
```diff
@@ -224,6 +229,49 @@ class DatasetSelector:
...
+                        ToolMessage(
+                            f"No dataset selected: {selection_result[0].reason}",
+                            tool_call_id=tool_call_id,
+                        )
+                    ]
+                }
+            )
+        elif len(selection_result) > 1:
+            msgs = "Please select one of the following datasets:"
+            name_by_id = {
```
jterry64, line 250: Wouldn't we need to add to the state for the FE to use for nudge buttons? I'll merge in your PR and then add that today

jterry64 (approved):

PR #689, opened 2026-05-26
bump version in pyproject.toml

batpad:
Missed bumping version in previous PR so now `lint` on `main` is failing.

Am not sure why that didn't error on that PR.

Also, @yellowcap I see we've moved to a version like `2026.05.22` - are we thinking of versioning by date or so? What happens when we have multiple releases on the same day?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.5.22"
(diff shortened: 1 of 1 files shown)


yellowcap:
Its CalVer bump with `.x` addition increment for multiple from one day. It's documented here:

https://github.com/wri/project-zeno#versioning

PR #690, opened 2026-05-26
FAO FRA skill using dataset handler

01painadam:
## Summary

Adds FAO Global Forest Resources Assessment 2025 to Zeno's analytics capabilities, country-reported national forest statistics across **21 variables** (forest area, carbon, restoration) for 236 countries.

image

This is a v2 of the FAO design, The parallel-tool / skill / subagent approach from [skill/fao_fra](https://github.com/wri/project-zeno/pull/686) is refactored to use existing handler pattern documented in `src/agent/datasets/handlers/README.md`, The idea is that this introduces **No new tools, no new subagent, no new skill.** FAO becomes another dataset addition.

- New `FAOFRAHandler(DataSourceHandler)` plugged into `DataPullOrchestrator` alongside `AnalyticsHandler`.
- The 21 variables live as `context_layers` entries on the FAO dataset YAML; `pick_dataset`'s existing structured-output LLM picks one.
  
The pipeline for an FAO query is the standard one: pick_aoi → pick_dataset (FAO + context_layer) → pull_data (→
FAOFRAHandler) → generate_insights

## Separation of concerns at a glance

| Concern | Owns it |
|---|---|
| Routing | Unchanged from main — `analyze` skill |
| Variable selection (21 FAO variables) | `pick_dataset`'s existing `context_layers` mechanism |
| FAO API contract | `FAOFRAHandler` + `fao_fra_client` |
| Per-variable date ranges (disturbances 2002-2020, fire 2007-2019, 2025-only tables) | Per-context_layer `start_date`/`end_date` in YAML, honoured by existing `revise_date_range` |
| Universal chart-reply quality (citations, cautions, marker placement) | Orchestrator policy + analyst ToolMessage — benefits GFW datasets too |

## What changed

**New (FAO-specific, ~870 lines source + 800 lines tests):**
- `src/agent/datasets/catalog/fao_fra_2025.yml` — 21 `context_layers` with `fao_table`, `fao_variables`, per-variable date ranges
- `src/agent/datasets/handlers/fao_fra_client.py` — async httpx client; parses both single-year and period-key (`"1990-2000"`) FAO response shapes
- `src/agent/datasets/handlers/fao_fra_handler.py` — `DataSourceHandler` implementation
- `tests/unit/agent/datasets/{handlers/test_fao_fra_client.py, handlers/test_fao_fra_handler.py, test_fao_fra_catalog.py}` adds 39 new tests

**Modified — universal (benefits every dataset, required fro FAO compatibility):**
- `src/agent/tools/pull_data.py` — register handler; widen statistics build to accept inline-record output alongside the existing URL-backed pattern
- `src/agent/subagents/pick_dataset/schema.py` — `tile_url` and `analytics_api_endpoint` now `Optional[str]` (non-tile / non-analytics datasets are now first-class)
- `src/agent/subagents/pick_dataset/tool.py` — guard `None` tile_url in `get_tile_services_for_dataset`; harden `selected_context_layer` lookup
- `src/agent/subagents/analyst/tool.py` — NaN→None sanitization before JSONB persist; surface `dataset["cautions"]` in ToolMessage so chat replies can cite them; chart marker placement instruction with the per-chart identifier
- `src/agent/graph.py` — orchestrator policy gains: cite `state["dataset"]["citation"]`, surface relevant dataset cautions

## Reply quality (also universal) — what's different in the chat narrative

| Element | Before | After |
|---|---|---|
| 1–2 sentence chart summary | ✅ | ✅ (unchanged) |
| Follow-ups | ✅ | ✅ (unchanged) |
| Inline `[Chart …]` marker | ❌ | ✅ — hardens Insight card injection in FE |
| Dataset cautions surfaced from YAML | ❌ | ✅ |
| Dataset citation as markdown link | ❌ | ✅ |

A reply that was *"Brazil's forest area declined from 618M ha to 486M ha over 1990–2025. Follow-ups: …"* now reads roughly:

> *"I've generated an insight from FAO FRA 2025. **[Chart …]** Brazil's forest area declined from ~618M ha in 1990 to ~486M ha in 2025. Note: FRA figures are country-reported, not satellite-derived; FAO's 'forest' definition differs from GFW tree cover. Source: [FAO. 2025. Global Forest Resources Assessment 2025](https://doi.org/10.4060/cd3148en). Follow-ups: …"*

Nothing from main is removed; three additive elements, paid for at the orchestrator-policy / analyst-ToolMessage boundary, and no per-skill duplication.

## Real contract widenings exposed by FAO

Most of the friction-fix commits are contract loosening that any future non-analytics handler would have hit (FAO is just the first dataset to exercise them). Each is small and standalone:

| Commit | Latent assumption | Why failed for FAO |
|---|---|---|
| `30a4697` | `tile_url` always begins with `http` | FAO has `tile_url: null` |
| `4ef09b7` | `DatasetSelectionResult.{tile_url, analytics_api_endpoint}`  are non-optional `str` | FAO has both None |
| `f180608` | Handler-specific fields on `context_layers` survive into `state["dataset"]` | They don't — the schema projects to `{name, tile_url}` only |
| `c1b9052` | FAO API uses single-year time keys uniformly | Rate-of-change tables (`forestAreaChange`) use period keys (`"1990-2000"`) |

After this PR the handler contract is more generic.


src/agent/subagents/pick_dataset/schema.py
```diff
@@ -115,8 +115,13 @@ def validate_parameters_for_dataset(self) -> "DatasetOption":
 
 
 class DatasetSelectionResult(DatasetOption):
-    tile_url: str = Field(
-        description="Tile URL of the dataset that best matches the user query.",
+    # `tile_url` and `analytics_api_endpoint` are None for datasets that
+    # don't render a map layer or don't go through the GFW analytics
+    # API (e.g. FAO FRA, which is country-reported statistics served by
+    # its own client via DataSourceHandler).
```
gtempus, line 134: This feels like the DatasetSelectionResult abstraction is already breaking down. This will likely cause more "paranoid" code checks in the callers.

src/agent/subagents/pick_dataset/tool.py
```diff
@@ -376,11 +376,15 @@ def get_tile_services_for_dataset(
     selection_result, selected_row, start_date, end_date
 ):
     context_layers = []
+    # `tile_url` is None for datasets without a map layer (e.g. FAO FRA,
+    # which is country-reported statistics with no pixel imagery to
+    # render). Leave it None so downstream rendering knows to skip the
+    # tile layer.
```
gtempus, line 451: We see the abstraction further breaking down with the first additional 'handler'. The two guards added below confirm it. `context_layer` used to be invariant, now it's nested again within the existing `if`.

src/agent/tools/pull_data.py
```diff
@@ -37,6 +38,7 @@ class DataPullOrchestrator:
     def __init__(self):
         self.handlers = [
             AnalyticsHandler(),
+            FAOFRAHandler(),
```
gtempus, line 44: `DataPullOrchestrator` is a "Chain of Responsibility" and really shouldn't know about its handlers, they should be passed in during construction.

src/agent/datasets/handlers/fao_fra_client.py
```diff
@@ -0,0 +1,218 @@
...
+
+This client is used directly by `src.agent.tools.query_fra_data` rather than
+through `DataPullOrchestrator` — FAO's response shape and request semantics
+(country-level only, fixed reporting years, no parameter sweeps) don't fit the
+analytics-API contract that handler exposes.
+"""
+
+from typing import Optional
+
+import httpx
```
gtempus, line 22: This dependency shouldn't be added at this level of abstraction. Probably worth rethinking a client that accepts an HTTP specialist instead. Plus, it avoids the mocking in the tests and better unit testing instead of specific helper functions.

src/agent/datasets/handlers/fao_fra_client.py
```diff
@@ -0,0 +1,218 @@
...
+        raise FAOAPIError(
+            "FAO FRA API requires authentication for this endpoint. "
+            "Please contact the GNW team."
+        )
+    if response.status_code >= 400:
+        raise FAODataNotFoundError(
+            f"FAO FRA API returned {response.status_code} for "
+            f"{iso3} / {table}. The country or variable may not be "
+            "available in FRA 2025."
+        )
```
gtempus: As mentioned above, the client at this level shouldn't know all of the HTTP particulars. Parsing the payload is _fine_ but it shouldn't know that it came across the wire. Again, it'll help keep the code focused on the information coming in, not on how it's getting to the app. Plus, again, the testing is easier.

src/agent/datasets/handlers/fao_fra_handler.py
```diff
@@ -0,0 +1,190 @@
...
+handler just translates the chosen variable into a FAO API call and
+returns the records inline.
+
+Why not use the analytics-API contract:
+- FAO is country-level only (the analytics API supports admin/KBA/WDPA/etc).
+- FAO has fixed reporting years per variable (the analytics API is
+  continuous time-series with custom date ranges).
+- FAO responses are small (country × years × variables) and complete in
+  one request — no async polling, no source_url indirection.
+"""
```
gtempus, line 15: Fine as long as the interface stays the same (it doesn't seem to).

src/agent/datasets/handlers/fao_fra_handler.py
```diff
@@ -0,0 +1,190 @@
...
+from src.agent.datasets.config import DATASETS
+from src.agent.datasets.handlers.base import (
+    DataPullResult,
+    DataSourceHandler,
+)
+from src.agent.datasets.handlers.fao_fra_client import (
+    FAOAPIError,
+    FAODataNotFoundError,
+    _build_source_url,
+    fetch_fra_data,
```
gtempus: This ties the handler to the implementation-specific (HTTP) client. The `FAOFRAHandler` is a class, treat it that way (data encapsulation, information hiding, message passing, etc).

src/agent/datasets/handlers/fao_fra_handler.py
```diff
@@ -0,0 +1,190 @@
...
+    if catalog_entry is None:
+        return None
+    for layer in catalog_entry.get("context_layers") or []:
+        if layer.get("value") == value:
+            return layer
+    return None
+
+
+class FAOFRAHandler(DataSourceHandler):
+    """Routes FAO FRA 2025 dataset requests to the FAO public API."""
```
gtempus, line 82: Should the handler know it's doing that?

src/agent/datasets/handlers/fao_fra_handler.py
gtempus: Again, tests would be far more focused and simpler.

src/agent/datasets/handlers/fao_fra_handler.py
gtempus: Actually, should the handler know any of what it takes responsibility for in `pull_data`? It has the job of handling, but here it's knowledgeable of the domain, understands the data is coming over HTTP, it knows about URLs and what should be used as a source url, etc. It's taking on way too much work.

gtempus: Tests would be far more focused and simpler.

tests/unit/agent/datasets/handlers/test_fao_fra_client.py
```diff
@@ -0,0 +1,330 @@
...
+def _patch_transport(monkeypatch, handler):
+    """Replace httpx.AsyncClient with one driven by a MockTransport."""
+    transport = httpx.MockTransport(handler)
+    original_init = httpx.AsyncClient.__init__
+
+    def patched_init(self, *args, **kwargs):
+        kwargs["transport"] = transport
+        return original_init(self, *args, **kwargs)
+
+    monkeypatch.setattr(httpx.AsyncClient, "__init__", patched_init)
```
gtempus: We shouldn't have to do this. But the current design forces us to.

tests/unit/agent/datasets/handlers/test_fao_fra_client.py
```diff
@@ -0,0 +1,330 @@
...
+    url = _build_source_url("BRA", "extentOfForest")
+    assert url.startswith("https://fra-data.fao.org/api/explorer/data?")
+    assert "countryISOs[]=BRA" in url
+    assert "tableNames[]=extentOfForest" in url
+    assert "assessmentName=fra" in url
+
+
+# ---------------------------------------------------------------------------
+# _parse_response — happy path and edge cases
+# ---------------------------------------------------------------------------
```
gtempus, line 106: Why test helper functions of a larger module? It could be feedback that the module is doing too much.

tests/unit/agent/datasets/handlers/test_fao_fra_client.py
```diff
@@ -0,0 +1,330 @@
...
+    by_period = {
+        r["period"]: r
+        for r in records
+        if r["variable"] == "forestAreaNetChange"
+    }
+    assert set(by_period.keys()) == {"1990-2000", "2010-2015"}
+    # End-year is used for x-axis ordering
+    assert by_period["1990-2000"]["year"] == 2000
+    assert by_period["2010-2015"]["year"] == 2015
+    assert by_period["1990-2000"]["value"] == pytest.approx(-400.0)
```
gtempus: Try to stay away from multiple asserts. As soon as the first one fails, the test stops. It's not possible to see how many of the asserts are failing. Much better to either have focused tests or have a nice diff of the data structure.

tests/unit/agent/datasets/handlers/test_fao_fra_client.py
```diff
@@ -0,0 +1,330 @@
...
+    captured: dict = {}
+
+    def handler(request: httpx.Request) -> httpx.Response:
+        captured["url"] = str(request.url)
+        return httpx.Response(200, json=_ok_payload("BRA", "extentOfForest"))
+
+    _patch_transport(monkeypatch, handler)
+    records = await fetch_fra_data(
+        "BRA", "extentOfForest", variables=["forestArea"], year=None
+    )
```
gtempus: We really shouldn't try to assume the behavior of libraries we don't own.

tests/unit/agent/datasets/handlers/test_fao_fra_handler.py
```diff
@@ -0,0 +1,301 @@
...
+
+    assert result.success
+    assert isinstance(result.data, dict)
+    records = result.data["data"]
+    assert len(records) == 2
+    # Handler tags each record with the human-readable AOI name
+    assert all(r.get("aoi_name") == "Brazil" for r in records)
+    assert result.data_points_count == 2
+    assert result.analytics_api_url
+    assert "BRA" in result.analytics_api_url
```
gtempus: It's okay to mock at this level but now the test has multiple facets it's trying to verify. Interface specification with the `call["blah"]` is different from the result processing. So many asserts will cause issues when the first assert fails since it's not possible to see how _incorrect_ the test actually is.

gtempus (commented):
Thanks for the example handler code, @01painadam.
My comments are entirely regarding the structure and intention of the existing design. Seems like there are lots of opportunities to improve how we build out the application.
80% of a project's cost is the maintenance of the existing codebase.

01painadam:
@gtempus poked around with this a bit - I think I addressed your bigger concerns. Lemme know what you think.

01painadam:
Closing - no re-review and extremely stale at this point. Keeping the branch up in remote since this is a valuable feature.

PR #693, opened 2026-05-26
PZB-835: Add tree_cover_loss_from_fires.yml to describe TCLF dataset

danscales:
Add tree_cover_loss_from_fires.yml to describe TCLF dataset

We don't yet know exactly what we will do for the analytics backend yet, so I didn't try to add anything there (just kept what was there for tree_cover_loss.yml).

I put a short description under 'description', and the full description under methodology, since that seemed to be the pattern elsewhere.


src/agent/datasets/handlers/analytics_handler.py
gtempus, line 299: @danscales, following up on your comment tha the analytics API is still being decided, this is close to what updates need to be made. Basically, adding `fire` as the intersection:
```python
elif dataset.get("dataset_id") in [
  TREE_COVER_LOSS_ID,
  TREE_COVER_LOSS_BY_DRIVER_ID,
  TREE_COVER_LOSS_BY_FIRES_ID,
]:
forest_filter = None

if dataset.get("context_layer") == "primary_forest":
  forest_filter = "primary_forest"

elif dataset.get("context_layer") == "intact_forest":
  forest_filter = "intact_forest"
intersections = []
if dataset.get("dataset_id") == TREE_COVER_LOSS_BY_DRIVER_ID:
  intersections = ["driver"]
if dataset.get("dataset_id") == TREE_COVER_LOSS_BY_FIRES_ID:
  intersections = ["fire"]
```
Check with @dmannarino to confirm.

I would've used the _suggest_ markup for this comment but it seems that GH doesn't do multiline suggests very well (or it might be because there are no changes here for this PR). 🤷

src/agent/datasets/catalog/tree_cover_loss_from_fires.yml
```diff
@@ -0,0 +1,100 @@
...
+start_date: '2001-01-01'
+end_date: '2025-12-31'
+content_date_fixed: false
+keywords:
+- deforestation
+- forest loss
+- disturbance
+- fires
+analytics_api_endpoint: /v0/land_change/tree_cover_loss/analytics
+prompt_instructions: 'Reports gross annual loss of tree cover due to fires ≥ 5 m height (2001-2025). Show yearly loss with bar charts or
```
jterry64: fyi, I think this might be a legacy field when all the prompt instructions used to be in one field

danscales: It does seem like prompt_instructions is not actually used anywhere (including in pick_dataset and analyst tools) if code_instructions is non-empty for the same dataset, and currently all datasets do have a code_instructions and a prompt_instructions. I could make another PR to remove prompt_instructions everywhere.

src/agent/datasets/handlers/analytics_handler.py
danscales: OK, thanks, I added that change to my PR, and will show it to Daniel to see if that's what he's aiming for.  Seems reasonable for the change to be in my PR, since it depends on the TREE_COVER_LOSS_BY_FIRES_ID to be defined and doesn't hurt any other datasets, but can revert this analytics_handler.py change before checkin if we want to sequence it differently.

dmannarino: Yep, interface posted by @gtempus is correct. intersections=["fire"].

jterry64 (approved):

danscales:
I looked at Justin's added prompt changes to my PR, and they look good.  So, I think this is ready to check in.

PR #696, opened 2026-05-27
Fix markers for chart references in AI response

yellowcap:
This PR shoudl fix the marker handling so that the frontend can nicely display links to insights / charts in the responses.


yellowcap:
this is stale, can redo if needed

PR #697, opened 2026-05-29
PZB-914: Use Retry-After header from analytics API, wait up to 30 seconds

jterry64:
Currently, the analytics handler is using an exponential backoff for polling the analytics API. This likely makes the latency much slower for pull_data, because it'll increase the poll time to a couple seconds after the first two tries. The analytics API includes a "Retry-After" header specifying when to poll, so changing this function to respect that.


src/agent/datasets/handlers/analytics_handler.py
danscales, line 418: Just curious, if the polling failed, will result every have data.link? In the pending case, does result.data.link have something useful (maybe the new link that isn't ready yet?)  If not, do you just want to print endpoint_url here, rather than result.data.link? 

No need to make any change here, unless you want to.

danscales (approved):

src/agent/datasets/handlers/analytics_handler.py
jterry64: It should always have a link to a resource with status set to "pending" or "failed"

gtempus (approved):
Nice! Yeah, this should speed things up a little. 🐎

PR #699, opened 2026-06-02
Pull out dataset instructions and repeat on top and bottom

jterry64:
I've been playing around with a couple ideas to improve dataset suggestions and selection. While reading some [prompt guidance](https://developers.openai.com/cookbook/examples/gpt4-1_prompting_guide), I found the tip (although might not apply to newer/bigger models):

"Especially in long context usage, placement of instructions and context can impact performance. If you have long context in your prompt, ideally place your instructions at both the beginning and end of the provided context, as we found this to perform better than only above or below. If you’d prefer to only have your instructions once, then above the provided context works better than below."

I wanted to try it out quickly, and I also noticed all our dataset-specific instructions are actual stuffed in the middle in a nested csv structure. That might be why we've struggled a lot with it following instructions in the dataset YAML; it viewed the instructions as "data" and not as "rules".

I tested a few things, but I found below significantly increased the quality of dataset selection and suggestions:
- Consolidate the existing rules into a smaller bullets
- Pull out the selection hints from the candidate datasets structure 
- Put all the rules and selections hints both on top and bottom of the full dataset info

See evals below on three new evals I added to ask vague questions expecting suggestions. I did 5 trials, and found staging very inconsistent and score poorly on 2/3 questions, whereas my local changes now score perfectly. This does increase the context size, but not sure if it's significant enough for this internal model call to matter in terms of latency or cost.

Against local:
[eval_results_gold_sample_3_workers_9_offset_93_20260602_160058_summary.csv](https://github.com/user-attachments/files/28529550/eval_results_gold_sample_3_workers_9_offset_93_20260602_160058_summary.csv)
Against staging:
[eval_results_gold_sample_3_workers_9_offset_93_20260602_160058_summary.csv](https://github.com/user-attachments/files/28529567/eval_results_gold_sample_3_workers_9_offset_93_20260602_160058_summary.csv)

Also made a few test and YAML changes, since now it started always following selection hints, pointing out a few weird questions/instructions, and I think also we forgot to update to use TCL from fires instead of drivers for fire-related questions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.6.1.1"
+version = "2026.6.2"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.6.2"
+version = "2026.6.2.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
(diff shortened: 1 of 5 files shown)


solomon-negusse (approved):
Nice! This also did better on a challenging set from the online llm judge failures.

danscales (approved):

PR #700, opened 2026-06-03
PZB-931: Ingest Langfuse traces and expose in Zeno API

batpad:
This adds:

 - Schemas to hold langfuse traces in the postgres db
 - An ingest CLI command (that will be run as a daily cron or so) to fetch trace data from Langfuse and ingest into Postgres
 - API endpoints (auth-gated) to fetch trace data from the API as well as aggregated analytics

This tries to replicate the functionality in https://github.com/01painadam/tracey, but bringing it into our db and API.

This bit was heavily Claude-d, and I think I need to spend some time reviewing this closely. Once we are reasonably satisfied with this, next steps would be:

 - Deploy this. Manually run CLI to fetch traces, test that API endpoints seem alright
 - Create k8s CronJob to update with traces every day (or more frequently if we want)
 - Move the `tracey` streamlit app to reading from our API instead of querying Langfuse directly

Once that all works and seems good, we can discuss / ticket if we want to make some of the trace analysis available in our frontend.

cc @01painadam

### DIFF
--- a/db/alembic/versions/8391583bd224_add_langfuse_traces_analytics_view.py
+++ b/db/alembic/versions/8391583bd224_add_langfuse_traces_analytics_view.py
@@ -1,0 +1,68 @@
+"""add langfuse_traces_analytics view
+
+Revision ID: 8391583bd224
+Revises: a7d17cad8386
+Create Date: 2026-06-03 15:15:32.632677
+
+The sanctioned analytics surface over langfuse_traces. It exposes the turn-level
+derived columns plus a computed ``is_final_turn_in_thread`` flag so consumers can
+pick one representative row per thread (for cumulative fields) without re-deriving
+the dedup logic. Session-less traces (null session_id) are treated as singleton
+threads via COALESCE(session_id, id).
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "8391583bd224"
+down_revision: Union[str, None] = "a7d17cad8386"
+branch_labels: Union[str, Sequence[str], None] = None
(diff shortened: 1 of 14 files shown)


batpad:
The latest commit here makes a change to NOT ingest all the raw individual trace data into postgres. It now only ingests data needed for analytics and basic metadata for each trace to enable search and filtering. Querying individual trace data from LangFuse seems straightforward enough, and this way we avoid duplicating all the individual trace data into postgres, store only computed aggregates, and the single source of truth remains the Langfuse backend.

This allays my fears of putting too much into Postgres, duplicating data, etc. I _think_ I feel okay with this approach now.

Will go over with @01painadam , ensure this hits the use-cases we are looking for - discuss some inconsistencies I noticed with the `tracey` parsing logic and confirm I got things right, and then I think we can go ahead with deploying this on staging and I can manually run the ingest / backfill on staging and test things before creating the CronJob container to run regularly and moving to frontend implementation.

batpad:
Am having a hard time figuring out how to test this properly locally beyond the automated tests in the PR. I think I'd like to deploy this to staging and run the CLI commands manually to test. If that goes well, I can go ahead with setting up the cron-job infrastructure and other next steps.

@yellowcap - will it be possible to get a quick review? (cc @sunu if you do have a bit of time, would really appreciate a really quick glance and sanity / gut-check from you).

yellowcap (approved):
Looks good, have not cross checked if this is what the `tracey` app uses, but from what I understood, @batpad you mapped the variables from that app. So lets put this in and test it instaging! 🚀

01painadam:
Nice - I'll test with Tracey

PR #701, opened 2026-06-03
Hotfix/tclf dataset name

01painadam:
Renames dataset: TCL _by fires_ --> TCL _due to fires_ to match Flagship

Enables dataset matching on FE

### DIFF
--- a/src/agent/datasets/handlers/analytics_handler.py
+++ b/src/agent/datasets/handlers/analytics_handler.py
@@ -136,7 +136,7 @@ SLUC_EMISSION_FACTORS_ID = [
 TREE_COVER_LOSS_BY_FIRES_ID = [
     ds["dataset_id"]
     for ds in DATASETS
-    if ds["dataset_name"] == "Tree cover loss by fires"
+    if ds["dataset_name"] == "Tree cover due to fires"
 ][0]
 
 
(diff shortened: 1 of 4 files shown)


01painadam:
Ignore the revert, I got muddled 😓

solomon-negusse (approved):

src/agent/datasets/handlers/analytics_handler.py
```diff
@@ -136,7 +136,7 @@ class BooleanResponse(BaseModel):
 TREE_COVER_LOSS_BY_FIRES_ID = [
     ds["dataset_id"]
     for ds in DATASETS
-    if ds["dataset_name"] == "Tree cover loss by fires"
+    if ds["dataset_name"] == "Tree cover due to fires"
```
gtempus:
```suggestion
    if ds["dataset_name"] == "Tree cover loss due to fires"
```

gtempus (changes requested):

src/agent/datasets/handlers/analytics_handler.py
01painadam: Ah, thats the issue. Thanks Gary!

gtempus:
I think you're missing your `pyproject.toml` update. The pre-commit assigns a new calendar version.

gtempus (approved):
Thanks for knocking this out, @01painadam 💥

PR #704, opened 2026-06-05
Analyze endpoint

jterry64:
See mermaid diagram for details. Add analyze endpoint to create Insight Chart resources based off a dataset and an AOI. Uses Long Running Operation (LRO) pattern, with request-acknowledge for /analyze and a link to a long running /job resource. /job resource will return links to /insight resource when completed.


src/api/routers/analyze.py
```diff
@@ -0,0 +1,54 @@
...
+        aois=[aoi.model_dump() for aoi in request.aois],
+        dataset_id=request.dataset_id,
+        start_date=request.start_date,
+        end_date=request.end_date,
+    )
+
+    if result.data.success and request.thread_id:
+        zeno = await fetch_zeno()
+        config = {"configurable": {"thread_id": request.thread_id}}
+        await zeno.aupdate_state(
```
jterry64: I tried a couple things here so the agent is aware there's been data loaded. However, because generate_insights creates charts and text, I couldn't get it to talk about the data without trying to make a chart. So we may need that tool split this sprint to make the agent by in sync with the default charts.

gtempus:
So much good design in here, @jterry64! 💪 
Claude and I reviewed. Can you address the Auth issue and adopt depenendancy injection with FastAPI Depends in the routers? Then I think we'd be good to ship.

# PR Review: `analyze_endpoint` — Async Analysis LRO Pattern 

---

## Overall Grade: **B+**

This PR introduces a well-structured Long-Running Operation (LRO) pattern for geospatial analysis. The architectural instincts are strong — a clear port/adapter split, excellent API documentation, structured logging, and a focused test suite written against fakes rather than mocks. The issues below are real but fixable; none block the feature, though the authorization gap and the incomplete abstract port should be addressed before this ships.

---

## What's Good 🟢

### 1. Hexagonal Port/Adapter Design (Almost There)

The service layer correctly defines abstract ports:

```python
# services/job.py — port
class JobRepository(ABC): ...

# services/charts.py — port
class ChartGenerator(ABC): ...

# repositories/job_repository.py — adapter
class DBJobRepository(JobRepository): ...

# services/charts.py — adapter
class TCLChartGenerator(ChartGenerator): ...
```

`AnalysisJobRunner.__init__` takes the `JobRepository` interface and receives `AnalyzeService` by constructor injection — this is exactly right and makes unit testing trivial without mocks. The fake implementations in the test suite prove the ports are honoured cleanly.

---

### 2. LRO Pattern is Correct

`POST /api/analyze → 202/pending → GET /api/jobs/{id} → 200/completed → follow resource_url` is the textbook pattern for async HTTP. The `Retry-After: 1` header is a developer-experience detail that many teams forget. The OpenAPI response schema documents it properly via the `responses` dict on the route.

---

### 3. API Documentation is Excellent

The Mermaid sequence diagram in `src/api/docs/analyze_flow.md` is one of the best pieces of in-repo LRO documentation I've seen on this project. The endpoint docstrings are detailed enough to render clearly in Swagger UI. The `Field(description=...)` annotations on every schema field make the generated OpenAPI spec self-documenting. This sets a high standard for the rest of the API.

---

### 4. Structured Logging Throughout

```python
logger.info("analysis_job_started", job_id=..., user_id=..., dataset_id=..., ...)
logger.info("analysis_job_completed", job_id=..., duration_ms=..., ...)
logger.error("analysis_job_failed", severity="high", ...)
```

Every lifecycle transition emits a structured log event with the right context. `duration_ms` on completion makes p95/p99 latency tracking trivial. This is production-ready observability.

---

### 5. Test Quality

Tests cover the state-machine transitions (`pending → running → completed/failed`), chart generation edge cases (zero-value rows dropped), generator routing (dataset ID matching), and argument pass-through fidelity. The `FakeHandler`/`FakeJobRepository` fakes are in-file and clearly scoped — no heavy fixture overhead. `test_drops_rows_where_area_ha_is_zero` is a particularly good domain-rule test.

---

### 6. Defensive Fix in `analytics_handler.py`

```python
# Before (would KeyError if "name" absent)
format_id(item["src_id"]): item["name"].split(",")[0]

# After — safe fallback to src_id
format_id(item["src_id"]): item.get("name", item["src_id"]).split(",")[0]
```

Small change, high-value resilience fix. Good catch.

---

## Issues Requiring Attention 🔴

### 🔴 P0 — Authorization Gap on `GET /api/jobs/{id}`

Any authenticated user can poll any job by its UUID. The `get_job` handler checks only `if not job: raise 404` — there is no ownership check.

```python
# jobs.py — current code (vulnerable)
job = await repo.get_job(job_id)
if not job:
    raise HTTPException(status_code=404, detail="Job not found")
# ← no check that job.user_id == user.id
```

**Fix:**

```python
job = await repo.get_job(job_id)
if not job or job.user_id != user.id:
    raise HTTPException(status_code=404, detail="Job not found")
```

Return `404` (not `403`) for the not-owned case — don't reveal that the UUID exists. This must ship before the endpoint is reachable by external callers. `JobData` needs to carry `user_id` for this check (it currently does not).

---

### 🔴 P1 — `JobRepository` Abstract Port is Incomplete

`services/job.py` defines the `JobRepository` port but omits `create_job`, which is called directly on `DBJobRepository` in the router:

```python
# analyze.py — router bypasses the port
repo = DBJobRepository()        # concrete class, not the port
job_id = await repo.create_job(...)
```

The whole point of the port is that the router should only know about the `JobRepository` interface. If you ever swap `DBJobRepository` for an in-memory fake in integration tests, this call will break.

**Fix:** Add `create_job` to the abstract base in `services/job.py`:

```python
class JobRepository(ABC):
    @abstractmethod
    async def create_job(
        self, user_id: str, thread_id: Optional[str], type: JobType
    ) -> UUID: ...

    @abstractmethod
    async def update_job_status(self, job_id: UUID, status: JobStatus) -> None: ...

    @abstractmethod
    async def create_insight_resource(self, ...) -> None: ...

    @abstractmethod
    async def get_job(self, job_id: UUID) -> Optional["JobData"]: ...
```

The `get_job` method is also missing from the port (it's only on `DBJobRepository`), which means `jobs.py` also bypasses the abstraction.

---

## Issues to Address as the Feature Evolves 🟡

### 🟡 `JobData` / `JobResourceData` are in the Wrong Layer

These data classes currently live in `repositories/job_repository.py`:

```python
# repositories/job_repository.py — wrong home
@dataclass
class JobData:
    id: UUID
    type: JobType
    status: JobStatus
    ...
```

They are domain data transfer objects — the shape of a Job as the rest of the application sees it. They belong in `services/job.py` (the domain layer) or a dedicated `domain/` module, not in the repository adapter. The repository should import them from the domain, not define them. As written, moving to a different persistence backend requires changing where these types live.

---

### 🟡 Router Instantiates Repository Directly (Twice)

Both routers construct `DBJobRepository()` inline:

```python
# analyze.py
repo = DBJobRepository()
job_id = await repo.create_job(...)

# jobs.py
repo = DBJobRepository()
job = await repo.get_job(job_id)
```

Use FastAPI's `Depends()` to inject the repository — this makes it swappable in tests and keeps the router free of infrastructure concerns:

```python
def get_job_repository() -> JobRepository:
    return DBJobRepository()

@router.post("/api/analyze", response_model=JobResponse)
async def create_analysis_job(
    request: AnalyzeRequest,
    background_tasks: BackgroundTasks,
    user: UserModel = Depends(require_auth),
    repo: JobRepository = Depends(get_job_repository),
):
    ...
```

The `_make_runner()` factory at module level in `analyze.py` has the same problem — it constructs concrete infrastructure inside the router. Move it to a `Depends`.

---

### 🟡 Module-Level Singletons in the Router

```python
# analyze.py — top of module
handler = AnalyticsHandler()
generators = [TCLChartGenerator(TREE_COVER_LOSS_ID)]
```

This couples the router module to concrete infrastructure at import time and makes it impossible to override in tests without patching. Use `Depends()` or application lifespan state instead.

---

### 🟡 `start_date` / `end_date` are Unvalidated Strings

```python
class AnalyzeRequest(BaseModel):
    start_date: str  # accepts "banana"
    end_date: str
```

Pydantic handles date parsing natively. Use `datetime.date` and let the framework reject malformed inputs before they reach the service layer:

```python
from datetime import date

class AnalyzeRequest(BaseModel):
    start_date: date
    end_date: date
```

This also means the OpenAPI spec advertises the correct `"format": "date"` type to API consumers.

---

### 🟡 `aois` List Has No Minimum Length Validation

An empty `aois: []` will silently produce an empty result rather than a `422 Unprocessable Entity`. Add a minimum length constraint:

```python
aois: List[AreaOfInterest] = Field(
    min_length=1,
    description="One or more areas of interest to analyse."
)
```

---

### 🟡 Charts Are Untyped `list[dict]` Throughout

The chain `ChartGenerator.generate() → list[dict] → JobRepository.create_insight_resource(charts: list[dict]) → InsightChartOrm fields` relies on implicit string keys (`"xAxis"`, `"yAxis"`, `"type"`, etc.) with no type safety. When the second chart generator is added, it will be easy to produce a dict with the wrong key spelling and silently create malformed chart records.

Consider introducing a `ChartSpec` dataclass in `services/charts.py`:

```python
@dataclass
class ChartSpec:
    id: str
    title: str
    type: str   # or Literal["bar", "line", ...]
    x_axis: str
    y_axis: str
    data: list[dict]
    insight: str = ""
```

The repository can accept `list[ChartSpec]` and access fields by attribute, not string key.

---

### 🟡 `thread_id or ""` Coerces `None` to Empty String

```python
# job_repository.py
insight = InsightOrm(
    thread_id=thread_id or "",   # ← None becomes ""
    ...
)
```

This conflates "no thread" with "empty string thread", which may cause issues if the application later queries `WHERE thread_id = ''`. If `InsightOrm.thread_id` is nullable in the DB, pass `None` directly. If it isn't nullable in the schema, add a migration. Pick one truth.

---

### 🟡 `AnalyzeResponse` Schema is Dead Code

```python
# schemas.py — defined but never used
class AnalyzeResponse(BaseModel):
    success: bool
    message: str
    charts_data: Optional[List[dict]] = None
    source_urls: Optional[List[str]] = None
    codeact_parts: List[dict] = []
```

This appears to be a leftover from an earlier iteration. Remove it before it becomes a source of confusion for the next developer.

---

### 🟡 `updated_at` Has No Database-Level Trigger

The Alembic migration sets `server_default=now()` for `updated_at` but there is no `ON UPDATE` trigger. The SQLAlchemy ORM `onupdate=datetime.now` only fires when updates are made through the ORM session. A direct SQL `UPDATE` will leave `updated_at` stale. Either add a Postgres trigger or document the limitation explicitly in the model:

```python
# One option: add a trigger in the migration
op.execute("""
    CREATE OR REPLACE FUNCTION update_updated_at()
    RETURNS TRIGGER AS $$
    BEGIN NEW.updated_at = now(); RETURN NEW; END;
    $$ LANGUAGE plpgsql;

    CREATE TRIGGER jobs_updated_at BEFORE UPDATE ON jobs
    FOR EACH ROW EXECUTE FUNCTION update_updated_at();
""")
```

---

### 🟡 `AnalyzeService` Passes `query=""` to the Handler

```python
# analyze.py
result = await self._handler.pull_data(
    query="",           # ← artifact of the agent-facing interface
    dataset={"dataset_id": dataset_id},
    ...
)
```

`query` is a concept from the conversational agent world that has no meaning in a direct API call. This is the agent domain leaking into the API domain through a shared handler. This is acceptable for now, but as the feature grows, consider either:

1. Adding a `pull_data_direct()` method to the handler without the `query` parameter, or  
2. Splitting `AnalyticsHandler` so the API and agent have purpose-fit interfaces.

---

## Minor / Nits 🔵

- **`ResourceStatus` has only one value.** If a job fails mid-way, individual resources cannot reflect that failure state. Probably fine today; worth a comment explaining the intentional simplification.
- **`JobType` has one value.** No action needed — the single-value enum is the right forward-looking design. Just noting it is intentional.
- **`_column_to_rows` is a module-level private function in `charts.py`.** It's generic enough to belong in a `shared/` utility module if it gets used elsewhere. Fine here for now.
- **The migration revision ID `b3c1d2e4f5a6` looks hand-crafted** (sequential hex). Alembic generates random IDs for a reason — collision avoidance in long-lived projects. This is cosmetic but worth noting.

---

## Summary Table

| Area | Assessment |
|---|---|
| LRO design (`POST → poll → follow`) | ✅ Correct and well-documented |
| Port/adapter (hexagonal) structure | ✅ Good intent, incomplete port |
| Authorization on job access | ❌ Missing ownership check — P0 |
| FastAPI idiomatic DI (`Depends`) | ⚠️ Bypassed in both routers |
| OpenAPI / schema documentation | ✅ Excellent field-level descriptions |
| Input validation (`date`, `min_length`) | ⚠️ Gaps in `AnalyzeRequest` |
| Structured logging | ✅ Comprehensive |
| Unit tests | ✅ Good coverage, clean fakes |
| Type safety (chart dicts) | ⚠️ Untyped `list[dict]` chain |
| DB model correctness | ⚠️ `updated_at` trigger, `thread_id` null |

---

> The architecture direction is correct and the documentation quality is genuinely high. Fix the authorization gap and the incomplete port before merge. The rest can be tracked as follow-on tickets without blocking the feature.

gtempus:
@jterry64, here are two ADRs that explain the why behind our decisions to go with request/ack LROs with consequences for the api and for the client:

# ADR-001: Classify Geospatial Analysis Requests as Long-Running Operations

**Status:** Accepted  
**Date:** 2026-06-10  
**Deciders:** Engineering team  
**Related:** ADR-002 (Request/Acknowledge/Poll pattern)

---

## Context

When a user submits an analysis request — one or more areas of interest against a dataset over a date range — the system must fetch data from an external analytics API, transform the results, and persist charts and statistics. This pipeline has the following observed properties:

**Duration is externally bounded and unpredictable.** Response time from the upstream analytics service depends on the number of areas, the dataset complexity, and backend queue depth. Measured wall-clock times range from under one second to over thirty seconds for large multi-area requests. We do not control this ceiling.

**The HTTP request–response contract assumes bounded, short-lived exchanges.** Browser fetch timeouts, API gateway limits, and load-balancer idle-connection timeouts typically cut off at 30–60 seconds. Any request that might exceed this window is structurally incompatible with the synchronous HTTP model — not by policy, but by the timeout arithmetic of the infrastructure layers involved.

**Failure modes for long synchronous calls are ambiguous.** When a synchronous request times out, the client cannot distinguish between: the server crashed, the upstream service is slow, the network dropped, or the result was produced but the connection closed before delivery. This ambiguity makes retry logic unsafe — the client cannot know whether retrying would produce a duplicate or recover a lost result.

**Compute and connection resources are held for the full duration.** In a synchronous model, every in-flight analysis ties up one server thread/coroutine and one open client connection for its entire duration. Under concurrent load, this degrades latency for all other API consumers sharing the server.

**The question is not whether the work is slow today, but whether it can be guaranteed to be fast.** Even if median latency is acceptable, P99 behaviour is not. An architecture that works for the median but fails at P99 under production load is not a production architecture.

---

## Decision

We classify geospatial analysis requests as **Long-Running Operations (LROs)**: work whose completion timeline cannot be contractually guaranteed to fit within a single synchronous HTTP exchange.

This classification is a system boundary decision, not a performance optimisation. It applies regardless of how fast the upstream service happens to be on any given day. The classification holds as long as:

1. Duration is determined by an external dependency we do not control.
2. Duration has no reliable upper bound below standard HTTP timeout thresholds.
3. Failure of the execution path is meaningfully different from failure of the acceptance path.

---

## Consequences

### Reliability

Separating *accepting a request* from *executing it* means each can fail independently with a clear, observable outcome. If execution fails, the job record captures a `failed` status the client can discover. If acceptance fails, nothing was started — the client can safely retry. This eliminates the ambiguous timeout failure mode entirely.

**Tradeoff:** The execution environment must be durable. A `running` job that is interrupted by a process crash (SIGKILL, OOM) will remain stuck in `running` indefinitely without a watchdog or timeout policy. The LRO classification makes this class of failure *visible* but does not resolve it — a dead-job cleanup mechanism becomes a required operational concern that did not exist under a synchronous model.

### Durability

A job record in persistent storage is a first-class artefact with a lifecycle. It survives server restarts, deployment events, and client disconnects. The result of analysis work exists independent of the connection that requested it. This is a fundamental shift from "the result lives in the response" to "the result lives in the database."

**Tradeoff:** The job table is now infrastructure that must be operated: migrated, indexed, and eventually purged. Stale job records accumulate. The cost of this is low but real — it is operational surface area that did not exist in a synchronous design.

### Developer Experience (DX)

The client receives an immediate, deterministic acknowledgement with a job identifier and a status URL. It always knows the job was accepted. Polling with a `Retry-After` hint gives the client a reliable rhythm for checking progress. A terminal `completed` or `failed` state tells the client unambiguously what happened.

Under a synchronous model, DX degrades with latency: a slow response feels like a hang, a timeout is indistinguishable from a server error, and partial failure leaves the client with no recoverable state. The LRO model decouples "did you receive my request" from "is the work done," which are different questions deserving different answers.

**Tradeoff:** The client is now responsible for implementing a poll loop. This is more code than a single `await fetch(...)`. A client that forgets to poll, or polls incorrectly, will never retrieve its results. The client contract is more complex; the failure surface moves from server-side to client-side.

### Scalability

The server accepts work at one rate and executes it at a rate determined by the backend. These two rates can be tuned independently. The server never holds open connections for the duration of analysis, which means connection pool exhaustion under concurrent load is not an inherent property of the design.

**Tradeoff:** If many clients submit jobs simultaneously, accepted jobs may queue and sit in `pending` state for longer than expected. The client sees "pending" but the server is not necessarily making progress — it may be waiting for upstream capacity. The LRO pattern does not itself solve backpressure; it only makes queuing observable rather than hidden behind a slow response.

---

## Alternatives Considered

**Synchronous HTTP with extended timeouts.** Configure the API gateway to allow 60–120 second timeouts. This is operationally simple but transfers the reliability problem to the infrastructure layer. P99 latency for all API consumers rises. Timeout ambiguity remains. Rejected because it optimises for the happy path while worsening the failure path.

**Synchronous HTTP with a strict fast-fail SLA.** Enforce that the upstream analytics service must respond in under N seconds or we return an error. This requires an SLA we cannot enforce on a third-party service and shifts the problem from "slow" to "frequently failing." Rejected because it trades a latency problem for a reliability problem.

**Do nothing (accept only fast datasets).** Restrict the API to datasets that are known to be fast. This artificially constrains the product and does not survive new datasets or increased data volume. Rejected because it bounds the feature set to current performance characteristics rather than designing for the realistic envelope.

---

> **Summary:** We classify geospatial analysis as an LRO because duration is externally determined, unguaranteed, and architecturally incompatible with synchronous HTTP at P99. This is the least-worst framing: it introduces client complexity and operational overhead in exchange for eliminating timeout ambiguity, enabling observable failure, and decoupling acceptance from execution. The synchronous alternatives are simpler today and unreliable in production.

----

# ADR-002: Use Request/Acknowledge/Poll for LRO Client Communication

**Status:** Accepted  
**Date:** 2026-06-10  
**Deciders:** Engineering team  
**Supersedes:** —  
**Related:** ADR-001 (LRO classification)

---

## Context

ADR-001 establishes that geospatial analysis requests are Long-Running Operations. That decision creates a new problem: *how does a client learn that the work is done and retrieve the result?*

The synchronous model has no answer — the response *is* the result. Once we decouple acceptance from execution, we need a mechanism to reunite them. Several mechanisms exist. The choice among them has direct consequences for reliability, durability, developer experience, and scalability that outlast any particular implementation.

The candidate patterns, at the protocol level, are:

| Pattern | Description |
|---|---|
| **Request/Acknowledge/Poll** | Server returns a job token immediately; client polls a status endpoint until terminal state |
| **Server-Sent Events (SSE)** | Server pushes progress and final result over a persistent HTTP stream |
| **WebSocket** | Bidirectional persistent connection; server pushes updates |
| **Webhook / Callback** | Client registers a URL; server POSTs to it on completion |
| **Long-Poll** | Client holds a request open; server responds only when work completes |

---

## Decision

We use **Request/Acknowledge/Poll**: the submission endpoint returns a job identifier and status URL immediately (HTTP 202); the client polls a separate status endpoint at a client-controlled cadence guided by a `Retry-After` response header; when the job reaches a terminal state (`completed` or `failed`), the response includes resource URLs pointing to the results.

The pattern has three invariants that must hold regardless of implementation:

1. **The submission response contains everything the client needs to poll.** No out-of-band communication, no session state, no shared memory.
2. **The status endpoint is idempotent and stateless.** Any client, on any connection, with only the job ID, can retrieve current status at any time.
3. **Terminal states are permanent.** A `completed` or `failed` job does not revert. The result is retrievable for as long as the record exists.

---

## Consequences

### Reliability

**Survives every failure the synchronous model cannot.** Client disconnects, network interruptions, browser tab closes, and server restarts between submission and completion are all non-events. The job record is the source of truth; the client reconnects and polls at any time.

**Failure is observable.** A `failed` terminal state with a message is categorically different from a timeout. The client knows the work was attempted, it failed, and the specific job that failed. This makes retry decisions explicit — the client can submit a new job knowing the previous one definitively did not complete.

**Tradeoff — the "stuck running" problem.** A job in `running` state at the moment of a hard process termination (SIGKILL, OOM kill, host failure) cannot update its own status. The record stays `running` forever without an external watchdog. This failure mode does not exist in the synchronous model because there is no persistent state to become inconsistent. The poll pattern makes this class of failure *observable* (the job is visibly stuck) but *does not resolve it* — a timeout and cleanup policy for stale running jobs is a required complement to this ADR that is not addressed here.

### Durability

The job record and its associated resources are written to durable storage before the client is notified of completion. A client that receives `completed` with resource URLs can fetch those resources at any future time — minutes, hours, or days later — without re-running the analysis. Results are an artefact of the system, not a transient response.

**Tradeoff — storage is not free.** Job records, chart data, and insight resources accumulate. A retention and purge policy is a consequence of this design. Without one, the database grows without bound. This is operational overhead with no equivalent in a synchronous model.

### Developer Experience (DX)

**Immediate, deterministic feedback.** The submission response arrives in milliseconds regardless of analysis duration. The client always knows its request was received and accepted. The job ID is a handle the client can log, display, bookmark, and share.

**Discoverable progress.** `Retry-After: 1` in the status response is a contract: "come back in this many seconds." The client does not need to guess a polling interval or implement exponential backoff for the happy path. Progress visibility (pending → running → completed) lets the frontend communicate accurately to the user rather than showing an indefinite spinner.

**Error recovery is explicit.** When a job fails, the status endpoint returns `failed` with a reason. The user can be told *what* failed and *why*, not just "something went wrong." This is categorically better DX than an opaque timeout error.

**Tradeoff — poll loop is client complexity the synchronous model does not require.** Every client of this API must implement: submit → receive job ID → loop until terminal → follow resource URL. This is three to four network interactions instead of one. A frontend developer accustomed to REST request/response must understand the asynchronous model. Incomplete or incorrect poll implementations (forgetting to poll, not handling `failed`, not following resource URLs) are a predictable class of integration bug that does not exist in the synchronous model.

**Tradeoff — polling granularity is coarser than real-time.** A job that completes in 300ms will not be visible to the client until the next poll at ~1s. The `Retry-After` hint reduces but does not eliminate this latency. For use cases where sub-second result delivery matters, this pattern is the wrong choice.

### Scalability

**Acceptance and execution are independently scalable.** The submission endpoint is lightweight — it writes one row and enqueues a background task. It can handle a high acceptance rate without regard for how fast analysis runs. Execution throughput is a separate concern governed by backend capacity, worker pools, and upstream API rate limits.

**No connection held for work duration.** The server does not hold an open connection per in-flight analysis. Under concurrent load, the number of open connections is proportional to current polling activity, not to the number of analyses in progress. These are different and typically much smaller numbers.

**Tradeoff — polling traffic under load.** N clients each polling once per second is N additional requests per second to the status endpoint. For large user populations with many concurrent jobs, polling overhead becomes measurable. Mitigation strategies (longer `Retry-After` intervals, WebSocket push for high-frequency users) are valid as scale demands them, but they add complexity this pattern defers.

---

## Alternatives Considered

**Server-Sent Events (SSE).** The server pushes progress updates and the final result over a single long-lived HTTP stream. Better result latency than polling; no client poll loop. Rejected because: (1) the connection is held for the duration, reintroducing the server resource problem for concurrent jobs; (2) SSE does not survive client reconnects gracefully without implementing a `Last-Event-ID` recovery protocol, which approaches the complexity of the poll pattern anyway; (3) SSE is not uniformly supported across HTTP infrastructure (proxies, CDNs, API gateways often buffer or drop streams).

**WebSocket.** Bidirectional; the server pushes completion events. Better latency, richer protocol. Rejected because: (1) WebSocket requires a stateful connection the server must maintain per client; (2) it does not compose cleanly with REST semantics, CDN caching, or API gateway infrastructure; (3) the added complexity is not warranted when the only server-to-client message is "job done, here is the URL." WebSocket is the right answer when interaction is high-frequency or bidirectional; analysis result delivery is neither.

**Webhook / Callback.** Client registers a URL at submission; server POSTs to it on completion. Eliminates polling entirely; low server overhead. Rejected for the current context because: (1) browser-based clients cannot expose public callback URLs without additional infrastructure (relay service, hosted endpoint); (2) webhook delivery is not guaranteed — the server must implement retry logic for failed deliveries, which is more complex than a status endpoint; (3) security surface area (callback URL validation, authentication) is non-trivial. This pattern is the right answer for server-to-server integrations and may be appropriate as an *additional* delivery mechanism in future.

**Long-Poll.** Client holds a request open; server responds only when the job completes. Simpler client than polling. Rejected because: (1) it holds server connections for the full analysis duration, identical to the synchronous problem; (2) timeout handling is indistinguishable from completion; (3) it cannot carry progress updates — only a terminal response.

---

## What This Pattern Does Not Decide

- **How the background work is executed.** In-process `BackgroundTasks`, a separate worker process, a message queue, a Dask cluster — any of these is compatible with this pattern. The pattern specifies the client-facing contract, not the execution infrastructure.
- **Polling interval policy.** `Retry-After: 1` is a starting point. The right value depends on median job duration, user tolerance, and server load. This is tunable without changing the pattern.
- **Result retention policy.** How long completed job records and their resources are retained is a product and operational decision not addressed here.
- **Push notification as a supplement.** Nothing in this decision precludes adding an SSE or WebSocket push channel *alongside* polling for clients that want lower latency. Poll must remain authoritative; push can be additive.

---

> **Summary:** Request/Acknowledge/Poll is chosen because it is stateless, survives all client and server failures, composes cleanly with HTTP infrastructure, and makes both progress and failure explicitly observable. The costs are real: client poll loop complexity, coarser result latency than push mechanisms, polling overhead at scale, and a stuck-running failure mode that requires a watchdog. These costs are lower than the costs of the alternatives for the current use case and team context. Revisit if: result latency becomes a product requirement measured in milliseconds, or if server-to-server webhook delivery becomes a first-class integration need.

gtempus (changes requested):
See review [here](https://github.com/wri/project-zeno/pull/704#issuecomment-4670768348), @jterry64 

Nice work! 💯

gtempus (approved):
This is a nice-looking endpoint with favorable DX outcomes, @jterry64! 💪 
It was fun working on this with you in real-time!

gtempus:
For followup at a later sprint, here's the updated review with items to address later:

# PR Review: `analyze_endpoint` — Async Analysis LRO Pattern

---

## Overall Grade: **A-**

This PR delivers a production-quality Long-Running Operation pattern for geospatial analysis. The hexagonal architecture is properly implemented end-to-end: domain objects live in the right layer, ports are complete, adapters honour the contracts, and FastAPI's dependency injection is used correctly throughout. Authorization, input validation, structured logging, and test quality are all at a high standard. What remains is a handful of medium-priority design debts — none of which block correctness today, but each of which will make the feature harder to extend.

---

## What's Working Well 🟢

### 1. Hexagonal Architecture Is Correctly Implemented

The port/adapter split is clean and complete:

```
services/job.py       ← JobRepository (ABC port)  +  JobData / JobResourceData (domain objects)
services/charts.py    ← ChartGenerator (ABC port)
repositories/         ← DBJobRepository (adapter)  +  get_job_repository() factory
```

All four methods (`create_job`, `update_job_status`, `create_insight_resource`, `get_job`) are declared on the abstract `JobRepository`. Domain objects (`JobData`, `JobResourceData`) are defined in the service/domain layer, not in the repository adapter. This is the correct dependency direction: the adapter imports from the domain, not the other way around.

`AnalysisJobRunner` and `AnalyzeService` accept only interfaces — both are fully testable without touching a database.

---

### 2. FastAPI DI Is Idiomatic

Both routers inject the repository through the proper factory:

```python
# repositories/job_repository.py
def get_job_repository() -> JobRepository:
    return DBJobRepository()

# analyze.py / jobs.py
repo: JobRepository = Depends(get_job_repository)
```

Because both the handler and `get_analysis_runner` reference the same `Depends(get_job_repository)` function, FastAPI correctly deduplicates the dependency within a single request — only one `DBJobRepository` instance is created per request cycle. This is an idiomatic and correct pattern.

---

### 3. Authorization Is Correct (OWASP IDOR Prevention)

```python
# jobs.py
job = await repo.get_job(job_id)
if not job or job.user_id != user.id:
    raise HTTPException(status_code=404, detail="Job not found")
```

Returning `404` (not `403`) for a job that belongs to another user is exactly right — it prevents enumeration of valid job UUIDs by unauthorized callers (OWASP A01 Broken Access Control). `JobData.user_id` is present in the domain object and properly propagated from the ORM.

---

### 4. LRO Pattern Is Textbook Correct

`POST /api/analyze → 202/pending → GET /api/jobs/{id} (poll) → follow resource_url` is the standard async HTTP pattern. The `Retry-After: 1` header on `pending`/`running` responses is a developer-experience detail that many teams skip. The Mermaid sequence diagram in `src/api/docs/analyze_flow.md` is excellent in-repo documentation that will save future engineers hours of onboarding time.

---

### 5. Input Validation Is Solid

`AnalyzeRequest` covers all the important cases:

```python
class AnalyzeRequest(BaseModel):
    aois: List[AreaOfInterest] = Field(min_length=1, ...)   # rejects empty list → 422
    dataset_id: int = Field(...)
    start_date: date = Field(...)                            # rejects "banana" → 422
    end_date: date = Field(...)
    thread_id: Optional[str] = Field(default=None, ...)
```

Using `datetime.date` means the OpenAPI spec advertises `"format": "date"` to consumers, and malformed inputs are rejected at the framework boundary before reaching the service layer. The `min_length=1` guard on `aois` is correct — an empty list would silently produce an empty result without it.

---

### 6. Database Correctness Is Complete

- `InsightOrm.thread_id` is `nullable=True` — `None` is stored as `NULL`, not coerced to an empty string.
- Migration `c4d5e6f7a8b9` adds a `BEFORE UPDATE` trigger on `jobs` so `updated_at` is maintained correctly for direct SQL updates, not only through the ORM.

---

### 7. Test Quality Is High

The `FakeHandler` / `FakeChartGenerator` / `FakeJobRepository` fakes are defined inline, minimal, and prove every port is honoured by a real substitute. State-machine transitions (`pending → running → completed/failed`), chart axis correctness, generator routing, and `test_drops_rows_where_area_ha_is_zero` (a domain-rule test) are all covered. No mocks, no DB, no network — this is the correct way to test hexagonal code.

---

### 8. Structured Logging Throughout

```python
logger.info("analysis_job_started", job_id=..., user_id=..., dataset_id=..., ...)
logger.info("analysis_job_completed", job_id=..., duration_ms=..., ...)
logger.error("analysis_job_failed", severity="high", ...)
```

`duration_ms` on completion enables p95/p99 latency tracking with no additional instrumentation. All lifecycle transitions emit structured events with the right context.

---

## Issues to Address 🟡

### 🟡 Module-Level Singletons in the Router

```python
# analyze.py — lines 22-23
handler = AnalyticsHandler()
generators = [TCLChartGenerator(TREE_COVER_LOSS_ID)]
```

These concrete infrastructure objects are instantiated at import time. This means:
- Tests that import the router module pay the construction cost of `AnalyticsHandler` unconditionally.
- Swapping `AnalyticsHandler` in tests requires `monkeypatch` or module-level patching rather than constructor injection.
- `get_analysis_runner` silently closes over `handler` and `generators` from module scope — this dependency is invisible in its signature.

**Preferred approach** — move construction to application lifespan state:

```python
# app.py
@asynccontextmanager
async def lifespan(app: FastAPI):
    app.state.analytics_handler = AnalyticsHandler()
    app.state.chart_generators = [TCLChartGenerator(TREE_COVER_LOSS_ID)]
    yield

# analyze.py
def get_analysis_runner(
    request: Request,
    repo: JobRepository = Depends(get_job_repository),
) -> AnalysisJobRunner:
    return AnalysisJobRunner(
        service=AnalyzeService(
            request.app.state.analytics_handler,
            request.app.state.chart_generators,
        ),
        repo=repo,
    )
```

---

### 🟡 Agent Domain Leaking Into the API Service Layer

```python
# services/analyze.py — line 32
result = await self._handler.pull_data(
    query="",           # ← meaningless in a direct API call
    ...
)
```

`query` is a concept from the conversational agent interface. Passing `query=""` is a shim that works today, but as `AnalyzeService` grows, other agent-specific parameters will accumulate alongside the empty sentinel.

**Options:**
1. Add `pull_data_direct()` to `DataSourceHandler` without the `query` parameter.
2. Create a thin `AnalyticsPort` adapter specifically for the API domain that wraps `AnalyticsHandler` and hides `query=""` internally — keeping the API service clean.

Option 2 is the hexagonally correct solution. Option 1 is acceptable as a short-term fix.

---

### 🟡 Charts Typed as `list[dict]` Throughout

The chain `ChartGenerator.generate() → list[dict] → create_insight_resource(charts: list[dict]) → InsightChartOrm` relies entirely on string-key access (`"xAxis"`, `"yAxis"`, `"type"`, etc.) with no compile-time safety:

```python
# job_repository.py — silent if "xAxis" is misspelled
x_axis=chart.get("xAxis", ""),
```

When a second `ChartGenerator` is added, a misnamed key will silently produce a blank chart field. Introduce a `ChartSpec` dataclass in `services/charts.py`:

```python
@dataclass
class ChartSpec:
    id: str
    title: str
    type: str
    x_axis: str
    y_axis: str
    data: list[dict]
    insight: str = ""
    color_field: str = ""
    stack_field: str = ""
    group_field: str = ""
    series_fields: list[str] = field(default_factory=list)
```

Update `ChartGenerator.generate()` to return `list[ChartSpec]`, and `create_insight_resource` to accept `list[ChartSpec]`.

---

### 🟡 Background Task Durability Gap

```python
# analyze.py
job_id = await repo.create_job(...)        # persisted to DB
background_tasks.add_task(_run_job, ...)   # in-process only
```

If the FastAPI worker process is killed (OOM, `SIGKILL`, unexpected crash) between these two lines, the job is created with `status=pending` and will never progress. For the current ECS deployment with graceful shutdown this window is small and acceptable for MVP. When this feature handles production traffic, consider:
- A DB-backed job queue (`procrastinate`, `pg_boss`) that survives process restarts, or
- A periodic task that alerts on jobs stuck in `pending`/`running` beyond a timeout threshold.

Document the known limitation in the module docstring so the next engineer understands the trade-off explicitly.

---

### 🟡 `created_at` in the 202 Response Is from the Application Clock, Not the DB

```python
# analyze.py
return JobResponse(
    id=job_id,
    ...
    created_at=datetime.now(),   # ← application server clock
)
```

The `created_at` in the immediate `202` response comes from `datetime.now()` in the router, not from the database `server_default=now()`. If application server and DB server clocks differ (common in containerised environments), the `created_at` on the `202` will not match the `created_at` on subsequent `GET /api/jobs/{id}` responses.

**Fix:** Return `created_at` from `repo.create_job()`:

```python
async def create_job(...) -> tuple[UUID, datetime]: ...

job_id, created_at = await repo.create_job(...)
return JobResponse(..., created_at=created_at)
```

---

### 🟡 `ResourceStatus` Cannot Represent Partial Failure

```python
class ResourceStatus(str, Enum):
    COMPLETED = "completed"
```

There is no `FAILED` value. If charts succeed for one AOI but fail for another in a future multi-AOI scenario, individual `JobResource` entries cannot reflect that failure state. Add it now while the migration cost is low:

```python
class ResourceStatus(str, Enum):
    COMPLETED = "completed"
    FAILED = "failed"
```

---

## Nits 🔵

- **`_run_job` is a module-level function.** It cannot be individually tested or replaced without patching the module. Converting it to a method on `AnalysisJobRunner` (e.g., `runner.run_as_background_task(...)`) would keep all logic within the tested service boundary.

- **Migration revision `b3c1d2e4f5a6` is hand-crafted.** Alembic generates random IDs to avoid branch collisions in long-lived projects. Use `alembic revision --autogenerate -m "..."` and retain the generated ID.

- **`AnalyzeResult.charts: Optional[Any]`** disables type checking for chart data. Should be `Optional[list[dict]]` at minimum, `Optional[list[ChartSpec]]` once `ChartSpec` is introduced.

- **`_column_to_rows` in `charts.py`** is generic enough for `shared/` if reused. Fine here for now; worth a comment noting its reuse potential.

---

## Summary Table

| Area | Assessment |
|---|---|
| LRO design (`POST → poll → follow`) | ✅ Correct and well-documented |
| Port/adapter (hexagonal) structure | ✅ Complete and correctly layered |
| Authorization — IDOR check on `GET /api/jobs/{id}` | ✅ Present, 404 not 403 |
| FastAPI DI (`Depends`) | ✅ Idiomatic, dependency deduplication correct |
| Input validation (`date`, `min_length`) | ✅ Solid |
| Domain objects in correct layer | ✅ In `services/job.py`, not repository |
| `thread_id` null handling | ✅ Nullable column, no empty-string coercion |
| `updated_at` DB trigger | ✅ Added in migration `c4d5e6f7a8b9` |
| Structured logging | ✅ Comprehensive, `duration_ms` included |
| Unit tests | ✅ Clean fakes, state-machine + domain rule coverage |
| Module-level singletons | ⚠️ Coupled at import time, invisible to DI |
| Agent `query=""` domain leak | ⚠️ Present in `AnalyzeService` |
| Chart type safety (`list[dict]`) | ⚠️ No `ChartSpec`; silent key mismatches possible |
| Background task durability | ⚠️ In-process only; stuck `pending` on crash |
| `created_at` clock drift | ⚠️ Router clock vs. DB clock on 202 response |
| `ResourceStatus` completeness | ⚠️ No `FAILED` value for partial failures |

---

> The architecture direction is sound and the implementation quality is high. The remaining issues are all medium-priority design debts — none affect correctness in the current happy path. The highest-value follow-ons are the module-level singleton refactor (improves testability), the `ChartSpec` type (prevents silent chart regressions when a second generator ships), and either mitigating or explicitly documenting the background task durability gap before this feature reaches production traffic.

PR #705, opened 2026-06-08
PZB-909 PZB-622 Have pick_aoi LLM specify specific aoi_type, if clear from user query 

danscales:
PZB-622 Have pick_aoi LLM specify specific aoi_type, if clear from user query

Added a new AOIType enum and added aoi_type as part of the PlaceQuery that the LLM will call to the pick_aoi tool. It will specify a specific aoi_type if that is clear from the query, else specify None, which means to do the usual search through all aoi_types. By searching through only one specific AOI table when it is clear from the user query, we fix some mistakes where it can't find the most obvious AOI, because of matches with non-relevant aoi types.

Fixes gold case in PZB-622:
"Which US state lost more forest due to wildfires? Washington or Oregon? Use the latest available data" (previously, it would usually choose a Washington park, rather than Washington state, and fail.)

Also fixes PZB-575: "show me tree cover loss in Cobaipo protected area in Bolivia" (previouly, it would get confused and choose all of Bolivia).

Also fixes a Gold case from Solomon:
"In the Kurtjar People territory in Australia, what percentage of land is non-natural?" (previously, it would get confused and usually choose Queensland).

Also now does the correct thing for:
"Go to Navajo territory"
"Go to Navajo nation"
(previously it would always find Navajo National Monument, but now correctly chooses the Navajo Nation AOI in Landmark)


src/agent/subagents/pick_aoi/tool.py
```diff
@@ -205,6 +207,15 @@ async def query_aoi_database(
...
         union_parts = []
 
+        if aoi_type is not None:
+            if aoi_type in existing_tables:
+                existing_tables = [aoi_type]
+            else:
+                logger.error(
+                    f"Required geometry table {aoi_type} does not exist in the database"
+                )
+                return pd.DataFrame()
```
yellowcap: This should never happen, so I think this should either 

1. Raise an error
2. do the "full search", i.e. just keep going as if `aoi_type` is none. See suggestion below.


I would suggest option 2, it will fall back to the past behavior.

```suggestion
```

yellowcap (approved):
I like the idea, it will make the lookups slightly faster and potentially less ambiguous. Lets make sure this passes the tools tests and does not degrade the gold set.

src/agent/subagents/pick_aoi/tool.py
jterry64: I'm assuming this is more when you run locally and are missing some tables? So I agree with raise an error, unless this can happen in production. This way you know it's because you're locally missing a table.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -574,6 +585,13 @@ async def check_duplicate_aois(
     "landmark",
 ]
 
+AOIType = Literal[
```
jterry64: I wonder if you'd get improved performance if you made these names more plain english, and then would require less prompting above. Like "gadm" as "administrative area", or even have multiple options like "country"/"state"/etc. like SubregionType above and then just map that to gadm.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -205,6 +207,15 @@ async def query_aoi_database(
         # Build the query based on existing tables
         union_parts = []
 
+        if aoi_type is not None:
+            if aoi_type in existing_tables:
+                existing_tables = [aoi_type]
```
jterry64: I see why you did it this way, and I'm fine with that for now because the better approach would take a fair amount of refactoring, but it's a bit confusing to read. I'd prefer if we instead refactored below so we separated out the code to call search a particular table, and then could just pass which aoi type we want.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -592,6 +610,13 @@ class PlaceQuery(BaseModel):
             "inside the place(s); otherwise leave null."
         ),
     )
+    aoi_type: Optional[AOIType] = Field(
```
jterry64: You could make this a list to deal with cases like where you seeing it call the tool twice, and have it be able to look through multiple tables. That's better than all four, unless it always chooses all four...

jterry64: I was also thinking originally it'd be a parameter on the tool itself, rather than this internal call. This makes sense since there was already a call to extract more info, but I wonder if that's why there was a disconnect between the agent and the tool result: the agent doesn't see this internal call, and seems to know better, probably because it's a bigger model.

jterry64 (approved):
@danscales very nice! I think it's overall good, I had a couple thoughts on design decisions but I don't think any are blockers if you want us to do them later.

jterry64 (changes requested):
@danscales actually while reviewing, I ran the tools tests (you can do that by adding a label on the PR called "run tool tests"), and it looks like a few test_pick_aoi.py are failing now, if you can check those out. Also I think the unit tests should be 100% passing.

src/agent/subagents/pick_aoi/tool.py
danscales: Yes, this should never happen in production, unless things are really messed up.  This is more to catch the case where you are running locally and you didn't load in all the AOI tables.  As suggested, I changed the code to raise an error. Thanks!

src/agent/subagents/pick_aoi/tool.py
danscales: When the LLM calls twice, it is not because of multiple AOI types. In all cases I've seen, it is choosing the same aoi_type on the 2nd try. As we discussed, we're just assuming that it is just using its general knowledge, and deciding it didn't like the first answer.

With respect to where the aoi_type parameter should be, I actually added it as a value that the pick_aoi subagent should figure out.  (See the SMALL_MODEL.with_structured_output(PlaceQuery) call around line 680, where PlaceQuery now requires choosing an aoi_type.)  I guess you may be proposing that instead I ask the top-level LLM to decide on an aoi_type.  That's seems OK,  although all the other agents/tools mostly take just the query.  Also, it seems less encapsulated to have knowledge about the AOI types all the way at the top agent's prompt.

But I'll try that out, and see how that compares. It's a difference of whether the top-level agent or the pick_aoi subagent decide on any relevant aoi_type.

danscales:
> @danscales actually while reviewing, I ran the tools tests (you can do that by adding a label on the PR called "run tool tests"), and it looks like a few test_pick_aoi.py are failing now, if you can check those out. Also I think the unit tests should be 100% passing.

Yes, I hadn't yet fixed up the tests to deal with the extra argument to query_aoi_database().  All tests are now passing, except for two that I think are also broken in main:

test_pick_dataset_reason_matches_query_language_with_llm_judge (spanish and portuguese cases)

src/agent/subagents/pick_aoi/tool.py
danscales: OK, with my latest commit, I changed it over to the way you suggest above, where the top-level agent is asked to choose the area_of_interest_type, if it clear from the user query. I changed the AOI type Literal to use full phrases (not abbreviations), as you suggested, to allow the LLM to understand the choice better without adding much to the prompt.  The new way is performing the same as the previous way (where the sub-LLM made the choice on AOI type) on the four cases that this change is fixing.  I have added those four cases (two of which are disabled gold set cases) in tests/eval/datasets/pzb-909.csv

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -39,6 +39,20 @@
 SUBREGION_LIMIT_ADMIN = 1000
 SUBREGION_LIMIT = 50
 
+Area_Of_Interest_Type = Literal[
```
jterry64: I don't think the underscores are standard Python, should just be camel case?

```suggestion
AreaOfInterestType = Literal[
```

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -39,6 +39,20 @@
...
+    "adminstrative area (country, state/region, country/subregion)",
+    "protected area, park, or reserve",
+    "indigenous region or territory",
+    "key biodiversity area",
+]
+
+aoi_to_table = {
```
jterry64, line 52: I think there's a simpler way to do this, like just make these an enum where these are the types and the strings above are the values? But I'm not sure

jterry64 (approved):
This looks good now! Just small style nitpicks. Did this change how it performed or was it roughly the same either way?

src/agent/subagents/pick_aoi/tool.py
danscales: I used underscores to make it more legible to the LLM, but meant to try no underscores as well. With no underscores, the LLM does just as well, so I changed it as you suggested.  Thanks for the suggestion!

src/agent/subagents/pick_aoi/tool.py
danscales: OK, changed to using a StrEnum, which has the right properties (the enum is actually a string, which is what we need).

PR #706, opened 2026-06-09
Implement option for context layers to be vector layers

yellowcap:
This updates the png endpoints for IFL to vector layers (pbf).

For the frontend to know when a context layer is a vector layer, the `ContextLayer` class now has a new field called `source_layer` that needs to be set to the name of the layer in the protobuffer response. [edit: before it was `vector`]

This is the backend implementation of https://github.com/wri/project-zeno-next/pull/516

### DIFF
--- a/src/agent/datasets/catalog/tree_cover_loss.yml
+++ b/src/agent/datasets/catalog/tree_cover_loss.yml
@@ -8,7 +8,8 @@ context_layers:
   start_date: '2002-01-01'
 - value: intact_forest
   description: Tree cover loss within intact forest landscapes (IFLs) — large, undisturbed natural forest expanses free of significant human activity. Use this layer when the user explicitly asks for intact forest loss, or as the default context layer for deforestation/forest loss requests when primary_forest is not available for the AOI.
-  tile_url: https://tiles.globalforestwatch.org/ifl_intact_forest_landscapes/v2025/default/{z}/{x}/{y}.png
+  tile_url: https://tiles.globalforestwatch.org/ifl_intact_forest_landscapes/v2021/default/{z}/{x}/{y}.pbf
+  source_layer: vector
   start_date: '2000-01-01'
 parameters:
 - name: canopy_cover
--- a/src/agent/datasets/catalog/tree_cover_loss.yml
+++ b/src/agent/datasets/catalog/tree_cover_loss.yml
@@ -9,7 +9,7 @@ context_layers:
 - value: intact_forest
   description: Tree cover loss within intact forest landscapes (IFLs) — large, undisturbed natural forest expanses free of significant human activity. Use this layer when the user explicitly asks for intact forest loss, or as the default context layer for deforestation/forest loss requests when primary_forest is not available for the AOI.
   tile_url: https://tiles.globalforestwatch.org/ifl_intact_forest_landscapes/v2021/default/{z}/{x}/{y}.pbf
-  source_layer: vector
+  source_layer: ifl_intact_forest_landscapes
   start_date: '2000-01-01'
 parameters:
 - name: canopy_cover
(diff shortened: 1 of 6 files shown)


src/agent/datasets/catalog/tree_cover_loss.yml
```diff
@@ -8,7 +8,8 @@ context_layers:
   start_date: '2002-01-01'
 - value: intact_forest
   description: Tree cover loss within intact forest landscapes (IFLs) — large, undisturbed natural forest expanses free of significant human activity. Use this layer when the user explicitly asks for intact forest loss, or as the default context layer for deforestation/forest loss requests when primary_forest is not available for the AOI.
-  tile_url: https://tiles.globalforestwatch.org/ifl_intact_forest_landscapes/v2025/default/{z}/{x}/{y}.png
+  tile_url: https://tiles.globalforestwatch.org/ifl_intact_forest_landscapes/v2021/default/{z}/{x}/{y}.pbf
```
yellowcap, line 11: @01painadam the previous layer is for 2025, but that does not work for the pbf layer. Is that expected?

kamicut:
Hey @yellowcap, for each vector tile endpoint, the `source_layer` can be different. For the intact forest vector tiles, the source layer name is `ifl_intact_forest_landscapes`. We should send this mapping to the frontend so that we know which vector layer needs to be rendered.

I made this quick tool to debug the vector tile endpoint and "discover" the source layers from a `.pbf` endpoint https://vt-viewer.exe.xyz/?url=https%3A%2F%2Ftiles.globalforestwatch.org%2Fifl_intact_forest_landscapes%2Fv2021%2Fdefault%2F%7Bz%7D%2F%7Bx%7D%2F%7By%7D.pbf

src/agent/datasets/catalog/tree_cover_loss.yml
01painadam: Yeah It's an older layer - @jterry64 is there a new version?

yellowcap:
ok that makes sense! thanks for clarifying @kamicut . I just pushed a change accordingly. Running tools tests now to see if we broke anything on the dataset picking side.

src/agent/datasets/catalog/tree_cover_loss.yml
jterry64: It's a bit confusing, the new one we ended up actually using a raster layer because there were some issues with the vector. But updates to IFL are always just the loss of intact forest extent. So the 2000 extent is the same for all of them, and it gets updated with how much extent was lost since then. So what we have currently is fine, or we can switch to this and only show 2000. Does that make sense?

src/agent/subagents/pick_dataset/tool.py
```diff
@@ -460,6 +460,7 @@ def get_tile_services_for_dataset(
         context_layer = ContextLayer(
             name=selected_context_layer.get("value"),
             tile_url=selected_context_layer.get("tile_url"),
+            source_layer=selected_context_layer.get("source_layer", None),
```
jterry64, line 463: If looks like you're setting source_layer to "ifl_intact_forest_landscapes" from the YAML, not "vector"?

jterry64 (changes requested):

src/agent/subagents/pick_dataset/tool.py
yellowcap: @jterry64 I updated the PR description, this value does not have to be `vector`, I misunderstood what this was for (see @kamicut's comment above). This is the name of the layer to be displayed as defined internally in the pbf files. My bad for not updating the PR description.

src/agent/datasets/catalog/tree_cover_loss.yml
yellowcap: I am even more confused now. The date in the url is 2021 not 2000. And I also thought that this is the main reason why we wanted vector layers in the first place. Do we have another use case for vector layers as context layers?

yellowcap:
I think we solved this already, closing, lets reopen if needed.

PR #707, opened 2026-06-11
 Add WRI Insights blog search

srmsoumya:
This adds the ability for the agent to search WRI Insights blog articles and answer questions with cited sources from [wri.org/insights](https://www.wri.org/insights)

What's in it
- Article store - fetches WRI Insights articles from the sitemap into `data/wri_insights/` as markdown files, one per article, plus an `index.json` of metadata. Run with `scripts/fetch_wri_insights.py`. Each paragraph is tagged with a citation anchor so answers can link back to the exact spot.
- Search over the filesystem - instead of a vector DB, the corpus is just a folder the agent can read. Two search tools feed it:
    - grep (ripgrep) for exact keyword matches.
    - sgrep, a new semantic search built on a small static embedding model (`potion-retrieval-32M`), for matching by meaning. Index lives on disk as a couple of files.
  - Blog search subagent - an agent that searches, shortlists, reads the promising articles, and writes a concise, cited answer. Exposed to the main agent as the `search_blogs` tool.
  - Skills - explore (turn a vague query from user into concrete analysis suggestions grounded in WRI research) and wri-insights (enrich an analysis with cited WRI context).

This is an experiment in replacing RAG with the filesystem for this corpus.


jterry64 (commented):
@srmsoumya it says the docker repository with the snapshot doesn't exist. Is it in a different region?

jterry64:
@srmsoumya if you don't mind, I added the eval set file pzb-919.csv so we at least have something to start with. It contains 20 evals, generated by choosing 5 documents and using an LLM to generate 3 questions about each document. It then does a loose LLM judge check to see if it answers the question correctly and cites the correct article. I then translated 5 of the rows to different languages to check it handles them with the same quality. 

Good news is it does really well! It only fails 2-3, usually because it tries to select an AOI/dataset instead, or they're harder questions requiring specific stats to be pulled out. So I think your small model approach is working fine. It can be pretty slow sometimes, but it's hard to tell how significant that is locally, so once it's on staging we should check how it affects latency. We also can update these evals based on researcher feedback, if they notice specific issues the judge doesn't.

yellowcap:
Lets get the tool registry in first, then add this tool as experimental initally
https://github.com/wri/project-zeno/pull/712

jterry64:
A couple fix suggestions from a manual and AI PR review. Feel free to revert any commits you don't want:

Made the feature completely behind feature flag, including the system prompt changes
I got rid of the blog endpoint in favor of just updating the state, so it's more consistent with how the FE is currently interacting with agent instead of requiring a separate endpoint, although unsure right now if we've thought through how the FE will integrate. So maybe we don't even need that?
Added more test coverage
Added an LRU cache around loading the embeddings, this was caught by AI review as an expensive call per search query
Grep only on search results from sgrep instead of the whole corpus. I noticed the agent called grep a lot. This change cut the latency in half, going from ~1 min per question locally to <30 seconds.
Switch to Gemini, since we don't have Anthropic/Haiku set up and might need to confirm if it's ok to switch to that. This unfortunately did slight reduce the accuracy.

PR #708, opened 2026-06-11
PZB-961: Remove analyze skill line to be proactive, which conflicts with instruction to provide suggestions

jterry64:
This line was confusing the LLM, causing it to call pick_dataset again a second time after receiving a suggestion to be proactive. I ran an eval on the question "How do fires impact nature in Spain?" 10 times before and after the change. Before this change, 40% of the time it both provides nudges and chooses a dataset proactively. After this change, it consistently only provides nudges and waits for user input. 

Also playing around with the idea of adding an eval csv per JIRA item requiring agent changes.


yellowcap (approved):
Out of curiosity, how is the regression csv used?

danscales (commented):
Cool, glad you found those conflicting instructions!

danscales (approved):

jterry64:
> Out of curiosity, how is the regression csv used?

Right now it's just a manual set to run while developing, but we're thinking it might be a good subset to run in CI on commits instead of having to run all evals. And then we can run the full gold set on merge to main. It's not necessarily regression either, you can do more challenging questions it's still flaky on just to see its limits. This also can be good source at the end of the sprint for updating the spreadsheet sets.

PR #710, opened 2026-06-12
PZB-938 feat: tool for dynamic rendering of Sentinel 2 imagery

yellowcap:
## Summary

  Adds a show_imagery agent tool and a /mosaic API that render a continuous Sentinel-2 L2A layer over the current AOI, centered on a target date. Scenes are discovered through the earth-search STAC API and pixels
  are read directly from the public sentinel-cogs AWS bucket — nothing is downloaded, processed, or stored ahead of time. Scoped to regional AOIs (≤ 50,000 km²) by design. All mosaic endpoints, tiles included,
  sit behind the same bearer auth as the rest of the API.

### Agent integration
 show_imagery(target_date, window_days, max_cloud_cover) reads the AOI selection from LangGraph state (same pattern as pull_data), builds a mosaic in-process, and writes an imagery entry into
  agent state — tile_url, tilejson_url, scene count, date span, and the search constraints used — which reaches the frontend through the chat stream and thread replay, mirroring how pick_dataset surfaces dataset
  tile layers. The session-context middleware shows the active imagery layer to the model on every turn.

###  Search defaults and escalation. 
Defaults are strict on purpose: scenes within ±7 days of the target date, under 20% cloud cover. When nothing matches, the tool's feedback message states exactly what was
  searched, and the new show-imagery skill instructs the agent to offer the user the two levers — widen the window (window_days=30/60) or allow cloudier scenes (max_cloud_cover=50/80) — rather than silently
  retrying. Both parameters are clamped server-side and baked into the mosaic token, so a loosened search stays loosened on rebuild.

###  Mosaic construction.
 A STAC search produces a MosaicJSON: per quadkey, an ordered list of COG URLs, capped at 12 scenes. Scenes are sorted by date-proximity to the target (cloud cover as tiebreak) before
  assembly — the mosaic renders the first valid scene per tile, so ordering decides what's displayed; without it, whatever order the STAC API returned would. Search timing, parameters and result counts are
  logged.

###  Recipe tokens instead of storage. 
The mosaic_id used in tile URLs is a compact base64 token encoding the build recipe: AOI references (source/src_id — geometries already live durably in the gadm/kba/wdpa/custom
  tables), the resolved target date, and search parameters. The in-memory TTLCache holding built MosaicJSONs is purely an L1: on a miss (restart, other worker/replica, replayed thread weeks later) an async route
  dependency (ensure_mosaic) decodes the recipe — clamping parameters to the same bounds the create endpoint enforces — and rebuilds the mosaic before the titiler endpoint runs, with a per-token lock so a cold
  map load triggers one rebuild rather than twenty. Token encoding is deterministic, so identical requests — across calls and users — yield byte-identical URLs and warm caches.

###  Auth.
 Tile/tilejson requests carry the standard Authorization bearer header (MapLibre transformRequest on the frontend). Validated tokens are cached per process for 24 h, so a burst of tile requests costs at
  most one upstream auth call. Because creation and serving sit behind the same gate, recipe tokens need no signature: crafting one grants nothing the authenticated create endpoint wouldn't. Custom-area recipes
  still carry the owning user_id, so imagery in a shared or replayed thread renders for any authenticated viewer.

###  Tile serving.
 A MosaicTilerFactory (routers/mosaic.py, paired with services/mosaic.py per the chat router/service convention) serves tiles/tilejson/WMTS backed by the cache. titiler's error mappings are
  registered app-wide (out-of-bounds tile → 404 instead of a logged 500). Tile reads are sequential with lazy early exit — with "first" pixel selection a tile fully covered by the nearest scene reads exactly one
  COG, instead of titiler's default of up to 70 concurrent S3 reads discarded after the first. GDAL is tuned for remote COGs (no sidecar probing, merged ranges, vsicurl caching). Mosaic GET responses carry
  Cache-Control: private, max-age=86400, since a token always maps to the same imagery — panning back or replaying a thread hits the browser cache.

##  Trade-offs

  - Cold-start latency over storage: the first tile request for a non-cached mosaic pays ~1–2 s (geometry fetch + STAC search). Accepted to avoid operating any persistent store.
  - Drift over snapshots: a rebuild months later can render slightly different pixels if scenes were reprocessed, or fail if a custom area was deleted (becomes a 404). Accepted as the cost of recipe-based
  durability.
  - Strict defaults over silent fallbacks: ±7 days / <20% cloud means "no scenes found" is common in cloudy regions (S2 revisits every ~5 days). The agent surfaces the constraint and lets the user choose between
  recency, clarity, and coverage instead of guessing.
  - Sequential reads over parallel: tiles at scene edges (first scene doesn't cover) are slower than concurrent fetching would be; in exchange, the common case does ~10x less S3 traffic.
  - Auth on tiles over open endpoints: requires the frontend to attach the bearer header via transformRequest and rules out header-less clients (QGIS WMTS, bare browser tabs). Accepted because the frontend is the
  only consumer, and it eliminates both an unauthenticated compute surface and the token-signing machinery.
  - Today's date freezes at creation: "show me current imagery" encodes the resolved date, so replays show what the user originally saw rather than silently updating.

 ## Alternatives considered and discarded

  - Postgres JSONB table for mosaics: fits existing alembic/ORM conventions and was the runner-up, but adds a table, migrations, and a retention policy for data that is fully derivable from inputs we already
  store.
  - S3 backend (cogeo-mosaic supports it natively): durable and cheap, but new bucket/IAM/env plumbing, minio for local dev, and no thread/user linkage.
  - Redis: runs in compose but only serves Langfuse; would be a new production infra dependency for the app.
  - External eoapi/pgstac registration: dataset layers already use eoapi, but it would require ingesting earth-search items into that catalog — an infra change outside this repo.
  - Random uuid ids + TTL cache (the initial MVP shape): simplest, but mosaics died on restart, were invisible across workers, and old threads lost their layers.
  - Signed tokens + unauthenticated tiles (an intermediate iteration): itsdangerous-signed recipes let plain map clients fetch tiles without headers while preventing token forgery. Superseded by requiring
  standard bearer auth on the tile path — one less secret to deploy and rotate, one consistent auth story, at the cost of a transformRequest hook in the single consuming frontend.
  - titiler pixel_selection changes for "first pixel only": unnecessary — first is already the default; the waste was read concurrency, fixed in the backend instead.
  - Auto-retry with loosened filters when no scenes match: rejected in favor of asking — cloudy or off-date imagery shown without consent looks like a bug to the user.

##  Breaking change for the frontend

  Tile/tilejson requests now return 401 without the Authorization header — the map client must attach it (MapLibre transformRequest, plus any direct TileJSON fetches). Mosaic URLs minted before this branch's
  final token format will 404; re-requesting imagery in chat produces a working layer. No deployment configuration is required (the earlier MOSAIC_TOKEN_SECRET is gone).


yellowcap:
Lets get the tool registry in first, then add this tool as experimental initally
https://github.com/wri/project-zeno/pull/712

src/api/services/mosaic.py
```diff
@@ -0,0 +1,336 @@
...
+
+from src.shared.geocoding_helpers import get_geometry_data
+from src.shared.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+# GDAL tuning for reading remote COGs (read by rasterio at open time, so
+# deployment env vars take precedence). Without GDAL_DISABLE_READDIR_ON_OPEN
+# every COG open issues extra (404ing) sidecar-file requests against S3.
+_GDAL_ENV_DEFAULTS = {
```
jterry64, line 59: I'm a little concerned about running the tile rending on the same service as the rest of our backend. It's very CPU intensive, especially if there's parallel requests for multiple tiles for larger areas. However, it does require specifically asking the agent right now to get the tiles, so I imagine right now it'd be infrequently used.

jterry64 (changes requested):
I'm not a huge fan of running the CPU-intensive tile rendering on the same instance as the rest of our backend, but since this is pretty restricted right now it's probably fine. But I do think there's an alternative approach that's simple and may be more sustainable:

- Instead of in-memory, save the MosaicJSON to S3
- Use eoAPI to render the tiles by passing the MosaicJSON URL to eoAPI's mosaic endpoint

This way, we can keep tile generating contained to eoAPI, and also the MosaicJSON caching will persist across worker instances instead of being in-memory on one worker. I'm not sure how eoAPI is set up, but then we could also add some CDN caching on eoAPI so it caches tile requests.

src/api/services/mosaic.py
```diff
@@ -0,0 +1,336 @@
...
+    geometry = await _load_geometry(recipe)
+    check_aoi_area(geometry)
+
+    actual_start = recipe.target_date - timedelta(days=recipe.window_days)
+    actual_end = min(
+        recipe.target_date + timedelta(days=recipe.window_days), date.today()
+    )
+
+    def _search() -> list:
+        catalog = pystac_client.Client.open(STAC_URL)
```
jterry64, line 373: One thing we prefer is following hexagonal architecture, which is mostly making sure any external dependencies are fairly separated from business logic. Are you AI generating most of this code? You can tweak your claude.md to have it follow these principles automatically.

jterry64:
> I'm not a huge fan of running the CPU-intensive tile rendering on the same instance as the rest of our backend, but since this is pretty restricted right now it's probably fine. But I do think there's an alternative approach that's simple and may be more sustainable:
> 
> * Instead of in-memory, save the MosaicJSON to S3
> * Use eoAPI to render the tiles by passing the MosaicJSON URL to eoAPI's mosaic endpoint
> 
> This way, we can keep tile generating contained to eoAPI, and also the MosaicJSON caching will persist across worker instances instead of being in-memory on one worker. I'm not sure how eoAPI is set up, but then we could also add some CDN caching on eoAPI so it caches tile requests.

I looked into this, it doesn't come out of the box with eoAPI so might require some infra work. Something to consider moving to when this gets more production usage or becomes a bottleneck.

yellowcap:
Changing this to rely on vanilla titiler and deploy a mosaicjson endpoint in our cluster
https://github.com/wri/project-zeno-deploy/pull/174

yellowcap:
Made three major changes today:

1. Add new "experimental" agent profile for this feature
2. Rely on S3 for mosaic storage
3. Use vanilla mosaicjson, but keep in app for now

This will make switching over to another titiler quite easy, we'll just make the mosaics accessible from the other titiler and it should work as is.

src/agent/agent_config.py
```diff
@@ -99,3 +106,6 @@ def resolve(self, ff: Optional[str] = None) -> AgentConfig:
 # Production registry — register new flag configs here.
 default_registry = AgentConfigRegistry()
 default_registry.register(AgentConfig(DEFAULT_PROFILE, specs=CORE_SPECS))
+default_registry.register(
```
jterry64, line 109: this is fine to just call it experimental, but if you want to test independently from other changes like the blogs, you can also name is "show-imagery".

jterry64 (approved):

src/agent/agent_config.py
yellowcap: I think we can bundle them. we can separate in more profiles if it leads to problems. I see the experimental one as everything together, and we can make sub profiles as needed

yellowcap:
Thanks for the review @jterry64 !! 🚀

PR #712, opened 2026-06-15
Agent tool profiles

yellowcap:
### Why

We want to ship experimental agent tools to opt-in testers without exposing them in production.

This is in preparation for two PRs that want to add new tooling that must be evaluated first:

- https://github.com/wri/project-zeno/pull/707
- https://github.com/wri/project-zeno/pull/710

 
### What

  Introduce a **tool-profile registry**. A `Profile` bundles a tool set with the prompt fragment that documents each tool, so the prompt is generated from the same source
  that binds the tools and can't drift.

  - **Two profiles:** `default` (production) and `experimental` (default + opt-in tools).
  - **Gating:** resolved per conversation from the user's existing `help_test_features` flag — testers get `experimental`, everyone else (incl. anonymous) gets `default`. No
  API-level profile selection; the flag is set via the existing `PATCH /api/auth/profile`.
  - **Profiles are objects:** `Profile` is a frozen dataclass holding the tool-name tuple and exposing `tools()`, `tool_descriptions()` and `skills()`.
  `resolve_profile(user)` returns a `Profile`; `get_profile(name)` does name lookup + fallback. This gives future per-profile parameters (model override, extra prompt block,
  tool cap) a home on the dataclass instead of more `_for_profile(name)` functions. `Profile` holds only tool *names* — `TOOL_REGISTRY` / `all_skills()` stay the single
  source of truth.
  - **Skills fold in too:** each skill declares its required tools via a `requires:` frontmatter field; a skill is only advertised in a profile that binds all of its tools,
  so recipes never reference unbound tools. (Forward-looking: a future `show-imagery` skill requiring `show_imagery` auto-scopes to `experimental`, no manual per-profile
  list.)

  To add an experimental tool later: register a `ToolSpec` in `TOOL_REGISTRY` and add its name to `_EXPERIMENTAL_TOOLS`; promote to `_CORE_TOOLS` to ship to everyone.

  ### Changes

  - `src/agent/tool_profiles.py` (new) — `TOOL_REGISTRY`, `Profile` dataclass, `PROFILES`, `resolve_profile`, `get_profile`.
  - `src/agent/graph.py` — resolves a `Profile` and derives bound tools + prompt sections from it (replaces the static `tools` list and hardcoded prompt sections).
  - `src/agent/skills/loader.py` + `analyze.md` / `pull-data.md` — `requires` frontmatter parsed into `SkillMeta`.
  - `src/api/routers/chat.py` — pass `help_test_features` through to the agent (was previously dropped).
  - `tests/unit/agent/test_tool_profiles.py` (new) — gating, prompt↔tool consistency, and skill-dependency invariants.

  ### Behavior change

  None today — both profiles currently bind the same core tools (`show_imagery` lands separately), so the `default` prompt renders **byte-identical** to before. This PR is
  the wiring that keeps tools, prompt, and skills consistent once profiles diverge.

  ### Testing

  - `ruff`, `ruff-format`, `mypy` pass (pre-commit).
  - 31 unit tests pass (`test_tool_profiles.py`, `test_skills.py`, `test_cli.py`).
  - Have ran this locally and works fine

### DIFF
--- a/db/alembic/versions/d5e6f7a8b9c0_add_agent_profile_column.py
+++ b/db/alembic/versions/d5e6f7a8b9c0_add_agent_profile_column.py
@@ -1,0 +1,35 @@
+"""add agent_profile column
+
+Revision ID: d5e6f7a8b9c0
+Revises: c4d5e6f7a8b9
+Create Date: 2026-06-15 00:00:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "d5e6f7a8b9c0"
+down_revision: Union[str, None] = "c4d5e6f7a8b9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.add_column(
+        "users",
+        sa.Column(
+            "agent_profile",
+            sa.String(),
+            nullable=False,
+            server_default="default",
+        ),
+    )
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_column("users", "agent_profile")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.6.13"
+version = "2026.6.15"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.6.15.1"
+version = "2026.6.15.2"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/graph.py
+++ b/src/agent/graph.py
@@ -17,23 +17,36 @@ from psycopg_pool import AsyncConnectionPool
 
 from src.agent.llms import FALLBACK_MODELS, MODEL
 from src.agent.middleware import SessionContextMiddleware
-from src.agent.skills import all_skills, read_skill
 from src.agent.state import AgentState
-from src.agent.subagents import generate_insights, pick_aoi, pick_dataset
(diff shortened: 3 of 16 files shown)


batpad:
From conversation:

 - Use user_type instead of help_test_features flag - experimental features should be available to users of type `admin` and `superuser` (we should eventually find a better way to do more granular scopes rather than these broad user_types)

@yellowcap I think this looks good. The thing I'm trying to think through a bit is if the list of tools that are experimental vs default can (or should) be tweakable via environment config. The idea being that adding or removing a "feature flag" should involve no "code changes" just config. Am not sure we need to do that in this case, but am just trying to think about that / whether it seems like its a good idea.

yellowcap:
> just config / admin and superuser

@batpad ok lets add an additional field with profile choices that we can then apply. Will update

jterry64 (changes requested):
@yellowcap do you have some time to discuss with @gtempus and I? Or we can discuss async here. Gary is adding some concept of feature flags on the FE, and it'd be good to have them aligned so it's easy for the FE to pass feature flags through: https://github.com/wri/project-zeno-next/pull/522/changes

I think the important principles, is just that you can pass the feature flag as a query param, rather than assigning to users, and having feature flags per feature than just experimental/default. This lets anyone testing something on the FE use the query param in the app URL, and then the FE can just forward it to the backend. E.g. for your sentinel 2 work, you could pass one "sentinel-2-imagery" feature flag to globalnaturewatch.org/app?ff=sentinel-2-imagery to load both the FE and BE side.

I think all the infra would support that. How your setting the system prompt and tools from a profile looks great, I think just instead of assigning users or having to patch, it'd just load a particular profile based on the feature flag passed to /api/chat. Do you think that makes sense? We can restrict accessing feature flags to specific users, although we could also use this to A/B test with public users in the future.

jterry64:
@danscales I branched off your PR and created https://github.com/wri/project-zeno/pull/714, closing this one as an archive since I made quite significant changes

batpad:
Just to add my two-bits here: I think it makes perfect sense and a lot more logical for the frontend to handle this and send these as query parameters, especially since we don't see the experimental features being "security" issues where we want to restrict access, so it doesn't need to be based on backend user_types. Thanks much @jterry64 @gtempus, this also addresses my concern around how to make this a bit more configurable - frontend handling the configuration of what the feature flags are and the backend being agnostic to that makes a whole lot of sense.

PR #720, opened 2026-06-22
fix db ingestion error when getting a nul byte character from langfuse

batpad:
Got some errors when running the langfuse ingest script on staging like:

```
asyncpg.exceptions.CharacterNotInRepertoireError: invalid byte sequence for encoding "UTF8": 0x00
```

Most of the ingests were happening fine, but the ingest crashed out with this when processing a trace from a few months ago. Currently unsure of why this trace returns a `nul` character, that then causes the db insert to fail. But this should fix it by stripping out all nul chars before we ingest into the db. If we see this happening a lot, we can separately investigate what's happening in Langfuse, but I think this is likely some edge-case that's not worth worrying about, and just stripping out the offending characters before ingesting should be fine.

cc @yellowcap

### DIFF
--- a/src/api/services/langfuse/ingest.py
+++ b/src/api/services/langfuse/ingest.py
@@ -61,6 +61,18 @@ def _parse_dt(raw: Any) -> Optional[datetime]:
         return None
 
 
+def _strip_nul(value: Any) -> Any:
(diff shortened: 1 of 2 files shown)


yellowcap (approved):
🧹

PR #721, opened 2026-06-22
add a concept of scopes for machine users to have granular scopes

batpad:
This landed up being a bit of a side-quest for the Traces API stuff.

### The Problem

With the new Traces API, we want access to be restricted to superusers, but also enable machine-to-machine use-cases, for example from the `tracey` app that @01painadam was working on. We currently have a concept of a "machine_user" type, but no ability to specify more granular permissions per machine user / token. This means we'd have to allow _all_ machine user tokens to access the Traces API, which would have been non-ideal.

### What this PR does

This PR introduces a new `scopes` field for machine users, with the `scopes` then being baked into the token generated. In this PR, we creates a `traces:read` scope and then check on the traces endpoint if the user is a `superuser` OR a machine user with that scope.

This adds the `scopes` field to the schema and db, so we can expand this to include granular scopes assigned to machine tokens for specific tasks. I think this will likely be useful long-term to ensure we are giving fine-grained access to tokens.

@sunu - I think we've spoken about this in the past - if you have a moment, it would be good to get your 👀 on this and any quick feedback or thoughts.

cc @yellowcap

### DIFF
--- a/db/alembic/versions/e0e882fba200_add_scopes_to_machine_user_keys.py
+++ b/db/alembic/versions/e0e882fba200_add_scopes_to_machine_user_keys.py
@@ -1,0 +1,39 @@
+"""add scopes to machine_user_keys
+
+Revision ID: e0e882fba200
+Revises: 8391583bd224
+Create Date: 2026-06-22 00:00:00.000000
+
+Per-key authorization scopes for machine API keys (e.g. ``traces:read``). A key
+carrying a scope grants access to the matching endpoints without an elevated
+user_type, decoupling authentication (the key) from authorization (the scope).
+Existing keys default to no scopes.
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = "e0e882fba200"
+down_revision: Union[str, None] = "8391583bd224"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "machine_user_keys",
+        sa.Column(
+            "scopes",
+            postgresql.ARRAY(sa.String()),
+            nullable=False,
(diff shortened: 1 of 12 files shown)


yellowcap (approved):
Looks good thanks @batpad !

sunu (approved):
@batpad I think this looks good to me overall as an immediate step. The one thing I'm not super happy about right now is that we special-case the `superuser` type and end up checking both `user_type` and `token scopes`. A side effect is that the current model can't grant a user a subset of superuser functionality without making them a full superuser since scopes are limited to machine user tokens only.

This is fine for now and solves our immediate need. But I think it'd be worth eventually moving toward a model where instead of **user types** we have **roles** (thinking of this as instead of each user belonging to a single type only they can have multiple roles), and roles grant a set of **scopes**, and all our authorization checks are against scopes. 

That'd make things a lot more uniform and every endpoint will just ask "does this request carry scope X?" instead of special-casing user types. It'd also let us give someone fine-grained access (say, a non-superuser who can read traces) without handing out the full superuser role, and down the line users could even mint their own tokens scoped to a subset of what they're allowed, where the token only ever narrows access and never widens it (sort of how Github tokens work).

In that world require_scope collapses to a single check, with all the policy living in one resolver. something like:

```python
def effective_scopes(user, token_scopes):
    """Scopes this request effectively carries.
    A token can only narrow the user's granted scopes, never widen them."""
    granted = scopes_from_user_roles(user.roles)
    if token_scopes:
        return granted & set(token_scopes)   # token downscopes
    return granted

def has_scope(effective, required):
    return WILDCARD in effective or required in effective
```

 and then endpoint dependency just has to do `has_scope(effective_scopes(user, token_scopes), scope)`. And in this set up, superuser will map to the wildcard scope which will work naturally without needing any scpecial-case handling.

This is not something blocking for this PR at all but just where I'd love to see us go eventually I think.

batpad:
Thanks much @sunu - I made an issue to refactor our permissions model a bit based on your comment: https://github.com/wri/project-zeno/issues/722

Going to go ahead and merge this to unblock on the traces API testing.

PR #724, opened 2026-06-24
PZB-979: insights tool uses a stand-alone narrative generator llm call

solomon-negusse:
This splits the analyst tool into two steps: the code executor handles chart generation, and a second LLM narrates the insight from the chart data. It also introduces a seam model for insights to improve interoperability between the chat and analysis chart-generation paths. Also this adds insight persistence with both the chat and /api/analyze paths now writing InsightOrm/InsightChartOrm through one place.

This is primarily groundwork: it makes deterministic chart generators easy to add and decouples the narrative stage from how the charts were built, so it can be reused across paths. This decoupled insight narrative generator can easily be a separate tool that we surface to the orchestrator as well.

### DIFF
--- a/src/agent/subagents/analyst/charts/model.py
+++ b/src/agent/subagents/analyst/charts/model.py
@@ -6,8 +6,7 @@ object passed between stages and into persistence, and it exposes adapters for
 the two wire shapes:
 
 - snake_case for the DB (`InsightChartOrm`) via `to_orm_kwargs()`
-- camelCase for the frontend `charts_data` via `to_frontend_dict()` /
-  `from_frontend_dict()`
+- camelCase for the frontend `charts_data` via `to_frontend_dict()`
 
 `InsightBundle` groups the resolved charts with the narrative produced by the
 text stage.
@@ -78,29 +77,6 @@ class InsightChart(BaseModel):
             "groupField": self.group_field,
             "seriesFields": self.series_fields,
         }
-
-    @classmethod
-    def from_frontend_dict(
-        cls, data: dict, position: int = 0
-    ) -> "InsightChart":
-        """Inverse of `to_frontend_dict()` — rehydrate a chart already in state.
-
-        `id` is derived from `position` on the way out, so it is not read back
-        here; the caller supplies `position` (defaulting to the list index).
-        """
-        return cls(
-            position=position,
-            title=data.get("title", ""),
-            chart_type=data.get("type", "bar"),
-            x_axis=data.get("xAxis", ""),
-            y_axis=data.get("yAxis", ""),
-            color_field=data.get("colorField", ""),
-            stack_field=data.get("stackField", ""),
-            group_field=data.get("groupField", ""),
-            series_fields=data.get("seriesFields", []),
-            chart_data=data.get("data", []),
-            insight=data.get("insight", ""),
-        )
 
     @classmethod
     def from_chart_insight(
--- a/src/agent/subagents/analyst/charts/model.py
+++ b/src/agent/subagents/analyst/charts/model.py
@@ -12,14 +12,32 @@ the two wire shapes:
 text stage.
 """
 
+import math
 from typing import List
 
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 
 from src.agent.subagents.analyst.code_executors.base import (
     CHART_TYPES_WITHOUT_AXIS,
     ChartInsight,
 )
+
+
+def _json_safe(value):
+    """Recursively replace NaN/Inf floats with None.
+
+    `chart_data` rows come from `pandas.DataFrame.to_dict()`, which emits
+    `float('nan')` for missing cells. `NaN`/`Infinity` are not valid JSON tokens
+    under Postgres' strict `JSONB` parser, so they must be scrubbed before the
+    row reaches `to_orm_kwargs()` (DB) or `to_frontend_dict()` (frontend).
+    """
+    if isinstance(value, float) and not math.isfinite(value):
+        return None
+    if isinstance(value, dict):
+        return {k: _json_safe(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_json_safe(v) for v in value]
+    return value
 
 
 class InsightChart(BaseModel):
@@ -40,6 +58,12 @@ class InsightChart(BaseModel):
     series_fields: List[str] = Field(default_factory=list)
     chart_data: List[dict] = Field(default_factory=list)
     insight: str = ""
+
+    @field_validator("chart_data")
+    @classmethod
+    def sanitize_chart_data(cls, rows: List[dict]) -> List[dict]:
+        """Scrub non-finite floats (NaN/Inf) so rows are valid JSONB."""
+        return [_json_safe(row) for row in rows]
 
     @model_validator(mode="after")
     def validate_axis_config(self) -> "InsightChart":
--- a/src/agent/subagents/analyst/charts/model.py
+++ b/src/agent/subagents/analyst/charts/model.py
@@ -1,0 +1,139 @@
+"""Canonical in-code chart model — the seam between the chart stage and the
+text stage of insight generation.
+
+`InsightChart` carries the chart spec *and* its row data. It is the single typed
+object passed between stages and into persistence, and it exposes adapters for
+the two wire shapes:
+
+- snake_case for the DB (`InsightChartOrm`) via `to_orm_kwargs()`
+- camelCase for the frontend `charts_data` via `to_frontend_dict()` /
+  `from_frontend_dict()`
+
+`InsightBundle` groups the resolved charts with the narrative produced by the
+text stage.
+"""
+
+from typing import List
+
+from pydantic import BaseModel, Field, model_validator
+
+from src.agent.subagents.analyst.code_executors.base import (
+    CHART_TYPES_WITHOUT_AXIS,
+    ChartInsight,
+)
+
+
+class InsightChart(BaseModel):
+    """A single chart: spec + the rows it renders.
+
+    Mirrors `InsightChartOrm` (snake_case) plus the embedded `chart_data`. The
+    optional `insight` field holds the per-chart narrative from the text stage.
+    """
+
+    position: int = 0
+    title: str
+    chart_type: str
+    x_axis: str = ""
+    y_axis: str = ""
+    color_field: str = ""
+    stack_field: str = ""
+    group_field: str = ""
+    series_fields: List[str] = Field(default_factory=list)
+    chart_data: List[dict] = Field(default_factory=list)
+    insight: str = ""
+
+    @model_validator(mode="after")
+    def validate_axis_config(self) -> "InsightChart":
+        if self.chart_type not in CHART_TYPES_WITHOUT_AXIS:
+            if not self.y_axis and not self.series_fields:
+                raise ValueError(
+                    f"Chart '{self.title}' (type '{self.chart_type}') is missing axis "
+                    "configuration: set 'y_axis' for single-series charts or "
+                    "'series_fields' for multi-series charts"
+                )
+        return self
+
+    # --- adapters -----------------------------------------------------------
+
+    def to_orm_kwargs(self) -> dict:
+        """Kwargs for `InsightChartOrm(**...)` (snake_case).
+
+        `insight` is excluded: per-chart narrative is not a chart-row column; it
+        lives on the parent `InsightOrm.insight_text`.
+        """
+        return self.model_dump(exclude={"insight"})
+
+    def to_frontend_dict(self) -> dict:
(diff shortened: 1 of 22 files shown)


src/agent/subagents/analyst/text_generator.py
```diff
@@ -0,0 +1,98 @@
...
+
+_PROMPT = ChatPromptTemplate.from_messages(
+    [("system", _SYSTEM), ("user", _USER)]
+)
+
+
+class InsightTextGenerator:
+    """Generates insight text from resolved charts."""
+
+    def __init__(self, model=GEMINI_FLASH):
```
yellowcap: Lets not hard code this here, use the small model not an explicit one.

yellowcap (commented):
I had a look, but not super thorough yet. Here a few initial comments:

The logic per se seems fine,  I do wonder what this means for latency and token usage, and if this deep change is required.

For latency: @solomon-negusse can you run the gold set evals against this using a localhost? I have done this previously and was a good indicator for what has changed in that regard.

About the change: If the main goal is to allow making descriptions on verified charts, then we can potentially do that separately and leave the generative insights generation as-is. Has that been considered?

If latency and token costs are ok then I am happy to support this split. We had something along those lines before, but bundled things to reduce the amount of llm calls.

solomon-negusse:
Thanks for the review @yellowcap. 

>About the change: If the main goal is to allow making descriptions on verified charts, then we can potentially do that separately and leave the generative insights generation as-is. Has that been considered?

That's the immediate goal. The other is that, given the insight tool dominates agent latency, we also wanted to reuse the deterministically generated chart data in the insight tool - we hypothesize most user queries can be served by that, improving latency dramatically. This lays the groundwork for that path as well.


>can you run the gold set evals against this using a localhost? I have done this previously and was a good indicator for what has changed in that regard.

I did this on a subset of eval (18 traces; I have only ingested admin areas locally) and not seeing significant jump in cost and latency:

Cost/tokens: ~flat — 482.86K → 496.42K units (+2.8%), $0.245 → $0.267. 
Latency: End-to-end trace p50 is unchanged within noise (53s → 48s). No significant regression.

left - main;  right - refactor
insights_cost_comparison

insights_latency_comparison

jterry64 (commented):
Looking now, but a couple initial questions:

- I do like that you refactored and neatened up the models to reflect charts vs. insights, but it seems like then you're adapting it to the current database model. Would it be better to just run a migration on the insights model while it's not being used much to reflect this schema better? Or did you want to do that later once we validate it?
- What were the gold set results in terms of accuracy? Any significant differences, or did it produce roughly the same results?

src/api/services/job.py
```diff
@@ -2,9 +2,12 @@
 from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 from uuid import UUID
 
+if TYPE_CHECKING:
```
jterry64: Should the model be moved out to somewhere common if it's used also in /analyze? It seems like you're having to deal with weird imports otherwise

src/agent/subagents/analyst/charts/model.py
```diff
@@ -0,0 +1,115 @@
...
+            y_axis=chart.y_axis,
+            color_field=chart.color_field,
+            stack_field=chart.stack_field,
+            group_field=chart.group_field,
+            series_fields=chart.series_fields,
+            chart_data=chart_data,
+        )
+
+
+class InsightBundle(BaseModel):
```
jterry64: Maybe just a knit, but I don't like bundle in the name. I feel like this should just be "Insight", and it has "Charts" and a "Narrative" or "Text". Or maybe they should just be totally decoupled - ostensibly with this change, you could have a lot of different text responses to the same chart, right?

src/agent/subagents/analyst/tool.py
```diff
@@ -237,246 +235,213 @@ def build_analysis_prompt(
...
+    if dataset_cautions:
+        tool_message += f"\n\nDataset cautions:\n{dataset_cautions}"
+
+    tool_message += "\n\nFollow-up suggestions:"
+    for i, suggestion in enumerate(bundle.follow_up_suggestions, 1):
+        tool_message += f"\n{i}. {suggestion}"
+    return tool_message
 
 
 class Analyst:
```
jterry64, line 301: I feel like it might good to do some more renaming here. I feel like in other places "insights" means the text, but now "text" is a separate tool, so "insights" is a bit confusing in this. Also, now "text" is a specific tool, whereas this is just the generic "analyst" tool that only makes charts. So I feel like this should just be named as the chart generation tool to be clear about what it does now.

src/api/repositories/insight_writer.py
```diff
@@ -0,0 +1,62 @@
...
+
+from src.agent.subagents.analyst.charts.model import InsightBundle
+from src.api.data_models import InsightChartOrm, InsightOrm
+from src.shared.database import get_session_from_pool
+from src.shared.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+async def persist_insight(
```
jterry64, line 18: nice, glad this got pulled out to a repository

jterry64 (changes requested):
Nice! There's a lot of good refactoring here, and I think the tool split will simplify things vs. having a tool that both makes charts + text and separate tool to just make text. Thanks you providing charts on latency/cost. In the future, it might be better to do that refactoring in a separate first PR, because it hard to disentangle the refactor from the tool changes when reading. Or, I've had some success just asking AI to make good descriptions and mermaid diagrams of the changes for the PR description to make it clearer what's going on for big PRs.

Per our discussion, let's just run the full gold set to check this isn't causing any regresssions.

src/api/services/job.py
solomon-negusse: Cleaned this up in https://github.com/wri/project-zeno/pull/724/commits/96b5e0b55ff456d508d98c816068f3f3e5fdb28e but left the path as is since it's an existing pattern and need to move several files - will have to be a follow up PR.

src/agent/subagents/analyst/charts/model.py
solomon-negusse: Good call, updated the name for now. Let's discuss the latter - as it stands, it's one to one insight text and chart mapping.

src/agent/subagents/analyst/tool.py
solomon-negusse: As we discussed, the tool is actually doing both things right now - just calls the stand-alone narrative llm function instead of doing it inline. Until we split that at tool level, I think the analyst name is ok.

solomon-negusse:
>Nice! There's a lot of good refactoring here, and I think the tool split will simplify things vs. having a tool that both makes charts + text and separate tool to just make text. Thanks you providing charts on latency/cost. In the future, it might be better to do that refactoring in a separate first PR, because it hard to disentangle the refactor from the tool changes when reading. Or, I've had some success just asking AI to make good descriptions and mermaid diagrams of the changes for the PR description to make it clearer what's going on for big PRs.
Per our discussion, let's just run the full gold set to check this isn't causing any regresssions.

Agreed @jterry64: the scope went a bit out of hand on this one :/ Thanks for the tips on the PR description. Here is eval comparison. I think the small dip is within normal variance but will run it with multiple number of trials to confirm. 

**main**:

EVALUATION SUMMARY
Environment:           localhost
API Base URL:          http://localhost:8000
Run timestamp:         2026-06-25 20:45:03 UTC
GNW code version:      unknown
GNW prompts version:   unknown
GNW agent LLM:         unknown (metadata unavailable)
Eval judge LLM:        Anthropic / claude-haiku-4-5
Test latency (s):      avg=73.3, std=37.5, min=9.0, max=195.5 (n=62)

Tests Run (after filters): 62

Agent Answer:              40 /  50 (0.80)

AOI ID Match:              45 /  47 (0.96)
Dataset ID Match:          46 /  50 (0.92)
Dataset Parameter Match:    0 /   0
Context Layer Match:        2 /   3 (0.67)
Data Pull Exists:          48 /  50 (0.96)
Date Match:                23 /  30 (0.77)
Charts Answer:             31 /  42 (0.74)
Expected Text Match:        7 /   8 (0.88)
Clarification Requested:    4 /   8 (0.50)
Suggested Datasets:         5 /   5 (1.00)

(warning: overall_score is experimental and untested)
Tests with overall score ≥0.7:   51 /  62 (82.3%)

**refactored**

EVALUATION SUMMARY
Environment:           localhost
API Base URL:          http://localhost:8000
Run timestamp:         2026-06-25 21:36:46 UTC
GNW code version:      2026.6.17.2
GNW prompts version:   unknown
GNW agent LLM:         primary gemini-flash (ChatGoogleGenerativeAI), small gemini-flash (ChatGoogleGenerativeAI)
Eval judge LLM:        Anthropic / claude-haiku-4-5
Test latency (s):      avg=68.9, std=40.2, min=10.4, max=239.8 (n=62)

Tests Run (after filters): 62

Agent Answer:              38 /  50 (0.76)

AOI ID Match:              45 /  46 (0.98)
Dataset ID Match:          45 /  50 (0.90)
Dataset Parameter Match:    0 /   0
Context Layer Match:        2 /   2 (1.00)
Data Pull Exists:          46 /  50 (0.92)
Date Match:                23 /  30 (0.77)
Charts Answer:             34 /  44 (0.77)
Expected Text Match:        8 /   8 (1.00)
Clarification Requested:    4 /  10 (0.40)
Suggested Datasets:         4 /   5 (0.80)

(warning: overall_score is experimental and untested)
Tests with overall score ≥0.7:   48 /  62 (77.4%)

jterry64 (approved):
Thanks for the changes, as long as the gold set doesn't regress it's good to merge!

solomon-negusse:
No regression running evals with 3 trials run.


**main**

EVALUATION SUMMARY
Environment:           localhost
API Base URL:          http://localhost:8000
Run timestamp:         2026-06-26 06:36:21 UTC
GNW code version:      2026.6.17.2
GNW prompts version:   unknown
GNW agent LLM:         primary gemini-flash (ChatGoogleGenerativeAI), small gemini-flash (ChatGoogleGenerativeAI)
Eval judge LLM:        Anthropic / claude-haiku-4-5
Test latency (s):      avg=56.3, std=39.4, min=12.7, max=278.5 (n=62)

Tests Run (after filters): 62
Trials per test: 3

Agent Answer:             0.74 ± 0.12 (n=50, 3 trials)

AOI ID Match:             0.96 ± 0.01 (n=46, 3 trials)
Dataset ID Match:         0.90 ± 0.06 (n=50, 3 trials)
Dataset Parameter Match:    0 /   0
Context Layer Match:      1.00 ± 0.00 (n=3, 3 trials)
Data Pull Exists:         0.91 ± 0.06 (n=50, 3 trials)
Date Match:               0.77 ± 0.21 (n=30, 3 trials)
Charts Answer:            0.77 ± 0.06 (n=47, 3 trials)
Expected Text Match:      0.96 ± 0.07 (n=8, 3 trials)
Clarification Requested:  0.31 ± 0.00 (n=13, 3 trials)
Suggested Datasets:       1.00 ± 0.00 (n=5, 3 trials)

(warning: overall_score is experimental and untested)
Tests with overall score ≥0.7:   47 /  62 (75.8%)


**refactored insight**

EVALUATION SUMMARY
Environment:           localhost
API Base URL:          http://localhost:8000
Run timestamp:         2026-06-26 05:42:12 UTC
GNW code version:      2026.6.17.2
GNW prompts version:   unknown
GNW agent LLM:         primary gemini-flash (ChatGoogleGenerativeAI), small gemini-flash (ChatGoogleGenerativeAI)
Eval judge LLM:        Anthropic / claude-haiku-4-5
Test latency (s):      avg=53.7, std=29.4, min=9.8, max=213.8 (n=62)

Tests Run (after filters): 62
Trials per test: 3

Agent Answer:             0.73 ± 0.08 (n=50, 3 trials)

AOI ID Match:             0.96 ± 0.00 (n=46, 3 trials)
Dataset ID Match:         0.90 ± 0.05 (n=50, 3 trials)
Dataset Parameter Match:    0 /   0
Context Layer Match:      1.00 ± 0.00 (n=3, 3 trials)
Data Pull Exists:         0.93 ± 0.03 (n=50, 3 trials)
Date Match:               0.79 ± 0.13 (n=30, 3 trials)
Charts Answer:            0.71 ± 0.13 (n=48, 3 trials)
Expected Text Match:      1.00 ± 0.00 (n=8, 3 trials)
Clarification Requested:  0.36 ± 0.00 (n=11, 3 trials)
Suggested Datasets:       0.93 ± 0.12 (n=5, 3 trials)

(warning: overall_score is experimental and untested)
Tests with overall score ≥0.7:   46 /  62 (74.2%)

yellowcap:
Thanks for the thorough review of latency and cost @solomon-negusse 🙏🏽 given @jterry64 approved, I am comfortable for this to move as well without further review.

PR #725, opened 2026-06-24
Integrated alerts

jterry64:
Replaces DIST alerts with integrated alerts. Keeping old DIST alerts yaml around for now, we can remove it later, but added a flag to no longer include it in the agent.


solomon-negusse (approved):
Looks good!

PR #728, opened 2026-06-30
Searchable AOIs: unified GET /api/aois endpoint

yellowcap:
## What

Adds a unified, authenticated **`GET /api/aois`** endpoint to search Areas of Interest across all sources (`gadm` / `kba` / `wdpa` / `landmark` / `custom`), with multi-select source filtering and fuzzy name search. Reuses the agent's existing `pick_aoi` pg_trgm search core.

## Why

This is in preparation for dashboards, where we will start single-area-dashboards from selecting an aoi. This allows the user to search for aois outside of the map explorer.

There was no way to search AOIs across sources from the API — custom areas could only be listed unfiltered (`GET /api/custom_areas`) and reference AOIs fetched one at a time (`GET /api/geometry/{source}/{src_id}`). Users want to search by name and filter by source type.

## Endpoint

`GET /api/aois`

| Param | Notes |
|---|---|
| `name` | Fuzzy, similarity-ranked search. Omit to **browse** alphabetically. |
| `source` | Repeatable multi-select: `gadm`, `kba`, `wdpa` (alias `protectedareas`), `landmark`, `custom`. Omit = all sources. |
| `limit` | 1–100, default 50. |
| `offset` | Offset pagination; next page returned via the `X-Next-Offset` response header. |

Custom areas are scoped to the authenticated caller.

```
GET /api/aois?name=amazon
GET /api/aois?source=gadm&source=custom
GET /api/aois?source=custom                 # browse my custom areas
GET /api/aois?source=protectedareas&name=yellowstone
```

## How (reuse)

- **`src/shared/geocoding_helpers.py`** — new `search_aois(name, sources, user_id, limit, offset)` generalizes the old union builder: optional name (fuzzy vs browse mode), a *list* of sources incl. `custom`, an explicit `user_id`, and offset pagination. Relocated the bbox SQL helpers here, plus `normalize_aoi_source` (handles the `protectedareas`→`wdpa` alias + validation).
- **`src/agent/subagents/pick_aoi/tool.py`** — `query_aoi_database` now delegates to `search_aois`; the geocoder's behavior is unchanged.
- **`src/api/routers/aois.py`** (new) + registered in `app.py`; **`AOISearchResult`** schema in `schemas.py`.

No DB migration required — the `custom` source is implicit.

## Testing

- New `tests/api/test_aois.py`: fuzzy search, browse-by-source, owner scoping, offset pagination + `X-Next-Offset`, invalid-source `422`, `protectedareas` alias, auth required.
- Existing `pick_aoi` / `custom_area` suites pass (incl. the replay-based geocoder test), confirming the delegation refactor is regression-free.

## Notes

- Browse-with-no-name over the large reference tables uses `OFFSET` pagination — fine for `limit<=100`; keyset is a future optimization if browse-all gets heavy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.1"
+version = "2026.7.1.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/subagents/pick_aoi/tool.py
+++ b/src/agent/subagents/pick_aoi/tool.py
@@ -26,7 +26,8 @@ from src.agent.subagents.progress import emit_progress
 from src.agent.tool_spec import ToolCategory, ToolSpec
 from src.shared.database import get_connection_from_pool
 from src.shared.geocoding_helpers import (
-    CUSTOM_AREA_TABLE,
+    BBOX_SQL,
+    CUSTOM_BBOX_SQL,
     GADM_STANDARD_ID_RE,
     GADM_TABLE,
     KBA_TABLE,
@@ -34,6 +35,7 @@ from src.shared.geocoding_helpers import (
     SOURCE_ID_MAPPING,
     SUBREGION_TO_SUBTYPE_MAPPING,
     WDPA_TABLE,
+    search_aois,
 )
 from src.shared.logging_config import get_logger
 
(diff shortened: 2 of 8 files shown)


batpad:
@yellowcap am slightly concerned with performing trigram similarity search across multiple tables - can you say a bit how you expect this will be used? i.e. via an autocomplete from the frontend, etc?

How many records is the current total number we are searching through, do we see this growing? I _think_ I'd be more comfortable creating  a single table that holds all the strings being searched through. Let's discuss this?

We should also ensure we have trigram indexes setup on all the fields we are doing similarity search on.

I think this looks fine - would just like to at least interrogate the possibility of doing this search in a single table and doing some sort of consolidation here - it could be something like a materialized view I think - would also make the SQL doing the search a bit easier to reason about.

yellowcap:
@batpad ok lets discuss this, I do see the concern. This re-uses search that the agent uses under the hood for a while, so it should scale ok to the size of the real DB.

The intended use is for search on a "table" view of AOIs, when a user wants to make a new dashboard but they are not on the explorer page. I.e. if they are on the "Create new dashboard" flow on the list of dashboards.

In that context typeahead would be great but may be too difficult to do for this kind of data. But I am not savvy on this so I can't really tell. Lets go throug hthis together in a meeting.

batpad:
@yellowcap I've been chatting with Claude a bit and working a bit on a structure on how we could refactor this. I do feel this is important to refactor for the long-term - will speed up search and filters by a lot, and make it a lot more convenient to add more search and filter features.

I also think that it's perfectly fine to do after merging this / as a separate PR. This PR does help to consolidate where search happens and make refactoring this easier.

I'll make a separate ticket about the table refactor. Esp. after my conversation with Claude, I do feel strongly about implementing that and having a single table to do search and filter on - currently, we are doing some very expensive things in the SQL, and as complexity grows, that's going to get harder to reason about and optimize. Am happy to take that on.

I think we can merge this and move ahead though! I'll ticket the refactor separately.

batpad (approved):

PR #729, opened 2026-06-30
Persist dataset_id and aoi_ids on statistics

yellowcap:
## Motivation

This adds ids for AOI and Datasets to the statistics ORM.

This is in preparation of making insights filterable by AOI and dataset for discovery on dashboards.

## What

Persist stable ids on `StatisticsOrm` so the insights listing filter can match by id instead of by human-readable name:

- **`dataset_id`** (`Integer`, nullable) — the catalog id, alongside the existing `dataset_name`.
- **`aoi_ids`** / **`aoi_sources`** (`JSONB`, parallel to `aoi_names`) — the AOI `src_id`s and their sources. `src_id` is only unique per source (gadm/kba/wdpa/landmark/custom), so the source is stored alongside each id.

`pull_data` now writes all three; the `Statistics` TypedDict in `state.py` is kept in sync.

## Migration

`a1b2c3d4e5f6` adds the three columns (down_revision `e0e882fba200`) and **backfills `dataset_id` for existing rows** from `dataset_name`, using a frozen `VALUES` snapshot of the 12-dataset catalog. The backfill is failsafe: it's an inner join, so any row whose name isn't in the snapshot is left untouched (stays `NULL`) — never errored, never mis-attributed. Names are unique, so a match can't pick the wrong id. It's idempotent (`AND dataset_id IS NULL`).

`aoi_ids` / `aoi_sources` are **not** backfilled — `src_id`/`source` were never persisted for historical rows and can't be reliably recovered (only ambiguous `aoi_names` was stored). Historical rows keep the `[]` default; id-based AOI filtering applies to new rows only.

## Scope

This PR only persists + backfills the columns. The listing-endpoint filter that consumes them (`dataset_name == dataset` → `dataset_id == dataset`, `aoi_names.has_key` → `aoi_ids.has_any`) lives in the sibling branch `feat/filter-insights`.

## Test plan

- [x] `tests/tools/test_pull_data.py` extended to assert `dataset_id`, `aoi_ids`, `aoi_sources` persist; passing.
- [x] Migration `VALUES` snapshot cross-checked against the live `DATASETS` catalog (12/12 exact match).
- [x] `alembic heads` → single head; migration parses.
- [x] Apply on a DB copy: `cd db && uv run alembic upgrade head`, then `SELECT dataset_name, dataset_id, count(*) FROM statistics GROUP BY 1,2;`

## Next steps

Add filters to the list insights endpoint
https://github.com/wri/project-zeno/blob/be54688d145df2eb1b54f82a2d67e389a1e380b4/src/api/routers/insights.py#L63

### DIFF
--- a/db/alembic/versions/a1b2c3d4e5f6_add_dataset_and_aoi_ids_to_statistics.py
+++ b/db/alembic/versions/a1b2c3d4e5f6_add_dataset_and_aoi_ids_to_statistics.py
@@ -1,0 +1,85 @@
+"""add dataset_id and aoi_ids/aoi_sources to statistics
+
+Revision ID: a1b2c3d4e5f6
+Revises: e0e882fba200
+Create Date: 2026-06-30 00:00:00.000000
+
+Persist the catalog dataset id and the src_ids (with their sources) of the
+analysed AOIs alongside the existing human-readable names, so the insights
+listing endpoint can filter by stable ids instead of names. src_id is only
+unique per source, so aoi_sources is stored parallel to aoi_ids. All columns
+are nullable / defaulted so existing rows stay valid without backfill.
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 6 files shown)


batpad (approved):
lgtm

PR #733, opened 2026-07-01
PZB-1026: Initial draft implementation for insights handling [not merge]

yellowcap:
⚠️ Do not merge, this is just for visibility on these initial experiments ⚠️ 

These are musings / initial experimentation for tooling that can work on insights.

### Motivation

This is in the context of dashboards, where the user will probably expect the agent to find past insights, edit them, or create new ones.

### What is in the experiments

Two new generative tools for working with already-persisted insights, plus the plumbing they share:

- update_insight_display: restyles an existing insight (narrative text, follow-ups, chart titles/types, field remapping) without pulling new data or running code. InsightDisplayReviser only re-maps among existing columns; chart_data rows are preserved. Persisted in place via update_insight() and pushed onto state so the frontend re-renders.
- search_insights: finds a past insight across ALL of the user's conversations (own + public) by ILIKE over summary, chart titles and chart_data text, surfaces the best match into state like the generator.

Both emit response_metadata msg_type "insight_updated" + insight_id so the frontend replaces/re-fetches the insight by id rather than rendering a duplicate card. search is steered to be terminal (no follow-on analysis) via prompt + tool-message instructions and a routing rule.

Supporting changes:
- InsightChart.from_orm_row() to rebuild the seam model from a DB row.
- update_insight() in insight_writer for in-place display updates.
- inspect_view_context: extra info logging.
- schemas: document visible_insights (list of uuids) in view_context.
- Tests for both new tools.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.2"
+version = "2026.7.2.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -29,7 +29,11 @@ from src.agent.tools.inspect_view_context import (
     SPEC as inspect_view_context_spec,
 )
 from src.agent.tools.pull_data import SPEC as pull_data_spec
+from src.agent.tools.search_insights import SPEC as search_insights_spec
 from src.agent.tools.show_imagery import SPEC as show_imagery_spec
+from src.agent.tools.update_insight_display import (
+    SPEC as update_insight_display_spec,
+)
 from src.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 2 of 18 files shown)


yellowcap:
A few thoughts here after playing around more:

1. How does the agent address insights? Do we put insight IDs into the message history and expect the agent to pass these to tools (fuzzy, error prone)? Or is it the currently active one that we pass through state?
2. Are insight updates in-place, vs creating new ones? Currently description and chart type are in-place, because no new data. But that is not necessarily desirable from the user point of view. User confirmation needed for that? Insights versioning instead of new instances?
3. How verbose is the "inspection tool". For dashboards with many insights, we dont want to print too many things into conversation history. 
     - An inspection of single insight, does it print data, only title? Is the description relevant?
     - Do we have tiered inspection? Verbose mode for inspection?
     - Can we specify keys for inspection? Like "verbose for insights A, B, and C"
4. How do we handle page awareness? I.e. where do we document what a page represents? Sounds like a skill, like "read about the dashboard page", but could also be in system prompt, or part of inspection (inspect current page type)
5. Maybe inspection is a sub-agent that does a lot of the decisions listed above dynamically without polluting the orchestrator too much, and just returns quite targeted digests.

yellowcap:
closing in favor of #741

PR #740, opened 2026-07-02
Fixes for feedback on integrated alerts

jterry64:
This includes a few fixes:
- Use the correct tile service for int dist alerts
- Always report in hectares
- Tweak prompt to use integrated alerts only for general alert queries, and still use DIST alerts for any contextual filer like forests, grasslands, natural areas, etc.
- Also only use one chart for DIST alerts output to avoid same bug with messed up multiple charts

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.2"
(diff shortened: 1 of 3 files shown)


danscales (approved):
Looks good!

PR #742, opened 2026-07-03
Add turn information to traces for easier analysis

batpad:
Based on this request from @01painadam to add turn information to the Traces API to facilitate better turn-level analytics:

# Turn-level analytics — plan for a position-independent endpoint

*2026-07-02 · Adam Pain (drafted with Claude)*

## 1. Current state: traces are already per-turn, but conversation position leaks in

A Langfuse trace **is** one agent turn — confirmed in project-zeno's
`src/api/services/langfuse/parse.py` docstring ("A Langfuse trace corresponds
to one agent turn") and `LangfuseTraceOrm` ("Derived analytics for one
Langfuse trace (= one agent turn)"). So `GET /api/traces` genuinely returns
one row per turn, not one row per conversation — there isn't literally an
endpoint that aggregates traces into conversations. Two things nonetheless
make turn-level analytics behave as if conversations were lumped together:

1. **No turn-position field.** Neither `TraceListItem` nor the
   `/api/traces/sessions` `SessionItem` carries an ordinal (`turn_index`,
   `is_first_turn`). The only way to know "is this the 1st, 2nd, 5th turn of
   its session" is to fetch every row in the session and sort by
   `trace_timestamp` client-side — which this dashboard already does, purely
   to approximate turn order:
   - `lib/analytics/outcomeRefine.ts::earliestToolTurnBySession` groups all
     rows by `sessionId` to find the earliest tool-using turn per session, so
     it can tell whether a later no-tool DEFER is really an answer served
     from earlier context.
   - `lib/analytics/outcomeRefine.ts::computeFailureFollowUps` groups by
     `sessionId`, sorts by timestamp, and walks the list to find "the next
     turn after a failure."

   Both are expensive (require the full window, not just the page being
   analyzed), silently wrong under pagination or a date range that cuts a
   session mid-thread (a session's earlier turns may simply not be in the
   fetched set), and can't be pushed down as a query filter — there's no way
   to ask the API for "just first turns" or "just turn 3+".

2. **Some "per-turn" fields are actually thread-cumulative**, exposed under
   names that don't signal this to a consumer of `/api/traces`:
   - `datasets_analysed` is sourced from
     `derived["datasets_analysed_cumulative"]` server-side — the underlying
     key literally says *cumulative*, but the public field strips that
     suffix. `CLAUDE.md` in this repo already flags this: *"`datasetsAnalysed`
     is thread-cumulative, not per-turn."*
   - `primary_dataset_name`, `has_insight`, `insight_id`, `aoi_name` reflect
     the AgentState's *current* value, which may have been set several turns
     earlier and simply persisted forward (state uses `operator.add` /
     carries prior fields). `parse_state`'s own docstring draws this
     distinction — *"Current-state fields (aoi/dataset/insight) are
     per-turn-meaningful (the selection in effect this turn). Cumulative
     fields... reflect the whole thread"* — but the API response doesn't
     surface that distinction, so a turn where the agent touched nothing new
     still reports a non-null `primary_dataset_name` / `has_insight`.

   Net effect: `/api/traces` rows are genuinely per-turn, but several of the
   interesting columns answer *"what does this conversation look like as of
   this turn"* rather than *"what happened in this turn"* — and there's no
   way to condition a query on where in the conversation a turn falls.

3. **The existing `GET /api/traces/analytics` aggregate has no turn-position
   filter either.** It SQL-aggregates `outcome_breakdown`, `daily_volume`,
   latency/cost/token percentiles, etc. over every matching row with equal
   weight. A 20-turn power-user session and a 1-turn session contribute
   equally to `outcome.p95` / `outcome_breakdown` — there's no way to ask
   "what does turn 1 look like vs. turn 5+" from that endpoint today. (Note:
   this dashboard doesn't currently call `/api/traces/analytics` at all — per
   `CLAUDE.md` it pulls raw `/api/traces` rows and aggregates client-side in
   `lib/analytics/*`, so any fix needs to keep that path viable, not just
   patch the aggregate endpoint.)

## 2. What "by turn, independent of conversation position" should mean

- A turn ordinal per trace: 1-based position within its session, ordered by
  `trace_timestamp`.
- A way to filter/segment a query by that ordinal: exact turn, turn range, or
  "first turn only" — without pulling the full session window client-side.
- Turn-scoped counterparts for the fields that are actually cumulative, so
  "did this turn produce X" and "does the thread currently have X" are two
  distinguishable answers instead of one overloaded field.
- The `/api/traces/analytics` aggregate supporting the same turn-position
  filter, so a turn-depth comparison doesn't require pulling every row.

### Phase 0 — `turn_index` on `GET /api/traces` (½–1 day, BE)

Compute at query time, no migration or backfill needed:

```sql
row_number() OVER (
  PARTITION BY session_id ORDER BY trace_timestamp ASC NULLS LAST, id ASC
) AS turn_index
```

Wrap the existing `_LIST_COLUMNS` select in a CTE (or subquery) so filters
still apply against the base table and `turn_index` is available for both
`WHERE` and the response. For rows with `session_id IS NULL` (soft reference,
no FK — machine users, dev envs, deleted threads all produce these per
`LangfuseTraceOrm`'s docstring), `turn_index` should resolve to `1` rather
than partition all of them together as one giant fake "session."

Add to `TraceListItem`: `turn_index: Optional[int]`,
`is_first_turn: Optional[bool]` (`turn_index == 1`).

Add query params to `GET /api/traces`: `turn_index` (exact), `min_turn_index`,
`max_turn_index`, `first_turn_only: bool` — applied as `WHERE` on the CTE
output, same pagination contract as today.

This alone unlocks the core ask: the FE can request e.g.
`first_turn_only=true` or `min_turn_index=2` and get turn-position-controlled
rows without any client-side session reconstruction, and `lib/analytics/*`
can add a turn-depth breakdown using the same client-aggregation pattern
already used for everything else in this dashboard.

### Phase 1 — make cumulative fields honest (1 day, BE)

Add true per-turn counterparts for the two fields most likely to be misread,
additive (existing cumulative fields stay, same non-goal as the outcome
taxonomy plan — don't break existing consumers):

- `datasets_analysed_this_turn`: cumulative list for this turn minus the
  cumulative list as of the *previous* turn in the same session (`lag(...)
  OVER (PARTITION BY session_id ORDER BY trace_timestamp)`, set-difference the
  two arrays). Distinguishes "this turn ran new analysis" from "the thread has
  analysed datasets."
- `insight_created_this_turn: bool`: true only when `insight_id` differs from
  the previous turn's `insight_id` in the same session.

Document `datasets_analysed` / `primary_dataset_name` / `has_insight` in the
OpenAPI schema as explicitly thread-cumulative-as-of-this-turn, so future
consumers don't assume otherwise (this repo's `outcomeRefine.ts`
`CONTEXT_ANSWER` logic deliberately relies on the cumulative behavior — that
usage is intentional and should keep working).

### Phase 2 — turn-position-aware `GET /api/traces/analytics` (1–2 days, BE)

Add the same `turn_index` / `min_turn_index` / `max_turn_index` /
`first_turn_only` params here, computed the same way (CTE with `row_number()`
ahead of the existing aggregate SQL). Lets one endpoint answer "how does turn
1 differ from turn 3+" via two calls instead of a full client-side pull.

Stretch: a `group_by=turn_index` mode returning one row of the existing
aggregate shape per turn-index bucket (1, 2, 3, 4, 5+) — a single response for
a turn-over-turn trend chart (outcome rate / latency / cost by conversation
depth), which is the most direct answer to "how does turn depth affect
outcome."

### Phase 3 — client integration (this repo, ½–1 day, FE)

- `lib/api/zeno.ts`: map `turn_index` / `is_first_turn` onto `TraceRow`,
  extend `TracesFilter` with `turnIndex` / `minTurnIndex` / `maxTurnIndex` /
  `firstTurnOnly`, thread them through `filterParams`.
- `lib/analytics/outcomeRefine.ts`: replace `earliestToolTurnBySession` and
  the manual sort-by-timestamp in `computeFailureFollowUps` with the
  server-provided `turn_index` — same intent, no full-window reconstruction,
  and it fixes a latent correctness gap: today's client-side grouping is
  silently wrong whenever pagination or a date-range filter excludes a
  session's earlier turns from the fetched set.
- New analytics section: outcome / latency / cost bucketed by `turn_index`
  (1, 2, 3, 4, 5+), once Phase 2 ships `group_by=turn_index`.

## 4. Non-goals

- Not changing what a "trace" means (still one row per turn) or introducing a
  new object type.
- Not deprecating `/api/traces/sessions` — that's still the right endpoint for
  the conversation browser; this plan analyzes turns *without* conversation-
  position bias, a different axis from session summaries.
- Not backfilling `turn_index` into a stored column in Phase 0 — computing it
  at query time keeps it always consistent with whatever ordering exists,
  needs no migration, and is cheap at this dashboard's current scale (25k
  trace cap, see `lib/config.ts`). Revisit only if `EXPLAIN` shows the window
  function is a bottleneck.

## 5. Effort & sequencing summary

| Phase | Where | Effort | Unlocks |
|---|---|---|---|
| 0 `turn_index` + filters on `/api/traces` | project-zeno `traces.py` | ½–1 day | position-independent turn queries, retires client-side session reconstruction |
| 1 per-turn dataset/insight diffs | project-zeno `parse.py` + `traces.py` | 1 day | honest per-turn vs. cumulative signals |
| 2 turn-position filters + `group_by` on `/api/traces/analytics` | project-zeno `traces.py` | 1–2 days | single-call turn-depth trend, apples-to-apples aggregates |
| 3 client integration | this repo | ½–1 day | dashboard consumes `turn_index`, drops reconstruction hacks, new turn-depth chart |

------------ END AI PLAN -----

I have mostly followed the plan from the above document, with one significant change:

 - We don't compute the turn index at query time view the WINDOW function, but rather just add it as a column to the db, add it to the ingestion scripts and will run a backfill: I was a bit worried about the performance implications of doing that at query time as the dataset grows - creating separate fields makes it much easier to index, etc.
 
 @01painadam here's a note back from my agent to yours :D that should document all the changes and what needs to be done on the `tracey` side:
 
 # Turn-level analytics — backend is landed

Hi Adam,

The turn-position-aware analytics from your proposal are implemented on the traces API
(branch `feat/traces-per-turn`). Summary of what you can now consume, how to try it, and
where our implementation differs slightly from the doc.

## What's new

**`turn_index`** — the 1-based position of each turn within its session, on every
`TraceListItem` / `TraceDetail`. It's a **stored column**, so it's the *true* in-session
position even when a date range or pagination cuts a thread mid-way (the client-side
reconstruction bug you flagged is gone).

**Turn-position filters** on both `GET /api/traces` and `GET /api/traces/analytics`:
- `turn_index=N` (exact), `min_turn_index`, `max_turn_index`, `first_turn_only=true`.

**Honest per-turn fields** alongside the existing cumulative ones, on the list/detail:
- `insight_created_this_turn` — true only on the turn an insight first appears.
- `datasets_analysed_this_turn` — datasets *new to this turn* (this turn's cumulative
  set minus the previous turn's).
- The cumulative fields you already use (`datasets_analysed`, `has_insight`,
  `primary_dataset_name`) are unchanged, but their OpenAPI descriptions now flag them
  explicitly as *thread-cumulative-as-of-this-turn* so the schema stops being ambiguous.

**New endpoint `GET /api/traces/analytics/by-turn`** — your `group_by=turn_index` idea,
as a first-class endpoint. One call returns the same scalar metrics (counts, latency
avg/p50/p95, cost, tokens, tool usage) **broken down per turn position**, so "how do
turns evolve within a session" is a single request. Shape:

```jsonc
GET /api/traces/analytics/by-turn?turn_bucket_cap=10
{
  "turn_bucket_cap": 10,
  "ungrouped_traces": 0,          // rows with no turn_index (reconciliation)
  "grand_total": { "total_traces": 8200, "latency": {…}, "cost": {…}, … },
  "groups": [
    { "turn_index": 1, "is_terminal": false, "turn_index_min": 1, "turn_index_max": 1,
      "sessions_reaching": 2000, "total_traces": 4200, "latency": {…}, … },
    { "turn_index": 2, "is_terminal": false, … },
    …
    { "turn_index": 10, "is_terminal": true, "turn_index_min": 10, "turn_index_max": 37,
      "sessions_reaching": 120, … }   // terminal bucket: positions >= cap collapsed here
  ]
}
```

Notes on `by-turn`:
- Positions at/above `turn_bucket_cap` (default 10, max 50) collapse into one **terminal
  bucket**, so a 40-turn outlier session doesn't produce 40 rows. `turn_index_min`/`max`
  give the true range within a bucket; `is_terminal` marks the collapsing one.
- `sessions_reaching` is a **funnel** metric (distinct sessions that reached that depth)
  — it's non-increasing and does **not** sum across buckets.
- The cumulative sums (`cost.total`, `tokens.total_turn_tokens`) are within-bucket
  totals — compare **averages/percentiles** across buckets, not the sums.
- All the turn/env/date filters compose with it (filter first, then bucket).

## How to test

- Compare turn 1 vs later turns with two analytics calls:
  - `GET /api/traces/analytics?first_turn_only=true`
  - `GET /api/traces/analytics?min_turn_index=2`
- Or in a single call: `GET /api/traces/analytics/by-turn`.
- Confirm ordinals hold under a date filter that excludes early turns:
  - `GET /api/traces?session_id=<multi-turn>&min_turn_index=2`
- Per-turn signal spot-check: `GET /api/traces/{id}` → `insight_created_this_turn` /
  `datasets_analysed_this_turn` reflect only that turn's delta.

## Docs

Everything is in the live OpenAPI schema (`/docs`, `/openapi.json`) — every new param
and field carries a description, including the cumulative-vs-per-turn distinction.

## Where we diverged from the proposal (all deliberate)

1. **`turn_index` is a stored, backfilled column populated at ingest — not a query-time
   view/CTE window.** Same correct-under-filter guarantee you wanted, but a window
   function's output can't be filtered by an index, so computing it per request would
   force a full-table sort on both hot endpoints. We pay it once at write; it's sargable
   and indexed.
2. **No `is_first_turn` field** — it's exactly `turn_index == 1`, so we dropped it. Use
   `turn_index=1` or the `first_turn_only=true` alias.
3. **Null-session traces** are singleton threads via `COALESCE(session_id, id)` →
   `turn_index=1`, rather than a special case.
4. **The per-turn diff fields are computed cross-row at ingest** (comparing against the
   previous turn), not in the per-trace parser.
5. **`group_by=turn_index` is a dedicated `/analytics/by-turn` endpoint, not a query
   param** — a param that flips the response shape is awkward for typed/OpenAPI clients.
   The endpoint also adds honest terminal-bucket semantics (`turn_index_min`/`max` +
   `is_terminal`), an `ungrouped_traces` reconciliation count, and the `sessions_reaching`
   funnel beyond the original sketch.
6. **Out of scope (yours):** the `tracey` client integration.

Let me know if any field names or semantics don't fit how `tracey` wants to consume
them — easy to adjust before this merges.

----

### DIFF
--- a/db/alembic/versions/b1f7a3c9d2e5_add_turn_index_to_langfuse_traces.py
+++ b/db/alembic/versions/b1f7a3c9d2e5_add_turn_index_to_langfuse_traces.py
@@ -1,0 +1,137 @@
+"""add turn_index to langfuse_traces
+
+Revision ID: b1f7a3c9d2e5
+Revises: a1b2c3d4e5f6
+Create Date: 2026-07-03 10:00:00.000000
+
+Hand-written (alembic autogenerate is disabled). Adds a stored per-turn ordinal
+(``turn_index``, 1-based position within a session ordered by trace_timestamp) so
+turn-position analytics is index-filterable rather than a per-request window, and
+starts populating the previously-unused ``is_final_turn_in_thread`` column.
+
+Session-less traces (null session_id) are singleton threads via
+COALESCE(session_id, id) -> turn_index 1, is_final True. Ongoing rows are kept
+current by a session-scoped recompute in the ingest path; this migration backfills
+existing rows once.
+
+The ``langfuse_traces_analytics`` view is rewritten to read the *stored*
+``is_final_turn_in_thread`` (dropping its own row_number() window) and to expose
+``turn_index``.
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "b1f7a3c9d2e5"
+down_revision: Union[str, None] = "a1b2c3d4e5f6"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+_VIEW = "langfuse_traces_analytics"
+
+# Base column projection shared by both the new and old view definitions.
+_VIEW_BASE_COLS = """
+    id,
+    session_id,
+    user_id,
+    environment,
+    trace_timestamp,
+    outcome,
+    has_answer,
+    answer_is_refusal,
+    had_tool_call,
+    tool_error_count,
+    aoi_name,
+    aoi_type,
+    primary_dataset_name,
+    has_insight,
+    is_global,
+    insight_id,
+    turn_input_tokens,
+    turn_output_tokens,
+    turn_tokens,
+    turn_tool_calls,
+    latency_seconds,
+    total_cost,
+    prompt"""
+
+# New: read the stored is_final_turn_in_thread and expose turn_index (both now
+# maintained on the base table), so the view is a plain projection with no window.
+_VIEW_NEW = f"""
+CREATE OR REPLACE VIEW {_VIEW} AS
+SELECT{_VIEW_BASE_COLS},
+    is_final_turn_in_thread,
+    turn_index
+FROM langfuse_traces;
+"""
+
+# Old: is_final_turn_in_thread computed via a row_number() window, no turn_index.
+_VIEW_OLD = f"""
+CREATE OR REPLACE VIEW {_VIEW} AS
+SELECT{_VIEW_BASE_COLS},
+    (
+        row_number() OVER (
+            PARTITION BY COALESCE(session_id, id)
+            ORDER BY trace_timestamp DESC NULLS LAST, id DESC
+        ) = 1
+    ) AS is_final_turn_in_thread
+FROM langfuse_traces;
+"""
+
+# Backfill turn_index (1-based, ascending) and is_final_turn_in_thread for every
+# row. COALESCE(session_id, id) makes null-session rows singleton threads.
+_BACKFILL = """
+WITH ranked AS (
+    SELECT id,
+           row_number() OVER w AS rn,
+           count(*)     OVER (PARTITION BY COALESCE(session_id, id)) AS n
+    FROM langfuse_traces
+    WINDOW w AS (
+        PARTITION BY COALESCE(session_id, id)
+        ORDER BY trace_timestamp ASC NULLS LAST, id ASC
+    )
+)
+UPDATE langfuse_traces t
+SET turn_index = ranked.rn,
+    is_final_turn_in_thread = (ranked.rn = ranked.n)
+FROM ranked
+WHERE t.id = ranked.id;
+"""
+
+
+def upgrade() -> None:
+    op.add_column(
+        "langfuse_traces", sa.Column("turn_index", sa.Integer(), nullable=True)
+    )
+    op.execute(_BACKFILL)
+    op.create_index(
+        "ix_langfuse_traces_turn_index", "langfuse_traces", ["turn_index"]
+    )
+    # Serves the common "first turns, newest first" read directly (no sort).
+    op.create_index(
+        "ix_langfuse_traces_first_turn",
+        "langfuse_traces",
+        [sa.text("trace_timestamp DESC")],
+        postgresql_where=sa.text("turn_index = 1"),
+    )
+    # View now reads the stored is_final_turn_in_thread + exposes turn_index.
+    op.execute(_VIEW_NEW)
+
+
+def downgrade() -> None:
+    # The old view has fewer columns than the new one, so REPLACE can't shrink it.
+    op.execute(f"DROP VIEW IF EXISTS {_VIEW};")
+    op.execute(_VIEW_OLD)
+    op.drop_index(
+        "ix_langfuse_traces_first_turn", table_name="langfuse_traces"
+    )
+    op.drop_index(
+        "ix_langfuse_traces_turn_index", table_name="langfuse_traces"
+    )
+    # is_final_turn_in_thread was unpopulated before this migration.
+    op.execute("UPDATE langfuse_traces SET is_final_turn_in_thread = NULL;")
+    op.drop_column("langfuse_traces", "turn_index")
--- a/db/alembic/versions/b1f7a3c9d2e5_add_turn_index_to_langfuse_traces.py
+++ b/db/alembic/versions/b1f7a3c9d2e5_add_turn_index_to_langfuse_traces.py
@@ -6,12 +6,14 @@ Create Date: 2026-07-03 10:00:00.000000
 
 Hand-written (autogenerate disabled). Adds the stored per-turn ordinal ``turn_index``
 (1-based within a session by trace_timestamp) so turn-position analytics is
-index-filterable rather than a per-request window, and starts populating
-``is_final_turn_in_thread``. Null-session traces are singleton threads via
-COALESCE(session_id, id) -> turn_index 1. The ``langfuse_traces_analytics`` view is
-rewritten to read the stored flag (dropping its row_number() window) and expose
-``turn_index``. Ongoing rows are kept current by the ingest recompute; this backfills
-existing rows once.
+index-filterable rather than a per-request window, and starts using
+``is_final_turn_in_thread``. The ``langfuse_traces_analytics`` view is rewritten to
+read the stored flag (dropping its row_number() window) and expose ``turn_index``.
+
+Schema only: new rows are populated by the ingest recompute, and existing rows by the
+out-of-band ``backfill-turn-fields`` CLI command (data backfills stay out of the
+blocking deploy migration). Both columns read NULL for existing rows until that command
+runs — endpoints and the view tolerate NULL.
 """
 
 from typing import Sequence, Union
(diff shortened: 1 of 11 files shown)


batpad:
After we deploy this to staging, we just need to run a one-time backfill of data - I deliberately kept it out of the migrations since I don't want anything complex or blocking in the migrations.

So we need to run after deploy:

```
  kubectl exec $(kubectl get pods --no-headers | grep zeno-api | awk '{print $1}' | head -1) -- \
    uv run python src/api/cli.py backfill-turn-fields
```

01painadam (approved):
Functionally, looks good to me @batpad

yellowcap (approved):
let's roll, tracey 🛼

01painadam:
@batpad this is good to ship!

batpad:
Yup, sorry for the delay merging this! I'll resolve these minor conflicts and merge and deploy this today and let y'all know when done.

PR #744, opened 2026-07-08
fix(agent): enforce skill requires: at read_skill call time

### LINKED ISSUE #743: read_skill has no runtime check against a skill's requires: tools
## Context

Skills (`src/agent/skills/skills_md/*.md`) declare a `requires:` frontmatter field listing the tool names their workflow needs (e.g. `dashboard.md` requires `create_dashboard, add_to_dashboard, add_map_widget`). Today this is only enforced once, at prompt-generation time: `AgentConfig.skills()` (`src/agent/agent_config.py:93-95`) filters which skills get *listed* in the system prompt to those whose `requires:` are all bound in that profile.

The `read_skill` tool (`src/agent/skills/tool.py:10-18`, backed by `get_skill_body` in `src/agent/skills/loader.py:62-68`) does **no equivalent check**. It's a plain `name -> body` lookup and will return any skill's full body for any name, regardless of whether the calling agent's profile actually has the tools that skill's workflow depends on.

There's no live bug today — `create_dashboard`/`add_to_dashboard`/`add_map_widget` are only ever registered together in `EXPERIMENTAL_SPECS`, never split across a smaller profile (`src/agent/agent_config.py:55-65`). But it's a latent hole: if a future profile ever binds some-but-not-all of a skill's required tools, or a model hallucinates/guesses a skill name, `read_skill` will happily hand back workflow instructions for tools the model can't actually call, and the failure only surfaces later when it tries to invoke the missing tool.

## Proposed fix

Enforce the same `requires:` check at `read_skill` call time, not just at prompt-build time. Since `AgentConfig` is resolved once per request in `fetch_zeno` (`src/agent/graph.py:228`) and then only used to build the graph/prompt (never threaded into tool-call scope), follow the existing precedent set by `current_user_id` (`src/shared/request_context.py`, added in `fa7d36e` "fix(agent): raise when a tool runs without a bound user id"): a `contextvars.ContextVar` set once per request and read from inside the tool.

Concretely:

1. `src/agent/tool_spec.py` — already a dependency-free "leaf module" imported by both `agent_config.py` and every tool file. Add a `ContextVar[frozenset[str]]` for the bound tool names of the current request, plus `set_bound_tool_names()` / `bound_tool_names()` accessors.
2. `src/agent/agent_config.py` — factor the `{s.tool.name for s in self.specs}` computation (currently inline in `.skills()`) into an `AgentConfig.tool_names()` method, reused by both `.skills()` and the new binding call.
3. `src/agent/graph.py` (`fetch_zeno`) — right after `config = registry.resolve(ff)`, call `set_bound_tool_names(config.tool_names())`. Confirmed this runs once per request in the same coroutine that later invokes the graph (`src/api/services/chat.py:140` -> `:197`, no task handoff), matching the lifecycle guarantee `current_user_id` relies on.
4. `src/agent/skills/loader.py` — add `get_skill(name) -> SkillMeta | None` so the tool can inspect `requires`, not just fetch the body.
5. `src/agent/skills/tool.py` (`read_skill`) — look up the skill's `requires`, diff against `bound_tool_names()`, and return a clear "skill not available in this profile (missing tools: ...)" message when something's missing, instead of raising into the tool-error funnel (unlike the identity check, a missing-tool skill request is a normal "not applicable here" case, not a broken-invariant one).

## Verification

- Unit test: a profile missing one of a skill's `requires:` tools gets a refusal message from `read_skill`, not the skill body.
- Unit test: a profile with all `requires:` tools bound gets the normal body back (regression check against current behavior).
- Existing `AgentConfig.skills()` prompt-filtering tests stay green — no change to what gets listed, only to what `read_skill` will actually serve.

yellowcap:
## Summary

- `read_skill` was a plain `name -> body` lookup with no check against the calling agent's actually-bound tools, unlike `AgentConfig.skills()` which already filters what gets *advertised* in the system prompt (#743).
- Bind the resolved profile's tool names to a per-request `ContextVar` in `fetch_zeno` (mirroring the existing `current_user_id` pattern in `src/shared/request_context.py`), and have `read_skill` refuse when a skill's `requires:` aren't all bound. The refusal message is identical to the "unknown skill name" case (`"skill not found: {name}"`) — a skill excluded from this profile's advertised list is, from the model's perspective, indistinguishable from one that doesn't exist; the specific missing tools are only logged, not surfaced to the model.
- **Follow-up caught in review**: two prompt-building paths still named skills unconditionally, so the model could be routed straight into a `read_skill` call that gets refused — reproducing exactly this in a production trace where the agent asked for the `dashboard` skill immediately on a profile that doesn't bind its required tools:
  - `get_prompt()`'s `# Routing` table (`graph.py`) was a hardcoded string naming `dashboard`, `show-imagery`, `pull-data`, `analyze`, `capabilities`, and `search_insights` regardless of the active profile.
  - `view_pages.py`'s dashboard `# Current surface` section had the same `read skill \`dashboard\`` pointer, gated only by which frontend page the user is on, not by the bound profile.
- **Hardening pass**: the first cut of that prompt gating checked names against a single flat set mixing skill and tool names — correct only by the accident that skills are hyphenated (`show-imagery`) and tools snake_case (`show_imagery`). Availability is now a typed, two-namespace concept, gates declare which namespace they mean, and new guardrail tests turn every silent failure mode (name collision, renamed skill/tool dropping prompt content, typo'd `requires:` hiding a skill forever) into a test failure.

No behavior change for any currently-registered profile beyond the routing text: `create_dashboard`/`add_to_dashboard`/`add_map_widget` are only ever bound together today, so this closes a latent hole plus the prompt-side symptom of it, rather than fixing a different live bug. Rendered system prompts were verified **byte-identical** across both profiles and all registered pages before/after the hardening and simplification commits.

Closes #743

### Call-time enforcement (`read_skill`)
- `src/agent/tool_spec.py` — `ContextVar[frozenset[str]]` + `set_bound_tool_names()` / `bound_tool_names()` accessors
- `src/agent/graph.py` (`fetch_zeno`) — binds `set_bound_tool_names(config.tool_names())` once per request
- `src/agent/skills/loader.py` / `__init__.py` — added `get_skill(name)` to expose `SkillMeta` (including `requires`); `render_body(skill)` factored out and shared by `get_skill_body` and `read_skill` (single lookup, no assert)
- `src/agent/skills/tool.py` (`read_skill`) — checks `requires` against `bound_tool_names()`; returns the generic "not found" message when something's missing

### Prompt-side gating, typed namespaces
- `src/agent/tool_spec.py` — `Availability(skills, tools)` frozen dataclass with `has_skill` / `has_tool` / `allows(gate)`, and a `Gate` type alias (`tuple[Literal["skill", "tool"], str] | None`) so a misspelled gate kind is a **mypy error at the definition site**, not a silently-dropped prompt row
- `src/agent/agent_config.py` — `AgentConfig.tool_names()` (reused by `.skills()`) and `AgentConfig.availability()`, the one place the availability set is computed; `AgentConfigRegistry.configs()` public accessor for cross-profile invariant checks
- `src/agent/graph.py` — the `# Routing` table is a module-level `ROUTING_ROWS` constant of `(Gate, line)` pairs; `get_prompt()` renders only rows the profile's `Availability.allows()`
- `src/agent/view_pages.py` — `ViewPage` carries `prompt_base` plus a declarative `skill_pointer: (skill, suffix)` instead of prompt-section closures; a generic `render_prompt(available)` appends the pointer only when `has_skill` passes. `prompt_section(page, available)` now requires the availability argument (the empty default existed only for tests)

### Guardrail tests (`tests/unit/agent/test_availability.py`, new)
- skill names never collide with tool names across all registered profiles
- every `ROUTING_ROWS` gate resolves to a real skill/tool
- every `ViewPage.skill_pointer` names a real skill
- every skill's `requires:` frontmatter entry names a registered tool

Each was proven to bite by injecting a typo and watching the test fail.

## Test plan

- [x] `read_skill` refuses (with the generic "not found" message) when a required tool is unbound, when no tools are bound, and when the skill name is simply unknown — all indistinguishable to the model
- [x] `read_skill` serves the body when all required tools are bound, and when bound tools are a strict superset; skills with no `requires:` (e.g. `capabilities`) are always served
- [x] The missing-tools detail is captured in the warning log but not in the returned string (verified via a log-capturing test)
- [x] `AgentConfig.tool_names()`, the `bound_tool_names` ContextVar accessors, and `fetch_zeno` binding the resolved profile's tool names
- [x] `get_prompt()`'s Routing table excludes rows for skills/tools the profile doesn't bind (default profile) and includes them when it does (experimental profile)
- [x] `view_pages.py`'s dashboard surface section drops the `dashboard` skill pointer when unavailable, keeps it when available; a *tool* named like a skill does **not** satisfy a skill gate
- [x] The four availability guardrail tests pass, and each fails when its invariant is deliberately broken
- [x] Rendered prompts byte-identical to the pre-refactor output across (default, experimental) × (no page, map, dashboard, unknown page)
- [x] `uv run pytest tests/unit -q` — 352 passed
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on touched files (two pre-existing, unrelated errors on `graph.py` confirmed present on `main` too)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.7"
+version = "2026.7.8"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -121,6 +121,10 @@ class AgentConfigRegistry:
     def register(self, config: AgentConfig) -> None:
         self._configs[config.name] = config
 
+    def configs(self) -> tuple[AgentConfig, ...]:
+        """All registered configs, e.g. for invariant checks across profiles."""
+        return tuple(self._configs.values())
+
     def resolve(self, ff: Optional[str] = None) -> AgentConfig:
         """Return the config for ``ff``, falling back to the default."""
         if ff and ff in self._configs:
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -90,8 +90,11 @@ class AgentConfig:
                 blocks.append(category + "\n\n" + "\n".join(fragments))
         return "\n\n".join(blocks)
 
+    def tool_names(self) -> frozenset[str]:
+        return frozenset(s.tool.name for s in self.specs)
+
     def skills(self) -> list[SkillMeta]:
-        available = {s.tool.name for s in self.specs}
+        available = self.tool_names()
         return [sk for sk in all_skills() if set(sk.requires) <= available]
 
 
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -24,7 +24,7 @@ from src.agent.subagents.analyst.tool import SPEC as generate_insights_spec
 from src.agent.subagents.pick_aoi.tool import SPEC as pick_aoi_spec
 from src.agent.subagents.pick_dataset.tool import SPEC as pick_dataset_spec
 from src.agent.subagents.search.blog import SPEC as search_blogs_spec
-from src.agent.tool_spec import ToolCategory, ToolSpec
+from src.agent.tool_spec import Availability, ToolCategory, ToolSpec
 from src.agent.tools.add_map_widget import SPEC as add_map_widget_spec
 from src.agent.tools.add_to_dashboard import SPEC as add_to_dashboard_spec
 from src.agent.tools.create_dashboard import SPEC as create_dashboard_spec
@@ -96,6 +96,16 @@ class AgentConfig:
     def skills(self) -> list[SkillMeta]:
         available = self.tool_names()
         return [sk for sk in all_skills() if set(sk.requires) <= available]
+
+    def availability(self) -> Availability:
+        """What this profile can route to — the same rule read_skill enforces
+        at call time (a skill's ``requires`` must all be bound), so prompt
+        builders never advertise a skill or tool the model would just get
+        "not found"/refused for."""
+        return Availability(
+            skills=frozenset(sk.name for sk in self.skills()),
+            tools=self.tool_names(),
+        )
 
 
 class AgentConfigRegistry:
--- a/src/agent/graph.py
+++ b/src/agent/graph.py
@@ -43,14 +43,77 @@ def get_prompt(
     """
     if config is None:
         config = default_registry.resolve()
+    skills = config.skills()
     skill_lines = [
         f"- {s.name}: {s.description} (use when: {s.when_to_use})"
-        for s in config.skills()
+        for s in skills
     ]
     skills_block = "\n".join(skill_lines) if skill_lines else "(none)"
     tool_descriptions = config.tool_descriptions()
-    surface = prompt_section(page)
+    # What this profile can actually route to — same set read_skill checks
+    # requires: against, so a row is never printed for a skill/tool the
+    # model would immediately get "not found"/refused for.
+    available = {s.name for s in skills} | config.tool_names()
+    surface = prompt_section(page, frozenset(available))
     surface_block = f"\n# Current surface\n\n{surface}\n" if surface else ""
+    routing_rows = [
+        (
+            None,
+            '- Dataset-only (e.g. "pick tcl by driver"): call pick_dataset, '
+            "then stop. No AOI, pull or insights unless asked.",
+        ),
+        (
+            None,
+            '- AOI-only (e.g. "zoom to Pará"): call pick_aoi, then stop '
+            "unless asked for more.",
+        ),
+        (
+            "pull-data",
+            '- Pull-only (e.g. "pull dist alerts in Bern for last 2 '
+            'weeks"): read `pull-data`, run pick_aoi → pick_dataset → '
+            "pull_data, then stop. Do not call generate_insights unless "
+            "the user asked for a chart or analysis.",
+        ),
+        (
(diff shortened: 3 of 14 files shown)


srmsoumya:
Some suggestions to think about routing + profile definition (not blocking this PR)

Routing overlaps with skills - per-skill routing rows mostly restate each skill's `when_to_use`, and the two lists drift (e.g. `explore/wri-insights` have skills but no routing row). Only a few lines are truly unique - "pick one, don't escalate," the no-skill shapes (dataset/AOI-only), and stop semantics (recall is terminal). 
We would ideally like to drop routing from the prompt entirely over time - skills + tool descriptions should carry this, this might lead to more prompt-optimization.

Define profiles by skills, derive tools from them - now a profile is a tuple of tools with skills inferred (requires ⊆ tools) — so adding a tool can silently activate an unintended skill. Instead, we list the skills (the capability surface) and derive their tools, plus an extra_tools bag for ad-hoc ones:
```python
DEFAULT      = Profile(skills=("analyze", "pull-data", "capabilities"))
EXPERIMENTAL = Profile(skills=(*DEFAULT.skills, "dashboard", "show-imagery", "wri-insights", "explore"),
                       extra_tools=(inspect_view_context_spec, update_insight_display_spec, search_insights_spec))
```
This will provide explicit control over what each profile exposes, and correct-by-construction (can't list a skill without its tools).

yellowcap:
Thanks for the review @srmsoumya . Going to merge this one and make a new one with your suggested improvements for gating handling.

Issue #735, opened 2026-07-01
PZB-1218: Refactor AOI search to query a single table


batpad:
Context: As we are implementing more features for the frontend UI to interact with AOIs: Search, add to portfolio, etc. and not just fetching them via agent tool calls, it is going to be more important to have these endpoints be performant, well indexed, and easy to manage as we grow out search and filter features. We currently store AOI data across 4 different tables, and searching through involves searching through all tables - and things like sorting get more complex and harder to do while avoiding full table scans across all tables.

If the search were applied against a single table, this would be trivial to index, search, filter and sort across multiple axes. Below is some context, justification and plan for the refactor output by Claude:

# AOI Search: Recommendation & Plan for a Unified Search Table

## Context

The `feat/searchable-aois` branch (commit `e1fd611`) added `GET /api/aois`, a fuzzy AOI
search endpoint. It reuses `search_aois()` in `src/shared/geocoding_helpers.py:141`, which
searches across **five heterogeneous tables** by building a per-source `SELECT`,
`UNION ALL`-ing them into a CTE, and globally ranking by trigram
`similarity(LOWER(name), LOWER(:name))`.

The question this document answers: should the fields we **search and filter on** live in a
single table (with record *details* still federated across the source tables), or is the
current multi-table union approach fine?

**Recommendation: refactor to a single denormalized `aoi_search` table.** Merge the current
PR **as-is**; land the restructure as a **follow-up** that keeps the public API and agent
behavior unchanged.

---

## What exists today

Five AOI tables in two very different classes:

| | Reference tables (`geometries_gadm`, `_kba`, `_wdpa`, `_landmark`) | `custom_areas` |
|---|---|---|
| Managed by | Ingest scripts via `to_postgis(if_exists="replace")` — **no ORM, no Alembic** | SQLAlchemy ORM + Alembic |
| Geometry | Native PostGIS `geometry` (SRID 4326) | `geometries` JSONB (list of GeoJSON strings) |
| ID column | Heterogeneous: `gadm_id`, `sitrecid` (numeric), `wdpa_pid`, `landmark_id` | `id` (UUID) |
| Name index | GIN trigram on `name` | **None** (user-scoped, so small) |
| Scope | Global / static | Per-user, dynamic (CRUD at runtime) |

**Search** (`search_aois`, `src/shared/geocoding_helpers.py:141`) unifies these at query time:
it probes which tables exist, builds one `SELECT` per source, `UNION ALL`s them into a CTE,
[...]

PR #746, opened 2026-07-09
Feat/unified aoi table

batpad:
Refs #735 

Creates schema for a new unified AOIs table that brings together AOIs from KBA, GADM, etc. + Custom AOIs into a single AOIs table.

The scope of this PR is:

 - Create new unified AOI table
 - Write script to copy data from current AOI tables to new AOI table

It is purely additive in the database and does not touch APIs. The goal would be to merge this, run the schema and data migration and verify that all data has been copied over to the unified table correctly. Then we follow up with a PR to move the APIs, etc. to use the new table and deprecate the old tables.

cc @yellowcap @01painadam

### DIFF
--- a/db/alembic/versions/ceea2a027738_create_unified_aoi_tables.py
+++ b/db/alembic/versions/ceea2a027738_create_unified_aoi_tables.py
@@ -44,11 +44,16 @@ def upgrade() -> None:
         sa.Column("source_id", sa.String(), nullable=False),
         sa.Column("name", sa.String(), nullable=False),
         sa.Column("subtype", sa.String(), nullable=False),
+        # Typed MULTIPOLYGON: every AOI is one areal shape, so the typmod
+        # rejects any non-MultiPolygon / non-4326 insert. build-aois normalizes
+        # source geometry (ST_MakeValid + dissolve + ST_Multi) to satisfy this.
         # spatial_index=False: the GiST index is created explicitly below so
         # its name is controlled and not auto-emitted by geoalchemy2.
         sa.Column(
             "geometry",
-            Geometry(geometry_type="GEOMETRY", srid=4326, spatial_index=False),
+            Geometry(
+                geometry_type="MULTIPOLYGON", srid=4326, spatial_index=False
+            ),
             nullable=False,
         ),
         # [west, south, east, north]; precomputed, antimeridian-aware.
--- a/db/alembic/versions/ceea2a027738_create_unified_aoi_tables.py
+++ b/db/alembic/versions/ceea2a027738_create_unified_aoi_tables.py
@@ -136,8 +136,6 @@ def upgrade() -> None:
             "relationship",
             sa.Enum(
                 "owner",
-                "shared_edit",
-                "shared_view",
                 "saved",
                 name="aoi_relationship",
                 native_enum=False,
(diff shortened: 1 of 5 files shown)


batpad:
Ok, I think this is ready to review. There are a few schematic decisions encapsulated in this PR, and while I think this is the best path forward, I think each of these has some trade-offs - 

 - **Use a single table for all AOIs**: Bring all the different AOI tables (kba, gadm, etc.) as well as Custom Areas into a single, unified table. Cusom Areas keep a record of who uploaded them via an `uploaded_by` field, but all permissions between AOIs and Users are handled by a `user_aois` table that joins users to AOIs via a `relationship` ENUM. This will allow us flexibility of granting permissions, etc. across AOIs as well as handling things like "Saved AOIs"
  - **Coerce all geometries to Multipolygon**: We currently handle AOI geometries differently between kba, gadm, etc. and Custom AOIs. In KBA, GADM, we have a mix of Polygon and MultiPolygon geoms, and Custom AOIs are stored as a list of geometries (which we coerce to MultiPolygon when we return). This PR proposes standardizing everything to always be a MultiPolygon. Single Polygons are made MultiPolygons with a single member, and invalid geometries throw errors. Enforcing a MultiPolygon type on the schema seems like it should increase predictability and reduce branching in code that's reading these geometries, and more strictly enforce we have valid geometries.

I think these are really the two big decision points, and it would be good to get any thoughts / disagreements / questions. I worked with AI to make some HTML with the various options and pros and cons, mostly for myself as I worked through things and schema options. I did update it with the final implementation, so can see some more details of the plan and things considered here (with the warning that its heavily AI-generated): https://sanjay.bhangar.com/tmp/aoi-architecture/

Note: This PR does NOT deal with versioning of AOIs. This was considered, and I feel it's easy to add a concept of versioning once we have this table structure consolidated. So, ideally, we can continue the versioning discussion and add it as an additive PR once we have a clearer idea of how we want to deal with versioning, but not let it block this restructure.

If this consolidated / unified schema works for everyone and there's no major disagreement, next steps here would be:

 - Merge this PR and run the `build_aois` CLI command on staging - verify that the new table is created correctly, all geometries look correct, etc.
  - Work on a separate PR to migrate all the API endpoints and agent search to use the new table, ensuring backwards compatibility for all frontend-facing APIs
  - Ingestion scripts: look at the current ingestion scripts for GADM, etc. - we can continue to keep ingestion a two-step process -> i.e. ingestion script creates like a `gadm` table as it does currently, and then `build_aois` copies things into the unified table - this will allow us to go ahead with minimal changes to the ingestion scripts.
  - Implement new features like Saved Features, etc. with the new schema and structure ...

@yellowcap @01painadam - please tag whoever should review this and provide input. This is a fairly big schema change for our AOI handling, but should not represent anything backwards incompatible in terms of interface - the intention is to consolidate the structure before adding a lot of AOI-related features, and importantly, bring items that need to be searched, sorted, etc. into a single table so that can be easily optimized.

batpad:
@sunu @geohacker - just tagging y'all here in case you have strong opinions about any of this stuff, especially the geometry handling (coercing to MultiPolygon, etc... )

geohacker (approved):
This _looks_ good to me @batpad

yellowcap (approved):
Conceptually looks great. Feel free to move in when ready from your end.

PR #747, opened 2026-07-09
Add dataset_id and aoi filters to GET /api/insights

yellowcap:
## Summary
- Adds `dataset_id` and `aoi_source` + `aoi_id` query filters to `GET /api/insights`, alongside the existing `thread_id` filter.
- Filters resolve against `StatisticsOrm` rows linked via `InsightOrm.statistics_ids`, matching on the `dataset_id` and parallel `aoi_ids`/`aoi_sources` JSONB arrays added in #729.

## Details
- `aoi_source` and `aoi_id` must be passed together (400 if only one is given) — `src_id` is only unique per source.
- The AOI pair is matched pairwise by index (unnest `aoi_ids` `WITH ORDINALITY`, index into `aoi_sources`), not via independent membership checks — this avoids cross-matching an id from one source against an unrelated source in the same stats row.
- The statistics linkage is a correlated `EXISTS` against `statistics_ids`; `dataset_id` and the AOI pair combine with AND semantics and narrow within the caller's own insights.

Example:
```
GET /api/insights?dataset_id=42
GET /api/insights?aoi_source=gadm&aoi_id=IDN.24_1
GET /api/insights?dataset_id=42&aoi_source=gadm&aoi_id=IDN.24_1
```

## Test plan
- [x] `pytest tests/api/test_insights.py` — 29 passed, including 8 new tests covering: dataset_id match/no-match, aoi pair match, pairwise cross-match regression (index-zipping correctness), missing-half-of-pair 400s, combined AND filtering, and per-user scoping.
- [x] `ruff check` / `ruff format --check` / `mypy` clean (no new mypy errors vs. baseline).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.8"
+version = "2026.7.9"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/api/routers/insights.py
+++ b/src/api/routers/insights.py
@@ -4,12 +4,12 @@ from typing import Optional
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, status
-from sqlalchemy import select
+from sqlalchemy import String, cast, func, select, text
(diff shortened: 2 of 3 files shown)


jterry64 (approved):
LGTM, thanks!

PR #748, opened 2026-07-13
PZB-1095: Add agent tools to create and edit dashboard text widgets

yellowcap:
## Summary

Text widgets (`widget_type="text"`, markdown in `config.text`) already exist end-to-end via REST, but the agent had no way to create them. This adds the agent path with two new experimental primitives, following the same pattern as the other dashboard tools (resolve dashboard → owner check → write → `dashboard_updated` signal):

- **`add_text_widget(text, dashboard_id?, position?)`** — composes a markdown note onto a dashboard. Config is exactly `{"text": ...}`, the shape `validate_text_config` enforces on the REST path. The success reply surfaces the new widget id so the agent can reference it for later edits. No `title` parameter — headings go in the markdown, per the frontend contract.
- **`edit_text_widget(text, widget_id?, dashboard_id?)`** — replaces the full markdown body of an existing text widget. Without a `widget_id` it targets the resolved dashboard's only text widget; with several, the error lists each candidate's id plus an excerpt so the model can retry with an explicit id. With a `widget_id`, ownership is checked against the widget's own dashboard, and non-text widgets are rejected.

Supporting changes:

- New `dashboard_writer.get_widget(widget_id)` lookup (malformed id → `None`, matching module conventions), needed for the ownership check on explicit widget ids.
- Both SPECs registered in `EXPERIMENTAL_SPECS`, so the tools ship behind the `experimental` feature flag like the other dashboard primitives.

## Test plan

- [x] Unit tests per helper (no DB): `_normalize_text`, `_widget_config`, `_select_text_widget`, `_excerpt` in `tests/unit/agent/tools/`
- [x] `get_widget`: malformed-id contract in the writer unit tests; found/unknown-UUID round-trip in `tests/agent/test_dashboards_db.py`
- [x] Tool-level tests (`tests/agent/test_add_text_widget.py`, `test_edit_text_widget.py`): happy paths, empty text, dashboard resolution fallbacks (state / view_context), owner-only denial, ambiguity, non-text widget, mid-write disappearance
- [x] `ruff check` / `ruff format` clean; full unit suite (375 tests) passes
- [x] Manual smoke test via CLI: `uv run python src/agent/cli.py -i --ff experimental` — create dashboard, add note, edit note

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.9"
+version = "2026.7.13"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -49,8 +49,10 @@ from src.agent.subagents.pick_dataset.tool import SPEC as pick_dataset_spec
 from src.agent.subagents.search.blog import SPEC as search_blogs_spec
 from src.agent.tool_spec import Availability, ToolCategory, ToolSpec
 from src.agent.tools.add_map_widget import SPEC as add_map_widget_spec
+from src.agent.tools.add_text_widget import SPEC as add_text_widget_spec
 from src.agent.tools.add_to_dashboard import SPEC as add_to_dashboard_spec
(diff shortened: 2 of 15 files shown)


srmsoumya (approved):
LGTM.

PR #749, opened 2026-07-14
Persist mosaic build metadata in an S3 sidecar and report observed cloud cover

01painadam:
## Summary

A few changes to support the FE. `create_sentinel2_mosaic` computes the scene count and acquired date range at build time but never persists them, so any cache hit (mosaic already in S3) returns a bare `MosaicResult`. Downstream that means:

- `show_imagery` reports "imagery layer created" without scene/date context on reruns of the same recipe, and
- `add_map_widget` snapshots an `ImageryState` with `item_count` / `date_start` / `date_end` as `null` **permanently** into the dashboard widget config — the frontend's imagery legend (wri/project-zeno-next#593) then has nothing to show for the most useful fields.

### Changes

1. **Metadata sidecar** — the build now writes `{token}.meta.json` next to the MosaicJSON with `item_count`, `date_start`, `date_end` and `mean_cloud_cover`. Cache hits read it back, so the build-time stats survive. Degradation is explicit and safe:
   - sidecar absent (mosaics written before this change) → bare result, exactly as today
   - sidecar unreadable or malformed → bare result + warning log, never an error
   - sidecar write failure → logged, not raised (the mosaic itself is unaffected)
   - The mosaic object remains the cache marker; the sidecar is written after it, so a concurrent hit in that window just degrades to a bare result.

2. **Observed cloud cover** — the mean `eo:cloud_cover` across the mosaic's scenes is now computed at build time and threaded through `MosaicResult`, the `/mosaic/create` response, `ImageryState` and `_IMAGERY_KEYS`. This is what the imagery *actually looks like*, as opposed to the recipe's `max_cloud_cover` search threshold, which is all the frontend can currently display ("< 20%").

No changes to recipe tokens, S3 keys of existing mosaics, or the tiler contract.

## Test plan

- [x] `tests/api/test_mosaic.py`: cache hit serves sidecar metadata; missing sidecar → bare result (backwards compat); corrupt sidecar → bare result; mean cloud cover computed and items without `eo:cloud_cover` ignored; endpoint returns metadata on both build and cache hit
- [x] `tests/agent/test_show_imagery.py` / `test_add_map_widget.py`: `mean_cloud_cover` carried in `ImageryState` and the widget config snapshot
- [x] Full `tests/api tests/cli tests/unit tests/agent` run locally: 723 passed; the 7 failures are pre-existing on `main` (agent LLM tests needing API keys, and PostGIS-dependent `test_aois` in my local DB) — verified by running them on a clean `main` checkout
- [x] `ruff check` / `ruff format --check` clean on touched files
- [ ] Verify on staging that a `show_imagery` rerun of a cached recipe now reports scene count/date range

### DIFF
--- a/src/agent/models.py
+++ b/src/agent/models.py
@@ -29,6 +29,13 @@ class ImageryState(BaseModel):
     )
     window_days: int = Field(..., description="Search window in days")
     max_cloud_cover: int = Field(..., description="Max cloud cover percentage")
+    mean_cloud_cover: Optional[float] = Field(
+        None,
+        description=(
+            "Observed mean cloud cover across the mosaic's scenes, percent "
+            "(None when the cached mosaic predates the metadata sidecar)"
+        ),
+    )
     aoi_names: list[str] = Field(
         ..., description="Names of selected areas of interest"
     )
--- a/src/agent/tools/add_map_widget.py
+++ b/src/agent/tools/add_map_widget.py
@@ -42,6 +42,7 @@ _IMAGERY_KEYS = (
     "target_date",
     "window_days",
     "max_cloud_cover",
+    "mean_cloud_cover",
     "aoi_names",
 )
 
--- a/src/agent/tools/show_imagery.py
+++ b/src/agent/tools/show_imagery.py
@@ -146,6 +146,7 @@ async def show_imagery(
         target_date=recipe.target_date.isoformat(),
         window_days=recipe.window_days,
         max_cloud_cover=recipe.max_cloud_cover,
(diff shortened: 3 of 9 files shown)


yellowcap (commented):
I wonder if we need the sidecar files, or if TiTiler mosaics have either a "metadata" field directly, or allow extra fields without failing validators etc. In either of these cases, I would build this into the mosaic directly so we dont need two different files.

@01painadam if you havent tried this, can we try that and see if it could work?

01painadam:
@yellowcap I haven't and yeah I think that would be better, please feel free :)

PR #752, opened 2026-07-15
Pass code executor output to InsightTextGenerator to ground narrative in pre-computed totals

yellowcap:
## Summary

Thread the code executor's printed totals (e.g. `High total: 1306617.82 ha`) into the text generator so the narrative LLM can reference actual pre-computed statistics instead of trying to sum monthly CSV rows itself.

This is a classic "summarization drift" bug: the code executor computed correct totals, the monthly table was correct, but the prose narrative hallucinated `1,147,547.83` instead of `1,306,617.83` because it was forced to do arithmetic on raw chart data it couldn't reliably aggregate.

**Regression likely introduced by #724** (which decoupled narrative generation from the code executor stage).

## Root cause

The pipeline has two independent LLM stages:

| Stage | Model | Input | Output |
|---|---|---|---|
| 1. Code Executor | gemini-3.1-pro | CSV data | Correct totals in stdout (`High total: 1306617.82 ha`) |
| 2. InsightTextGenerator | gemini-3-flash | Chart CSV rows (6 rows) only | Hallucinates total |

The executor's `EXECUTION_OUTPUT` parts (printed totals, statistics) were stored as `codeact_parts` for debugging/trace export but never forwarded to any downstream LLM. `InsightTextGenerator` only saw the 6 rows of monthly chart data and the user query, forcing it to derive totals itself.

## Changes

- **`src/agent/subagents/analyst/text_generator.py`**: Add `executor_context` param and `## Executor findings` section to prompt template. Defaults to `"(none)"` when no context.
- **`src/agent/subagents/analyst/tool.py`**: Collect `EXECUTION_OUTPUT` parts in `_resolve_charts()` and pass them through to the text generator.
- **`tests/unit/agent/tools/test_insight_text_generator.py`**: Add unit tests for executor_context passthrough and default.
- **Integration test mocks**: Update signatures to accept the new kwarg (no behavior change).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.15"
+version = "2026.7.15.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/subagents/analyst/text_generator.py
+++ b/src/agent/subagents/analyst/text_generator.py
@@ -39,6 +39,11 @@ _SYSTEM = """You write the narrative insight for one or two charts that have \
 already been produced from geospatial data. Describe what the data shows — do \
 not invent numbers, and do not redescribe the chart mechanics.
 
+If a total, sum, or other statistic appears in "Pre-computed findings from \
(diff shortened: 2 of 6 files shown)


jterry64 (approved):
Good catch, LGTM!

PR #758, opened 2026-07-22
Persist mosaic build metadata in MosaicJSON extra fields

yellowcap:
**Draft — pending frontend design.** Opening this now so we can confirm the field set/shape lines up with what the frontend needs for the imagery legend (wri/project-zeno-next#593) before finalizing. Field names (especially the observed-vs-threshold cloud cover split) may still change based on that.

## Summary

- `create_sentinel2_mosaic` computes scene count, date range, and cloud cover stats at build time but previously discarded them on a cache hit (mosaic already in S3) — reruns of the same recipe returned a bare result.
- Persists these as extra fields inside the MosaicJSON object itself (the MosaicJSON v0.0.3 spec explicitly allows and requires implementations to tolerate unknown keys), and reads them back on cache hits with a single S3 GET.
- Threads the new fields (`mean_cloud_cover`, `min_cloud_cover`, `max_cloud_cover_observed`) through `MosaicResult` → `MosaicCreateResponse` / `ImageryState` → the `add_map_widget` dashboard snapshot, alongside the existing `item_count`/`date_start`/`date_end`.
- Note: `ImageryState.max_cloud_cover` keeps its original meaning (the search threshold); the new observed max is `max_cloud_cover_observed` to avoid silently changing the meaning of a field already persisted in existing dashboard widget configs.
- Degrades gracefully: mosaics written before this change, or with corrupt/non-object JSON, fall back to a bare result + warning log rather than erroring.

## Test plan
- [x] `pytest tests/api/test_mosaic.py tests/agent/test_show_imagery.py tests/agent/test_add_map_widget.py tests/api/test_dashboards.py` — all passing
- [x] Confirm field names/shape against frontend imagery legend design

## Related to 

- https://github.com/wri/project-zeno/pull/749


yellowcap:
moved the imagery tool into default profile https://github.com/wri/project-zeno/pull/776

yellowcap:
Moving in as per discussion with @01painadam

PR #761, opened 2026-07-22
Conversation language handling

yellowcap:
## Summary
- Language was fragmented: `UserOrm.preferred_language_code` never reached the agent, chat relied on a soft "reply in the same language" prompt line, and insight generation had no language signal at all.
- Adds `src/agent/language.py`: resolves a per-turn language (profile preference > `langid`-detected query language > thread's last-known value) into `AgentState`.
- Surfaces the resolved language explicitly in `SessionContextMiddleware`'s session block, seen by every subagent/tool prompt.
- Threads language through `InsightTextGenerator` and the analyst code-executor prompt so insight narrative, follow-ups, and chart titles are generated in it (data/field identifiers stay English).
- Persists the resolved language on `InsightOrm.language_code` (migration included) and exposes it via `InsightResponse`.
- Includes an unrelated one-line mypy fix (`dashboard_name` arg) that was already broken on `main` and blocked the pre-commit hook.

## Test plan
- [x] `pytest tests/tools/test_generate_insights.py tests/tools/test_generate_insights_tiered.py` — passing (one flaky live-model failure reproduced as a pass in isolation)
- [x] `pytest tests/agent/test_inspect_view_context.py tests/unit/agent/test_view_pages.py` — passing (session block formatting)
- [x] Manual verification of end-to-end non-English conversation → insight generation (not yet done — part of why this is a draft)
- [x] Decide on scope follow-ups: `pick_aoi`'s intentional forced-English place names (left alone), `blog.py` search subagent's separate soft language instruction (not wired to the new mechanism yet)


jterry64 (approved):
LGTM. Did those supported languages come from the PMs or were just AI suggested? If I AI suggested, I'd double check with @01painadam

yellowcap:
The languages were AI selected, but there is a fallback. So if the language is not pre-translated, an llm will be used to dynamically translate into the not-listed language. But defintively @01painadam we can extend this list anytime.

PR #767, opened 2026-07-23
PZB-1146: chore(Middleware): add CORS headers and request_id to api errors

gtempus:
# PR: Improve API error responses for CORS + observability

## Existing behavior

When an unhandled exception occurs (for example, response-model validation failure), the request path can bypass normal CORS header injection.  
Result: the browser reports a CORS error, and the real 500 response body is not visible to frontend code.

The response also does not reliably include a correlation identifier the client can pass back for debugging.

## Behavior after this PR

Unhandled exceptions are converted into a structured JSON `500` response in middleware:

- `detail: "Internal Server Error"`
- `request_id: "<uuid>"`

For cross-origin requests (`Origin` header present), the middleware explicitly adds:

- `Access-Control-Allow-Origin: <origin>`
- `Access-Control-Allow-Credentials: true`
- `Vary: Origin`

This ensures browsers can read the error response instead of masking it as a CORS failure.

## Why `request_id` helps investigations

The same `request_id` is already bound in server logs. Returning it in the error payload lets frontend, QA, or support report one ID that backend engineers can search to quickly reconstruct:

1. request start
2. exception details
3. final error response

That shortens triage time and removes ambiguity when multiple failures happen concurrently.

## Example browser-visible error response

Example `fetch` response payload/body:

```json
{
  "detail": "Internal Server Error",
  "request_id": "2bb17a20af5442b69dad7e78e351c1f7"
}
```

Example response headers:

```http
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Access-Control-Allow-Origin: https://horizon.globalnaturewatch.org
Access-Control-Allow-Credentials: true
Vary: Origin
```

## How this accelerates investigation

1. Frontend logs `request_id` from the response body and includes it in Sentry/user-facing error details.
2. Backend engineer searches logs for that exact `request_id`.
3. They immediately get the full request timeline (start log, stack trace, response log) without guessing by timestamp, user, or endpoint.

## Sequence diagram (error path with CORS headers)

```mermaid
sequenceDiagram
    autonumber
    participant B as Browser
    participant API as FastAPI app
    participant LM as logging_middleware
    participant H as Route handler

    B->>API: GET /api/insights/{id}<br/>Origin: https://horizon.globalnaturewatch.org
    API->>LM: Enter middleware
    LM->>LM: Generate request_id + bind log context
    LM->>H: call_next(request)
    H-->>LM: Raise unhandled exception
    LM->>LM: Log exception with request_id
    LM-->>B: 500 JSON<br/>{detail, request_id}<br/>Access-Control-Allow-Origin: https://horizon.globalnaturewatch.org<br/>Access-Control-Allow-Credentials: true<br/>Vary: Origin
    B->>B: JS can read response body and request_id
```

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -5,6 +5,7 @@ from typing import Optional
 import structlog
 from fastapi import FastAPI, Request, Response, status
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse
 
 from src.agent.graph import close_checkpointer_pool, get_checkpointer_pool
 from src.agent.utils.sgrep import data_status
@@ -31,6 +32,20 @@ from src.shared.logging_config import get_logger
 from src.shared.version import get_version
 
 logger = get_logger(__name__)
+
+
+def _add_cors_headers_for_error(request: Request, response: Response) -> None:
+    origin = request.headers.get("origin")
+    if not origin:
+        return
+    response.headers["Access-Control-Allow-Origin"] = origin
+    response.headers["Access-Control-Allow-Credentials"] = "true"
+    vary = response.headers.get("Vary")
+    if vary:
+        if "Origin" not in vary:
+            response.headers["Vary"] = f"{vary}, Origin"
+    else:
+        response.headers["Vary"] = "Origin"
 
 
 @asynccontextmanager
@@ -94,6 +109,7 @@ async def mosaic_cache_headers(request: Request, call_next) -> Response:
 async def logging_middleware(request: Request, call_next) -> Response:
     """Middleware to log requests and bind request ID to context."""
     req_id = uuid.uuid4().hex
+    request.state.request_id = req_id
 
     structlog.contextvars.clear_contextvars()
     structlog.contextvars.bind_contextvars(request_id=req_id)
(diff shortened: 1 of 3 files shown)


src/api/app.py
```diff
@@ -119,7 +135,14 @@ async def logging_middleware(request: Request, call_next) -> Response:
...
         response_code = 500
-        raise e
+        response = JSONResponse(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            content={
+                "detail": "Internal Server Error",
+                "request_id": req_id,
+            },
+        )
+        _add_cors_headers_for_error(request, response)
```
01painadam, line 149: Small note that this now consumes the exception so nothing downstream sees it. 

I suppose that's alright given logger.exception above, but just flagging that we're now not by catching raised exceptions at an outer layer. Not blocking, though.

01painadam (approved):
g2g once conflicts resolved

src/api/app.py
gtempus: Nice! Thanks for flagging that. 💪 
I added a TODO for when we get an APM integration started

PR #771, opened 2026-07-24
Add backend-owned color registry for insight charts and map legends (phase 1)

yellowcap:
## Summary

Chart and map-legend colors (land cover classes, tree-cover-loss drivers,
dataset series colors) currently live only in `project-zeno-next`, keyed by
literal English strings. This breaks once the agent translates category
labels for non-English users, and has already produced at least one drift
bug between the frontend's two color files (a GHG dataset name mismatch).

This PR is **phase 1** of a plan to make the backend the single source of
truth for these colors. Full design and rationale in
[`docs/insight-chart-colors-plan.md`](docs/insight-chart-colors-plan.md).

## Overall plan

- **Phase 1 (this PR)**: canonical color registry lives in the dataset
  catalog YAMLs (`categories`/`series_color`/`divergent_colors`, keyed by a
  stable English slug — verified no dataset's analytics API actually
  returns a numeric class code, only string labels). A new
  `GET /api/datasets/catalog` endpoint exposes it. The companion
  `project-zeno-next` PR wires the map legend to read from this endpoint
  instead of its own hardcoded colors. Ships independently, zero risk to
  chart generation, fixes an existing color-drift bug as a side effect.
- **Phase 2 (not started)**: thread a stable category slug through the
  code-executor LLM prompt and `InsightChart` schema (today the LLM writes
  already-translated category labels straight into `chart_data`, with no
  stable key surviving translation), add a deterministic (non-LLM) color
  resolution step, migrate `InsightChartOrm`, and cut the chart-rendering
  frontend over from its local color file to backend-supplied colors.
- **Phase 3**: delete the now-redundant frontend color files once phase 2
  is verified in production.

## Changes

- 9 catalog YAMLs gain `categories`/`series_color`/`divergent_colors`
  (verbatim ported hex values, no new color design) and, for SBTN Natural
  Lands, a `legend_categories: false` flag — that dataset's map legend
  intentionally collapses non-natural classes into one row, which this flag
  preserves without hardcoding a dataset-id special case anywhere.
- `src/agent/datasets/palette.py`: loader/registry with hex validation.
- `GET /api/datasets/catalog` (`src/api/routers/metadata.py`) + schemas.

## Test plan

- [x] `uv run pytest tests/tools/test_dataset_palette.py tests/api/test_metadata.py tests/unit` — 439 passed
- [x] Verified the 9 YAMLs still parse via `src.agent.datasets.config.DATASETS`
- [x] Confirmed pre-existing live-LLM test flakes in `test_pick_dataset.py` reproduce identically without this change (stash-tested)
- [x] ruff / ruff-format / mypy clean on touched files

### DIFF
--- a/docs/insight-chart-colors-plan.md
+++ b/docs/insight-chart-colors-plan.md
@@ -1,0 +1,258 @@
+# Insight chart colors — implementation plan
+
+*Status: agreed design, phase 1 in progress. Written 2026-07-24. This document
+is self-contained: a fresh session can pick up any phase from here without
+prior conversation context.*
+
+## Problem
+
+Chart and category colors for insights (land cover classes, tree-cover-loss
+drivers, dataset series colors, divergent GHG colors) are currently defined
+**only in the frontend** (`project-zeno-next`), keyed by literal English
+strings, and applied client-side when rendering `ChartWidget`. Two independent
+frontend files hold overlapping-but-unsynced color data:
+
+- `app/config/chartColorMappings.ts` — `CHART_COLOR_MAPPING` (category → hex,
+  per chart field), `DATASET_SERIES_COLORS` (dataset name → hex),
+  `DATASET_DIVERGENT_COLORS` (dataset name → {positive, negative})
+- `app/constants/datasets.ts` — map-legend `legend.items` (label → hex), same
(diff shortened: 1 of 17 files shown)


yellowcap:
Moving this in because its additive. Will open PR after that is more critical.

PR #772, opened 2026-07-24
add inspect command to inspect source geom tables

batpad:
Still debugging failure to run data migration with OOM errors on the db - doing the transactions in chunks, however small, did not make it work.

This adds an `inspect` flag to the `build-aois` command to check geometries with a huge amount of vertices. Also, does some further tuning of the queries. Based on the results of `inspect`, we can add a simplification step to the geometries, though I'd rather do simplification in a separate step / PR.

cc @yellowcap

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.24.3"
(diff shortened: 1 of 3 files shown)


yellowcap (approved):
Nice catch! 🎣

PR #775, opened 2026-07-27
handle OOM errors caused by multipolygons with many small polygons

batpad:
The main culprit was the "Old Growth Management Areas" (British Columbia)" feature in the `wdpa` table with id `555651780`. It had a total of 4.85 million vertices across over 19,000 separate polygons. This was causing the ST_MAKEVALID operation to follow across the entire MultiPolygon. This makes a small change to run makevalid on individual parts instead of the whole multipolygon.

I did a bunch of testing things running on a `psql` shell on staging, running equivalent commands and confirming none caused an OOM. I'm hoping with this, the migration finally works.

@yellowcap I'll try and clear my brain and write out a general issue around dealing with these large geometries. Just want to tackle that separately from this AOI unification step.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.27"
+version = "2026.7.27.1"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


yellowcap (approved):
🤞🏽 for the migration!

PR #776, opened 2026-07-27
Move show-imagery skill from experimental to default profile

yellowcap:
Move the `show-imagery` skill from the `experimental` profile to the `default` profile so satellite imagery is available alongside the core skills (analyze, pull-data, etc.).

**Changes:**
- `src/agent/agent_config.py`: moved `"show-imagery"` from `EXPERIMENTAL_SKILLS` to `DEFAULT_SKILLS`
- `tests/unit/agent/test_profile_manifest.py`: updated default manifest snapshot
- `tests/unit/agent/test_feature_flag.py`: updated feature flag assertions

The experimental profile still inherits show-imagery via the extends chain.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.7.27.1"
+version = "2026.7.27.2"
 description = "Language Interface for Maps & WRI/LCL data APIs."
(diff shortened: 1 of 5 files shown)


yellowcap:
Moving this in as per discussion with @01painadam

PR #778, opened 2026-07-28
inject chart data into inspect_view_context with size-gated formatting

yellowcap:
## Summary

When the agent calls `inspect_view_context` to examine insights on screen, it now sees the actual chart data instead of just a row count.

**Size-gated behavior:**
- Series with <= 20 rows → full data rendered as a compact text table
- Series with > 20 rows → per-column statistics (min/max/mean for numeric, distinct count + samples for string)
- Meta-columns (`aoi_id`, `aoi_type`) are skipped to avoid noise

## Context

Previously `format_insights` deliberately omitted chart data, showing only the row count. This meant the agent could see chart titles and axis configuration but had no visibility into the actual values — unlike the text generator and analyst tool which both receive the full data.

The analyst tool already had its own chart-data injection (raw CSV, gated at 4000 chars). Rather than keep two divergent implementations, the analyst tool now calls the same `format_chart_data` helper, so both paths share one size-gated formatting behavior instead of drifting apart.

## Changes

- `src/agent/tools/inspect_view_context.py`: added `format_numeric_stats` and `format_chart_data` helpers, called from `format_insights` for each chart. Named without a leading underscore since both are called from `src/agent/subagents/analyst/tool.py` as well — a leading underscore would misleadingly signal "module-private" for something used across module boundaries.
- `src/agent/subagents/analyst/tool.py`: replaced the old ad hoc CSV formatting (pandas `to_csv`, 4000-char gate) with a call to the shared `format_chart_data`.
- `src/agent/i18n.py`: removed the `analyst.chart_data_csv_header` translation entry, which was only used by the CSV formatting this PR replaces.
- `tests/agent/test_inspect_view_context.py`: new tests covering the formatting helpers, both paths (small/large), and edge cases (empty data, meta-column-only data, char-limit fallback, numeric edge cases).

## Trade-offs

- Adds potentially significant token count when multiple insights with many small series are on screen. The 20-row threshold keeps it bounded; larger series get stats instead of raw rows.
- The shared formatter's labels ("Data (N rows...)", "per-column stats", etc.) are hardcoded English, unlike the rest of the analyst tool's user-facing strings which go through `t()`. Acceptable for now since this is tool-message content the agent reasons over, not primary user-facing copy, but worth revisiting if that changes.


src/agent/tools/inspect_view_context.py
```diff
@@ -52,6 +52,14 @@
     "dashboard_name",
 }
 
+# Rows at or below this get full data injected; above get per-column stats.
+DATA_INJECT_THRESHOLD = 20
```
01painadam: I'd make this bigger (30?) TCL is 2001-2025 already

01painadam: We should consider not adding Curated Insights with more than DATA_INJECT_THRESHOLD datapoints to avoid hallucination

...NRT alerts dataset for example. Maybe those should cap out at "the last week / 2 weeks / month" (7, 14, 30 datapoints respectively)

01painadam (approved):

src/agent/tools/inspect_view_context.py
yellowcap: yes charts with too much data are not useful anyway. So keeping this bounded is healty.

The limit here is merely about injection size though, so that the tool message and the inspection results are reasonably bounded in case there is a lot of data in the chart.

PR #784, opened 2026-07-29
Add LGMS, allow datasets to be excluded in agent profile

jterry64:
### Add the LGMS dataset
Adds **Land GHG Monitoring System (LGMS)** (`dataset_id: 12`) — annual land greenhouse-gas emissions, removals, and net flux (2016–2024) for GADM admin areas, via `/v0/land_change/land_ghg_inventory/analytics`.

The endpoint returns a **per-section** result (`vegetation`, `mineral_soil`, `organic_soil`, `agriculture`). Currently, the generate insights tool can only handle one table at a time, so simplify things, I added a function in the analytics handler to flatten it to one table by adding a category and class column. Depending on how much we want to use the agent here and what the widget looks like, we can either push that code into the analytics endpoint itself, or figure out how to deal with multiple tables later; this is just to quickly get out a prototype. 

### Exclude datasets per agent profile

This dataset needs to be hidden from the public until NY Climate Week. To support that in feature flags, I added a new 1excluded_datasets1 property to a profile. This hides the datasets from **both** dataset selection (`pick_dataset`) and the capabilities listing, with a pull-time guard as backstop. Here's an example:

```python
# default profile hides LGMS; experimental reveals it
AgentConfig(DEFAULT_PROFILE, extends=BASE_PROFILE, skills=DEFAULT_SKILLS,
            excluded_datasets=frozenset({"Land GHG Monitoring System (LGMS)"}))

AgentConfig(EXPERIMENTAL_PROFILE, extends=DEFAULT_PROFILE, ...,
            excluded_datasets=frozenset())   # not inherited → LGMS visible
```

This will also be useful just when adding new datasets, since sometimes their release can get delayed by long QC process.


solomon-negusse (changes requested):
Looks good to me @jterry64, but a couple of data things to possibly update here:
- Livestock emissions are available in the API now, so can include it here
- Agriculture is being served from the API annualized to vegetation period (once my livestock PR is merged) consistent with how we were asked to present static soil data. That should let you simplify the data processing here.

solomon-negusse (approved):

PR #789, opened 2026-08-12
PZB-1195: Remove DIST context layers and clean up YAML/evals

guilherme-n:
## Pull request type
- [x] Feature

## What is the current behavior?

DIST-ALERT (`dataset_id: 0`) is the only dataset carrying the `driver` (LDACS), `natural_lands`, `grasslands` and `land_cover` context layers. Any query asking for a disturbance breakdown gets routed to DIST just to reach those intersections, and the DIST YAML reinforces it by defaulting to driver-classified alerts and class-based charts. Evals and unit tests encode the same assumption.

Backend half of PZB-1150; the FE side is already merged.

## What is the new behavior?

The four context layers are removed from DIST (`context_layers: null`), and its prompt, code and presentation instructions no longer produce class or driver breakdowns. Integrated Alerts becomes the default for near-real-time disturbance queries, and DIST is reserved for when the user asks for it by name.

Evals and tests follow: no row expects DIST any more, the four DIST/context-layer combinations are dropped from `test_pull_data`, and gold rows whose questions genuinely needed a removed intersection are re-golded as `no_selection` rather than repointed to a dataset that cannot answer them.

### Before
before

_______

### After
after


src/agent/datasets/catalog/global_all_ecosystem_disturbance_alerts_dist_alert.yml
```diff
@@ -28,60 +18,37 @@ keywords:
...
-  when it occurs near rivers. Severe boreal fires may be misclassified as conversion. Always interpret alerts as indicators,
-  not definitive outcomes.
-
-  '
+prompt_instructions: >-
+  Use this dataset to show recent vegetation changes across all ecosystems. Data is measured in hectares (ha); use this
+  as the unit of analysis in all cases. There are NO context-layer intersections (no driver/LDACS, natural lands,
+  grasslands, or land cover breakdowns) — report disturbed area by date and by confidence only. Clarify whether the user
+  cares about temporary disturbance or permanent conversion; recommend annual land cover or tree cover loss datasets for
+  historical or long-term changes. For recent trends, show alert areas over time using line charts. Prefer Integrated
```
danscales: Maybe move this important sentence:

Prefer Integrated Alerts for near-real-time disturbance queries unless the user specifically asks for DIST-ALERT.

to be right after the first sentence above? It may not matter to the AI, but it feels better to put the important note (only use if specifically asks) right near the beginning. It also makes it clearer to the human reader.

danscales (approved):
Looks great, except my one comment (which you can decide if you want to do).

src/agent/datasets/catalog/global_all_ecosystem_disturbance_alerts_dist_alert.yml
guilherme-n: Both applied. Moved the "Prefer Integrated Alerts" sentence up to right after the opening sentence, and rewrote the LDACS line in description so it no longer advertises driver classification as available here.

The description one turned out to matter more than it looks: that field feeds the dataset embeddings used for retrieval, and the selector was quoting it back when declining, claiming DIST "includes a filtered subset (LDACS) that can classify drivers". After the change that language is gone and the selector points at Tree Cover Loss by Dominant Driver instead.

PR #790, opened 2026-08-12
fix: guard pull_data against missing AOI/dataset selections

yellowcap:
## The error

Production, `2026-08-12T09:30:47Z`. Langfuse trace: [`d4c0aa713e096b7958d7cdcd9cdfeea9`](https://langfuse.globalnaturewatch.org/project/wri_lcl/traces/d4c0aa713e096b7958d7cdcd9cdfeea9?observation=c07b1325f270c8cb). Langfuse session: [81dbede5-ec64-4099-9e87-85339af777b3](https://langfuse.globalnaturewatch.org/project/wri_lcl/sessions/81dbede5-ec64-4099-9e87-85339af777b3)

```
{"tool_name": "pull_data", "event": "tool_execution_failed", "logger": "src.agent.graph",
 "level": "error", "query": "Paris, Île-de-France, France - (district-county) [FRA]", ...}

  File "/app/src/agent/tools/pull_data.py", line 123, in pull_data
    aoi_names = [a["name"] for a in state["aoi_selection"]["aois"]]
                                    ~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'aoi_selection'
```

Note what did *not* fail: line 100, `dataset = state["dataset"]`. The thread had a dataset but no AOI.

## Cause

`AgentState` is a plain `TypedDict` — keys exist only once a node writes them. `aoi_selection` was absent because no `pick_aoi` call ever completed a selection.

The `query` field is the giveaway. `"Paris, Île-de-France, France - (district-county) [FRA]"` is generated by `_format_aoi_candidate` — it's a nudge option, verbatim. The sequence:

1. `pick_aoi` ran on "Paris"; `check_duplicate_aois` found the same name in several countries.
2. It returned the `aoi_choice` nudge, whose `Command` writes **only** `nudge` and `messages` — not `aoi_selection`.
3. The user clicked an option. Clicking resubmits the option string as the next human message; there is no dedicated resolver tool.
4. The model saw a fully-qualified place name in the user's turn and called `pull_data` directly instead of re-running `pick_aoi`.
5. `KeyError`.

A `*_choice` nudge records the options offered, never a selection. `pick_dataset`'s `dataset_choice` branch has the identical write-free shape, so `state["dataset"]` was exposed to the same crash.

`pull_data` was the only unguarded consumer — `middleware.py`, `create_dashboard.py`, `show_imagery.py` and `pick_dataset` all use `.get()` with a fallback.

## Changes

**Guard both selections in `pull_data`.** Missing `dataset` or empty `aoi_selection.aois` now returns a `human_feedback` message naming the tool to run, instead of raising. When a matching `*_choice` nudge is pending, the message appends the offered options so the model re-resolves the user's actual click rather than guessing a different area.

This matters because of how the crash surfaced: `handle_tool_errors` returned `Tool 'pull_data' failed unexpectedly (KeyError)` to the model, with no hint that the fix was to call `pick_aoi`. That is the input condition for the guess-and-retry thrashing tracked separately (repeated `pick_aoi`/`pick_dataset`/`pull_data` calls after an uninformative failure).

**New i18n keys** `pull_data.no_aoi` / `pull_data.no_dataset`, all 15 languages, matching `show_imagery.no_aoi` — so no runtime LLM translation call on a hot path.

**Told the model a nudge reply is unresolved**, in the two places it reads: `NUDGE_ALREADY_SET_NOTE` (shared by `pick_aoi` and `pick_dataset`) and the system prompt's Nudges section, which now points at the session block as the check — if it still says `AOI: none`, resolve first regardless of what the user's message looks like.

**Documented that the option format is load-bearing** (`_format_aoi_candidate`). The re-resolution round-trip terminates only because the option carries the full comma-joined hierarchy: re-querying with it scores 1.0 by trigram similarity against the same `name` column, while same-name rows in other countries fall below `pg_trgm.similarity_threshold = 0.2`, so `check_duplicate_aois` finds no rivals and the nudge does not re-fire.

```
1.000  Paris, Île-de-France, France
0.182  Paris, Ontario, Canada          filtered
0.154  Paris, Texas, United States     filtered
0.143  Paris, Kentucky, United States  filtered
```

Shortened to a bare `"Paris"`, every Paris would match again and the nudge would re-offer the same choice indefinitely — with the new guard, on a loop.

**Refactor:** the repeated `Command(update={"messages": [ToolMessage(...)]})` block in `pull_data` is extracted to `_human_feedback`. It was written out three times; the guards would have made it five.

## No frontend change needed

`ChatNudge.handlePick` optimistically adds the picked dataset's layer to the map for `dataset_choice`, which round-trips into `ui_context.dataset_selected` and lets the backend set `state["dataset"]` without `pick_dataset`. There is no `aoi_choice` equivalent — which is why `dataset` was present and `aoi_selection` was not.

That asymmetry is a latency optimization, not the mechanism. The intended flow works on its own: click → text → `pick_aoi` → `aoi_selection` in state → streamed → `pickAoi.ts` adds the area layer and flies to it. The only defect was the model skipping `pick_aoi`, which is what this PR fixes.

## Tests

`TestMissingSelectionGuards` in `tests/tools/test_pull_data.py` — five cases: missing `aoi_selection` (the production case), missing `dataset`, present-but-empty `aois`, pending `aoi_choice` options echoed to the model, and an unrelated pending nudge *not* echoed. All five fail against `main` with `KeyError` and pass here.

`tests/unit` (418) and `tests/api` (294) green; ruff and mypy clean.


src/agent/tools/pull_data.py
```diff
@@ -97,27 +140,40 @@ async def pull_data(
...
     language = state.get("language") or DEFAULT_LANGUAGE
-    dataset = state["dataset"]
+
+    # Selection guards. pick_aoi/pick_dataset can end a turn with a *_choice
+    # nudge instead of a selection — they write `nudge`, not `aoi_selection`/
+    # `dataset` — and the user's click comes back as plain message text, so the
+    # model can reach this tool believing a choice was made while state still
+    # holds nothing. Answer with feedback rather than KeyError: an unhandled
+    # crash reaches the model as an opaque "failed unexpectedly" and it retries
+    # blind (see TICKET_AOI_RETRY_LOOP.md).
```
solomon-negusse: Is this file meant to be added?

solomon-negusse (approved):
This looks good to me @yellowcap with the caveat that I haven't been in this codebase for sometime. Just one minor comment.

src/agent/tools/pull_data.py
yellowcap: good catch @solomon-negusse . That was out of date, removed that wording

PR #791, opened 2026-08-13
Promote dashboard and explore skills to the default profile

yellowcap:
## Summary
- Moves the last two experimental-only skills, `dashboard` and `explore`, into `DEFAULT_SKILLS` — they now ship in every default agent instead of requiring `ff=experimental`.
- `EXPERIMENTAL_SKILLS` is now empty; the experimental profile is now just the default plus two standalone tools not yet owned by any skill (`inspect_view_context`, `update_insight_display`).
- **Fix:** `dashboard.md`'s body already instructed the agent to call `search_insights` to recall a past finding, but that tool wasn't in the skill's `requires:` — it only rode along because it happened to be bundled separately in `EXPERIMENTAL_TOOLS` alongside `dashboard`. That coincidence broke once `dashboard` graduated to `default` without `search_insights` following it, so `search_insights` is now added to `dashboard`'s `requires:` (structurally tying it to wherever the skill is declared) and dropped from `EXPERIMENTAL_TOOLS`.
- Updates the profile manifest snapshot tests and feature-flag/view-page/routing tests that hardcoded the old split.
- Fixes the `src/agent/docs/feature-flags.md` profile-chain diagram, which was already stale (missing `show-imagery` from `default`) even before this change.

## Test plan
- [x] `tests/unit/agent` (229 tests) passes, including updated `test_profile_manifest.py`, `test_feature_flag.py`, `test_view_pages.py`, `test_skills.py`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.8.12"
+version = "2026.8.13"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/agent_config.py
+++ b/src/agent/agent_config.py
@@ -124,7 +124,6 @@ EXPERIMENTAL_SKILLS = ()
 EXPERIMENTAL_TOOLS = (
     inspect_view_context_spec,
     update_insight_display_spec,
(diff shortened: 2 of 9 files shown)


yellowcap:
Merging, got 🟢 from @01painadam

PR #794, opened 2026-08-21
Use new AOI table: API and tools migration

batpad:
In #746, we created a new Unified AOI table that consolidates AOI data from different tables into a single table to make it much easier and performant to add features like better search, filter and also concepts like Saved Areas. #746 has been deployed onto staging and production - this created the new table with nothing using it yet.

This PR migrates all the AOI reads over to the new "Unified AOI table" - currently, the goal is to maintain feature parity with current behaviour, just switch to reading from the new table. Once this lands and we test, it should be trivial to add new filters we need, ability to upload shapefiles for custom areas, etc.

This did land up being quite a big chunk, and will require a bunch of thorough testing. I was working through HTML docs locally as I worked through this to visually get a sense of the table structures, etc. - I put this at https://sanjay.bhangar.com/tmp/aoi-architecture/ for reference - I think all the relevant docstrings, etc. have been captured in our code, so not sure there's value of moving these docs into the code-base, but happy to make a trimmed down `.md` version to check in to the code.

My ideal plan here would be to merge this to staging on Monday and test thoroughly, and hopefully be able to push to prod by end of next week.

cc @yellowcap @01painadam @kamicut


yellowcap (approved):
Tested this locally with full AOI DB setup, worked fine when using the frontend, both on chat and on the AOI search on dasboard page. 

Gold evals run gave perfect score for AOI selection, so agent based search is not affected or improved.

Great work @batpad !! 🙌🏽

PR #800, opened 2026-08-25
Planet imagery MVP

jterry64:
## Summary

Adds Planet’s monthly integrated-alert mosaic as an imagery option alongside Sentinel-2.

image

Planet imagery is high-resolution but limited to buffered areas around integrated alerts within its configured Amazon coverage. Currently it's using a test endpoint from planet that only covers a limited 10x10 degree tile around 0 N, 70 W (near Colombia). 

## Behavior

- Prefer Planet for Amazon alert workflows using completed monthly data.
- Use Sentinel-2 for recent imagery and areas outside Planet coverage.
- Default undated Sentinel-2 requests to the previous two weeks and suggest the prior month’s Planet mosaic when available.
- Return the selected provider, date range, bounds, and zoom limits through the existing `ImageryState`.

The imagery implementations are separated into Planet and Sentinel-2 provider classes behind the existing `show_imagery` tool.

## Testing

Besides automated tests, tested manually against the following questions we can add as evals:

- “Show me integrated deforestation alerts and imagery for Mirití-Paraná, Amazonas, Colombia.”
- “Show me alerts from the past two weeks and the latest imagery for Tabatinga, Amazonas, Brazil.”
- “Show me alerts and imagery from June 2026 for Tabatinga, Amazonas, Brazil.”
- “Show me recent alerts and imagery for Zurich, Switzerland.”
- “Show me recent alerts and imagery for Sheffield, England.”

## Important Notes
- This relies on the proxy endpoint from this PR, so don't merge until it's merged in: https://github.com/wri/gfw-tile-cache/pull/282 (I tested using a local tile cache server)
- I have a PR on the FE I'll up with shortly so it properly loads the Planet layer. This won't break anything, the Planet layer just won't load.
- I didn't add a specific Planet feature flag yet. That's going to require a bit of thinking and extra work, and just wanted to get an MVP out to staging. I think it should be safe to put this out and either add the feature flag as a follow up or revert this commit as needed.


src/agent/skills/skills_md/show-imagery.md
```diff
@@ -10,11 +10,23 @@ requires: pick_aoi, show_imagery
 1. `pick_aoi` — only if the AOI is not already in state. Imagery works for
    regional areas (up to ~50,000 km²); country-scale requests will be
    rejected — ask the user for a smaller region.
-2. `show_imagery(target_date)` — pass the user's date (YYYY-MM-DD) or omit
-   for the most recent imagery. Defaults: scenes within ±7 days of the
-   target date, under 20% cloud cover.
-3. **Stop.** Confirm what is shown (scene count and date span from the tool
-   message). No dataset, pull or insights unless asked.
+2. `show_imagery(target_date)` — pass the user's date (YYYY-MM-DD). When
```
yellowcap: originally, target date was the only arg probably. can be dropped

```suggestion
2. `show_imagery` — pass the user's date (YYYY-MM-DD). When
```

src/agent/skills/skills_md/show-imagery.md
```diff
@@ -10,11 +10,23 @@ requires: pick_aoi, show_imagery
...
+   showing imagery alongside alert data, set `target_date` to the alert
```
yellowcap: We need a stronger separation of planet vs s2 in this skill. 

There is some wording around alert data, and aligned alerts. But that is a bit out of context here. I think in this skill I would add an intro section about the differences and especially the relation of planet with alerts. Since planet is buffered around alerts, we should be very explicit about that in the skill.

I would suggest a deeper rewrite of the skill

src/agent/imagery/planet.py
```diff
@@ -0,0 +1,89 @@
...
+            ),
+            aoi_names=[aoi["name"] for aoi in request.aois],
+        )
+        message = (
+            "Showing the limited-coverage Planet monthly mosaic for "
+            f"{month_start.strftime('%B')} {month_start.day}–{month_end.day}, "
+            f"{month_start.year}. Sentinel-2 imagery is also available if "
+            "you'd like to compare it."
+        )
+        return ImageryProviderResult(
```
yellowcap, line 87: Nice class based approach! I like it

src/agent/tools/show_imagery.py
```diff
@@ -81,124 +104,61 @@ async def show_imagery(
...
+        provider != "sentinel-2"
+        and not prefer_sentinel
+        and PLANET_PROVIDER.covers(aois)
+    ):
+        return _provider_command(
+            await PLANET_PROVIDER.get_imagery(request), tool_call_id
         )
+
+    if provider == "planet":
         return _feedback(
```
yellowcap: shall we make this part of the tool msg, but keep going? Seems like showing some imagery at this point is a request, so why not fallback to sentinel always once the tool has been invoked? we can alert the user that planet is not available, hence we show s2

src/agent/tools/show_imagery.py
```diff
@@ -81,124 +104,61 @@ async def show_imagery(
...
 
 
 SPEC = ToolSpec(
     tool=show_imagery,
     category=ToolCategory.PRIMITIVE,
     prompt_fragment=(
-        "- show_imagery(target_date): show a Sentinel-2 satellite imagery "
-        "layer on the map for the AOI in state. Run pick_aoi first; regional "
-        "areas only."
+        "- show_imagery(provider, target_date): show Planet or Sentinel-2 "
```
yellowcap: same here, the args are incomplete anyway and not necessary to add in the prompt.

```suggestion
        "- show_imagery: show Planet or Sentinel-2 "
```

yellowcap (commented):
thanks @jterry64 looks good! 🛰️  a few small tweaks suggested, mainly around skills and the provider choice logic

jterry64:
@yellowcap thank you for the useful feedback! I removed the params like you mentioned, and I added a context section in the skill. It seems to work more consistently now!

One thing I observed though while testing is that the agent is mostly instructed to call just one skill. So if I ask for alerts AND imagery, it will usually call the analyze skill and not the show imagery skill (it'll just call the tool directly). Is this by design and I shouldn't mess with it? This also might just be an artifact of how I'm testing it, real users would probably ask for them in separate queries.

01painadam:
@jterry64 let's add some evals to GOLD. Do you have any examples you've been testing against so we dont duplicate work?

jterry64:
@01painadam I was testing those questions in the PR description, but just directly. IIRC there was a judge added for imagery already, so I think it'd just need to be adjusted to be able to check for a specific system. Do note though that this is still the test endpoint from Planet, so it's only for a tiny area (Tabatinga is an example area that does have Planet imagery right now).

jterry64:
@gtempus since the show-imagery tool/skill is already released, I just made this PR split them out temporaily into a different tool and skill that's only in the experimental profile. This way we can keep the Planet totally separate until it's ready for release. Once we want to release, we can swap or merge the tool/skills and delete the old parts. Is this what you were thinking?

01painadam:
> @01painadam I was testing those questions in the PR description, but just directly. IIRC there was a judge added for imagery already, so I think it'd just need to be adjusted to be able to check for a specific system

Nice, ok - I'll add a PR on gnw-gold-evals to cover these cases and we can merge when we're ready

yellowcap:
> One thing I observed though while testing is that the agent is mostly instructed to call just one skill. 

@jterry64 that is a good observation! I guess in the beginning we had few skills, so that was not a problem. I would start by changing the prompt wording. 

Replace the "exactly one" / "usually one skill" language with something like:                                                                                                                                                                
                                                                                                                                                                                                                                              

> "Match the request to the skill(s) that apply. Some requests span multiple skills (e.g. 'build a dashboard about deforestation' needs both dashboard and analyze). Load all that are relevant."

jterry64:
@yellowcap ok, that's good to know. It's working fine as is (I changed the "use_when" that makes more inclined to call two skills). I think in the final version I'll change to allow multiple skills more generally, but I might wait on that just because that's a more fundamental change I'd want to do deeper evals on, whereas this is a feature flagged prototype.

jterry64:
@yellowcap I moved the redundant code out to the base class and simplified the classes to more true singletons, and also added language support with initial AI translations.

yellowcap (approved):
Great thanks I did not test this locally, but I think it can move forward so we can test it online.

PR #801, opened 2026-08-26
PZB-1270: Fix AOI subregion: resolve GADM admin depth per country

yellowcap:
**Problem:** The same admin-unit word names different GADM depths in different countries. "Province" is ADM1 in Canada but ADM2 in Spain. "Municipality" is ADM1 in China (Beijing, Shanghai) but ADM3 in most other countries.

The geocoder uses a single global `subregion` guess (`state`, `district`, etc.), which is wrong for these cases. "Which provinces in Spain lost the most forest?" returned Spain's 17 Autonomous Communities (ADM1) instead of its 52 provinces (ADM2).

**Solution:** The geocoder extracts the user's actual admin-unit word alongside its subregion guess. At query time, `_resolve_effective_subregion` looks up the correct GADM depth per country and overrides the LLM's guess when they differ. Multi-country requests resolve each place independently.

**Changes:**

- `scripts/generate_gadm_admin_types.py` + `src/shared/fixtures/gadm_admin_types.json` — offline generator reads GADM's `ENGTYPE_N` columns, produces a per-country depth lookup with ~60 curated canonical terms. Generated once, checked in.
- `src/shared/gadm_admin_types.py` — `resolve_gadm_admin_level(term, iso3)` with exact match (shallowest-level preference on ties) and `difflib` fuzzy fallback scoped to the country's terms.
- `PlaceQuery.admin_term` — new optional enum field extracted by the geocoder alongside `subregion`.
- `src/agent/subagents/pick_aoi/tool.py` — `_resolve_effective_subregion` corrects the depth per selected AOI before querying.
- `src/agent/subagents/pick_aoi/prompts.py` — teaches the geocoder to extract `admin_term`.
- `src/agent/subagents/pick_aoi/selection_name_util.py` — displays the user's term ("52 Provinces in Spain") instead of the generic resolved label.

**Verified live:**
- "provinces of Spain" → 52 provinces (was 17 Autonomous Communities)
- "municipalities of China" → correct ADM1 depth (Beijing, Shanghai, ...)

**Note:** Originally this PR also fixed a GADM `NAME_1='NA'` data defect and a geocoder acronym/UK-subregion gap. Both landed on `main` via a different implementation (#797, #798, #799), so this was rebased down to just the admin depth resolution.


src/shared/gadm_admin_types.py
```diff
@@ -0,0 +1,82 @@
...
+    },
+)
+
+# {iso3: {level (as str "1".."5"): [ENGTYPE values GADM uses at that level]}}
+# Broader than GADM_ADMIN_TERMS on purpose: it also carries the long tail of
+# country-specific/misspelled ENGTYPE values, used only as fuzzy-match targets
+# below, never shown to the LLM.
+_BY_COUNTRY: dict = _FIXTURE["by_country"]
+
+_FUZZY_CUTOFF = 0.6
```
01painadam: Since this matches on character similarity a cutoff=0.6 it resolves actual English words to unrelated administrative units. 

Ran a check and found that `resolve_gadm_admin_level("City", "USA")` returns 2 by fuzzy-matching "City" → "County"; "Community" → "Commune" (AGO), "County" → "Commune" (AGO), "Region" → "Sub-Region" (ALA). 

Across all term/country pairs where the exact match fails, ~8% (1,162 of 14,146) fuzzy-resolve, and a large share of those are semantic mismatches rather than near-misses.

This matters because it is strictly worse than doing nothing: a wrong fuzzy hit overrides the LLM's global default, whereas returning None would fall back to it.

We can either raise the cutoff (e.g. 0.85, which keeps genuine typos like the "Sate" → "State" case the tests pin, while rejecting City/County at ~0.55–0.6), or drop the fuzzy pass entirely now that the enum prevents typos and let unresolvable terms fall through to the LLM's guess. 

Whichever you choose, a test asserting resolve_gadm_admin_level("City", "USA") is None confirm.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -863,7 +924,7 @@ async def lookup(
         logger.debug(f"Pick AOI tool message: {tool_message}")
 
         selection_name = build_selection_name(
-            match_names, subregion, len(final_aois)
+            match_names, subregion, len(final_aois), display_term=admin_term
```
01painadam: Could be minor... admin_term is passed as display_term regardless of whether 
 _resolve_effective_subregion accepted it.

When the term doesn't resolve for that country the query correctly falls back to the LLM's depth, but the label still uses the user's word — so "provinces in Kenya" returns Kenya's ADM1 "counties" under the name "47 Provinces in Kenya". Silently mislabelling the units is arguably worse than showing the generic "47 States", because it reads as confirmation that we understood the question.

Suggest passing display_term only when the resolved value differs from subregion, or having _resolve_effective_subregion return both the depth and the term to display

src/agent/subagents/pick_aoi/selection_name_util.py
```diff
@@ -1,31 +1,45 @@
...
+    """Pluralize *word*, handling the consonant+"y" -> "ies" case (country ->
+    countries, municipality -> municipalities). Everything else gets a plain
+    "s", which is correct for every subregion/admin term this module sees."""
+    if (
+        len(word) > 1
+        and word[-1].lower() == "y"
+        and word[-2].lower() not in "aeiou"
+    ):
+        return word[:-1] + "ies"
+    return word + "s"
```
01painadam, line 23: This +s rule will regresses on some plurals. 

Checked against all 62 terms in `canonical_terms` e.g. Parish → "Parishs", Islands → "Islandss", Metropolis → "Metropoliss".

A bit of a potential rabbit hole but an extra condition covers all three: if word.endswith(("s", "x", "z", "ch", "sh")): return word + "es", guarded so Islands (already plural) is returned unchanged.

...or keep a small explicit exceptions dict alongside the rule for the handful of terms that need it. Either way, the docstring's correctness claim should be softened or backed by a test parametrised over GADM_ADMIN_TERMS.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -506,6 +507,42 @@ async def check_duplicate_aois(
...
+LEVEL_TO_SUBREGION: Dict[int, str] = {
+    1: "state",
+    2: "district",
+    3: "municipality",
+    4: "locality",
+    5: "neighbourhood",
+}
+
+
+def _resolve_effective_subregion(
```
01painadam, line 523: Beyond this file there is no eval coverage for the LLM half of the feature. Everything tested here is deterministic.

Let's run a few evals against this for known cases it should solve

01painadam (changes requested):

yellowcap:
Not 100% but still quite good. Could be that the 2 failing ones, i'll inspect more 

```
EVALUATION SUMMARY
Environment:           localhost
API Base URL:          http://localhost:8000
Run timestamp:         2026-08-31 15:11:42 UTC
GNW code version:      2026.8.31
GNW prompts version:   unknown
GNW agent LLM:         primary gemini-flash (ChatGoogleGenerativeAI), small gemini-flash (ChatGoogleGenerativeAI)
Eval judge LLM:        Anthropic / claude-haiku-4-5
API latency (s):       avg=48.6, std=30.3, min=12.3, max=201.5 (n=95)

Tests Run (after filters): 95

Agent Answer:              50 /  62 (0.81)

AOI ID Match:              69 /  71 (0.97)
Dataset ID Match:          68 /  80 (0.85)
Dataset Parameter Match:    1 /   1 (1.00)
Context Layer Match:        6 /   7 (0.86)
Data Pull Exists:          58 /  65 (0.89)
Date Extraction:           31 /  37 (0.84)
Date Coverage (info):      32 /  37 (0.86)
Charts Answer:             43 /  51 (0.84)
Expected Text Match:       20 /  24 (0.83)
Clarification Requested:    2 /   2 (1.00)
Suggested Datasets:         0 /   3 (0.00)
Nudge Match:                2 /   2 (1.00)

Dashboard Created:          6 /   6 (1.00)
Dashboard AOI Match:        4 /   5 (0.80)
Dashboard Widgets Match:    2 /   4 (0.50)
Dashboard Widgets Valid:    2 /   5 (0.40)

(warning: overall_score is experimental and untested)
Tests with overall score ≥0.7:   75 /  95 (78.9%)
```

yellowcap:
The 2 AOI failures are:                                                                                                                                                       

  1. 1-099 — "Create a dashboard for the city of Bern, Switzerland and add a satellite imagery map to it" (aoi id match: 0.0, dashboard aoi match: 0.0)                          
  2. 1-104 — "Show me satellite imagery for Madrid, Spain for the first two weeks of January 2021" (aoi id match: 0.0)

yellowcap:
Both failures are the same underlying defect, and neither is caused by the admin-depth work in this branch.

1-099 — "city of Bern, Switzerland"
- Expected CHE.6.3.1_1 (commune/municipality), got CHE.6_1 (canton)
- The word "city" never reaches anything that could use it: AreaOfInterestType has no depth-aware value, so it's dropped during extraction, and the default selection path (score_best_aoi → best_candidate_row in scoring.py) never sees the raw user question at all — only extracted place strings.

1-104 — "Madrid, Spain"
- Expected ESP.8_1 (region, "Comunidad de Madrid"), got ESP.8.1_1 (coterminous province, leaf name literally "Madrid")
- best_candidate_row's hierarchy table favors state-province (0.9) over district-county (0.7), but the exact-leaf-name bonus (+0.2) goes to the province since its stored name segment matches "Madrid" exactly while the region's is "Comunidad de Madrid" — the bonus outweighs the hierarchy preference.

Root cause (shared): best_candidate_row in src/agent/subagents/pick_aoi/scoring.py is pure string-similarity + fixed-hierarchy-weight math with zero access to the user's original wording. The one function that does read the raw question and could disambiguate via context (select_best_aoi, tool.py:314) is only used on the normalized=False legacy fallback path, not the default.

Why it's not this branch: git blame on score_best_aoi/select_best_aoi (tool.py:298-422) and scoring.py shows no lines touched by this branch's two commits (3198a20f, 01af2f51). This branch's admin-depth resolution (_resolve_effective_subregion, admin_term) only activates inside the subregion comparison path (if subregion: in tool.py) — it doesn't run for single-place picks like these two queries, so it can't be responsible for, or fix, this gap.

Recommendation: File as a separate bug against best_candidate_row's level-selection logic — likely needs either an admin-level hint threaded from the user's phrasing, or a tie-break that doesn't let the exact-leaf-name bonus override the hierarchy preference for coterminous region/province pairs.

01painadam (approved):
Looks good - thanks for addressing the changes. I think this stuff could use a specific AOI selection test set so we can easier see where to gaps are (and address them) for this specific tool. I'll work on something in the gnw-gold-evals repo

PR #802, opened 2026-08-26
Removing DIST dataset completely from agent.

danscales:
AJ indicated that we should really just remove DIST alerts completely as a dataset.  This will make things much less ambiguous for the users and the evals, especially when a phrase like "disturbance alerts" is used.

I removed the dataset YML file, adjusted a bunch of code, documentation, and tests. I haven't yet gone through all the evals that might need to change.


jterry64 (approved):
@danscales LGTM! I added the label to run the tool tests just to make sure we caught all the ones that used DIST still. I think you can merge before fixing all the evals if you do it shortly after.

danscales:
FYI, as far as I can tell, all the failures in the tools tests existed before this change.  As Justin said, I'll check in and we'll work on fixing up the evals afterward. (AJ already has a change to fix up the gold set evals so DIST is removed from all test answers..)

PR #806, opened 2026-08-28
Default insights for remaining datasets

jterry64:
Adding default insights for remaining datasets following the existing patterns. These were based off the YAML, and below are pictures of what they all look like. 

The only one I'm unclear about is Tree Cover, which is just a single number, so it's hard to say if it should have a chart. The YAML mentions it being binned by density threshold, but that's not currently supported by analytics API.

### Global land cover

01-land-cover

### SBTN Natural Lands Map

02-natural-lands

### Global natural/semi-natural grassland extent

03-grasslands

### Tree cover gain

04-tree-cover-gain

### Tree cover

05-tree-cover

### Tree cover loss by dominant driver

06-tcl-by-driver

### Tree cover loss due to fires

07-tcl-by-fires

### Forest greenhouse gas net flux

08-forest-ghg-flux


01painadam (approved):
Looks good to me! Maybe Tree Cover could be a Pie (since in some cases it may end up bring TCL in primary/ifl etc?)

solomon-negusse (approved):
Looks good!

yellowcap (changes requested):
@jterry64 this is great, love to see that we get more curated charts! A couple of things I would improve:

- There is a lot of code duplication in the classes. `__init__` and `can_handle` look the same for all of them, and the `present = sorted(....` part is also used 4 times identical. Maybe unify this a bit more?
- Generative insights are translated, the curated ones here are hard coded in english. I think this should be relatively easy to add. The principle in the generative charts is to use `t()` for hard coded text for a fixed set of languages, and use gemini small for translating dynamically if not already existing in fixtures.
- I would use line charts for tcl and grasslands charts which are yearly for longer time series.

jterry64:
Thanks for the feedback @yellowcap. I agree with your first two, but I believe our "style" has always been to show yearly TCL as a bar chart across many of our products, and so I assume the same for grasslands. While it'd normally be line chart for a time series, but it's possible the researchers wanted it clear we only know how much happened per year and not continuously over the year. But I'll defer to @01painadam if we want to change up how we normally do it.

yellowcap (approved):
@jterry64 nice, thanks for the clarification and updates. The line chart question is more of a detail, and if we have prior art in this, let's keep it that way!

One of the reasons I mentioned this here is that the agent also returns bar charts, and in the back of my mind I always wanted to change that. But now I see that is actually desired 😄

01painadam:
Agree - its a house style thing.

@yellowcap I think @jterry64 is out - could you bump the ver and merge?

yellowcap:
rebased from main, updated version, will merge once tests passed

PR #808, opened 2026-08-31
PZB-1293: Dashboard sections — data model, repository, API, and agent

yellowcap:
Adds sections to dashboards: one flat level of grouping, backend only. The
agent side is a follow-up PR (PZB-1294) that stacks on this branch.

Corresponding frontend PR is in https://github.com/wri/project-zeno-next/pull/655

## Model

`dashboard_sections` (`id`, `dashboard_id`, `title`, `description`,
`position`, `created_at`) plus a nullable `dashboard_widgets.section_id`.

- **One flat level.** A widget belongs to at most one section. No nesting.
- **Ungrouped is a first-class state**, not a migration convenience. A widget
  with `section_id = NULL` sits at the top level and renders above the first
  section.
- **`position` is per container** — order within a section, or within the
  ungrouped top-level list. Sections carry their own `position`.
- **Deleting a section never deletes content by default.** Its widgets fall
  back to ungrouped.

## Endpoints

| Method | Path |
|---|---|
| POST | `/api/dashboards/{id}/sections` |
| PATCH | `/api/dashboards/{id}/sections/{section_id}` |
| DELETE | `/api/dashboards/{id}/sections/{section_id}[?delete_widgets=true]` |

`DashboardResponse` gains `sections`; each widget carries `section_id`.
Widget create and update accept `section_id`. On update the field is
three-valued: omitted keeps the current grouping, an id moves the widget, an
explicit `null` moves it back to the top level. The router tells the last two
apart with Pydantic's `model_fields_set`; the repository uses an `UNSET`
sentinel. A `section_id` belonging to another dashboard is a 404.

## Two things worth a reviewer's attention

**No default section, and no backfill.** The ticket title anticipates both.
This implements the nullable-`section_id` shape instead, which was the
explicit ask: existing widgets simply *are* the ungrouped top-level list, so
the migration adds a column and a table and touches no rows. Worth
reconciling the ticket wording. If a real "Ungrouped" section row is wanted
later, it can be added without a schema change.

**Renumbering on ungroup.** Positions are container-scoped, so a section's
widgets cannot keep their numbers when the section goes — ungrouped `0,1`
plus a section's `0,1` would give `0,0,1,1`. `_ungroup_widgets` appends them
as a block after the current last top-level widget, preserving their relative
order. The DB test asserts the exact resulting positions.

## Known limits, deliberate

- **The `widgets` array in the response is flat**, ordered by `position`
  across the whole dashboard, so it interleaves containers. The client must
  group by `section_id` and order within each group. The data is complete and
  unambiguous, but the shape invites a naive `widgets.map(...)` that renders
  scrambled. Nesting widgets inside each section in the response is a
  reasonable alternative if the frontend would prefer it — say so and I'll
  change it before this merges.
- **Positions can tie.** No uniqueness constraint, and setting one does not
  shift siblings, so inserting at 0 creates a second 0. A true reorder means
  rewriting each affected widget's position. This is pre-existing widget
  behaviour, carried over to sections rather than changed.

## Migration

`c9f1a2b3d4e5`, on top of `d4a1c7b93e02`. Applied, rolled back and re-applied
against a local Postgres; the resulting schema was checked by hand.

## Tests

611 pass on this branch alone (`tests/unit`, `tests/api/test_dashboards.py`,
the dashboard agent files). New coverage: section CRUD and ownership, widget
placement, per-container positions, moves in both directions, renumbering on
ungroup, `delete_widgets` both ways, and cross-dashboard section rejection.

Also fixes two tests that main left red when `dashboard_name` was added to
the `dashboard_updated` signal in f7298bec. Unrelated to sections, but they
sit in files this PR touches.

`tests/agent/test_graph.py::test_agent_for_disturbance_alerts_for_brazil`
fails on this branch and on main: it queries the DIST dataset that #802
removed. Not touched here.

https://claude.ai/code/session_01KaUVQViDyCfrSYcSEwW1Jd


src/agent/tools/move_dashboard_widget.py
```diff
@@ -0,0 +1,145 @@
...
+            tool_call_id,
+        )
+
+    target_section = None
+    if section:
+        target_section, message = resolve_section(dashboard, section)
+        if message:
+            return error_command(message, tool_call_id)
+        if target_section.id == widget.section_id:
+            return error_command(
```
01painadam, line 92: This guard fires even when position is supplied, so the message "Pass a position if you meant to reorder it" describes behaviour the tool does not have: a same-section move with a position is rejected, and reordering within a container is impossible from the agent (the prompt fragment promises "reorder"). 

Suggest if `target_section.id == widget.section_id and position is None:` and the same for the ungrouped branch, plus a test for the reorder case.

01painadam (changes requested):

src/agent/tools/common.py
```diff
@@ -90,6 +91,40 @@ async def load_editable_dashboard(
...
+    the id only from an earlier tool reply. ``None`` means "ungrouped", which
+    is not an error: ``(None, None)``. An unmatched name is an error that
+    lists the real sections so the model can retry or create one.
+    """
+    if not section:
+        return None, None
+    wanted = str(section).strip()
+    for row in dashboard.sections or []:
+        if str(row.id) == wanted or row.title.casefold() == wanted.casefold():
+            return row, None
```
01painadam: Hmmm, first match on title wins.

The REST API does not enforce unique section titles per dashboard, so two sections named the same (created in the UI) make this silently pick one. Either error when more than one title matches and list the ids, or add a uniqueness rule on the API/DB side so the agent-side duplicate check in add_dashboard_section is not the only guard.

src/agent/tools/add_dashboard_section.py
```diff
@@ -0,0 +1,144 @@
...
+def _normalize(text: Optional[str]) -> Optional[str]:
+    """Strip surrounding whitespace; None when nothing remains."""
+    if not text:
+        return None
+    return text.strip() or None
+
+
+@tool("add_dashboard_section")
+async def add_dashboard_section(
+    title: str,
```
solomon-negusse: Would it make sense to enforce character limit with Annotated string type instead of the "short" prose?

src/agent/tools/edit_dashboard_section.py
```diff
@@ -0,0 +1,142 @@
...
+logger = get_logger(__name__)
+
+
+def _select_section(dashboard, section: Optional[str]):
+    """The section to edit, as ``(section_orm, error_message)``.
+
+    With no `section` argument the dashboard's only section is the obvious
+    target; with several the error lists them so the model can retry.
+    """
+    if section:
```
solomon-negusse, line 39: ~Since you're only calling `resolve_section` when a section is present, would it make sense to make `section` a required argument to `resolve_section`. Feels a bit redundant that we pass `None` section to it for it to exit early with `None`.~
I get it now

solomon-negusse (approved):
This looks good to me (besides what AJ pointed out) - just have a couple of non-blocking suggestions.

src/agent/tools/move_dashboard_widget.py
yellowcap: Fixed in 2fc5fb3e. Both guards now fire only when `position is None`, so naming the container the widget is already in plus a `position` reorders it in place. The success reply says "to position N in ..." for that case, and the tool docstring/prompt fragment now state it. Added `test_move_widget_reorders_within_its_own_section` and `test_move_widget_reorders_at_the_top_level`.

src/agent/tools/common.py
yellowcap: Fixed in 2fc5fb3e. `resolve_section` now matches ids first, then collects every title match: exactly one resolves, more than one is an error that lists the candidate ids ("Dashboard X has 2 sections titled Y. Name the one you mean by id instead of by title: ..."). I kept the DB/API free of a uniqueness rule so existing dashboards and the UI stay unaffected — the agent asks for an id instead of guessing. Covered by `test_move_widget_refuses_an_ambiguous_section_title` and `test_move_widget_takes_an_id_over_a_duplicate_title`.

src/agent/tools/add_dashboard_section.py
yellowcap: Good call — done in 2fc5fb3e. `title` is now `Annotated[str, Field(max_length=SECTION_TITLE_MAX_LENGTH)]` in both `add_dashboard_section` and `edit_dashboard_section`, so the cap reaches the model as `maxLength` in the tool schema and an over-long title fails validation before the tool body runs. The same constant caps `DashboardSectionCreateRequest.title` and `DashboardSectionUpdateRequest.title` (100 chars, matching the other name caps in schemas.py), so the REST API enforces it too.

01painadam (approved):

PR #811, opened 2026-09-01
PZB-1428 feat(pick_aoi): log geocoding misses

yellowcap:
## Summary
- `pick_aoi`'s three miss paths (no place extracted, zero DB candidates for any place, partial matches) all return `ToolMessage(status="success", response_metadata={"msg_type": "human_feedback"})`. That's intentional — the LLM should treat these as a clarifying question, not a tool error — but it also means misses currently leave no trace anywhere. In particular `src/api/services/langfuse/parse.py` derives `tool_error_count` purely from `status == "error"`, so total geocoding misses are invisible to that metric today.
- Adds structured `logger.warning("geocoding_miss", ...)` / `logger.warning("geocoding_partial_miss", ...)` calls at each of the three miss branches in `src/agent/subagents/pick_aoi/tool.py`, carrying the question and unmatched/matched place names. `user_id`/`thread_id`/`request_id` are already bound to structlog's contextvars upstream, so they show up on these lines automatically.
- No change to the tool message contract or `status` field — this is additive observability only.

## Q&A

**Q (from a colleague):** After looking through this code, it seems there's no observable way to detect geocoding misses/issues in production. It even looks like total misses come back as "success" in the tool message — is that right?

**A:** Yes, that's correct. All three miss paths in `src/agent/subagents/pick_aoi/tool.py` — no place extracted, zero DB candidates for any place, and partial matches — return `status="success"`, tagged `msg_type="human_feedback"`. That's deliberate on the LLM-contract side: the agent is meant to treat these as a clarifying question to the user rather than a tool failure. But it means a total geocoding miss is indistinguishable from a genuine success at the `status` field level, and it's specifically why `tool_error_count` in `src/api/services/langfuse/parse.py` (which keys off `status == "error"`) never counts these misses. This PR adds dedicated structured logging at each miss branch so they're observable without changing that LLM-facing contract.

## Test plan
- [x] `pytest tests/unit/agent/tools/pick_aoi/test_tool.py -q` — 35 passed, including 3 new tests asserting the log calls fire with the expected fields for each miss branch
- [x] `pytest tests/tools/test_pick_aoi.py -q` — 21 passed (no regressions)
- [x] `pytest tests/unit -q` — 616 passed (no regressions)
- [x] `ruff check` and `ruff format --check` on changed files

Rebased onto `main` (`e6f8d726`). The branch previously carried an unrelated merge commit that swept in a dashboard-section migration, `docs/nrt-monitoring-section-plan.md`, `src/agent/datasets/layers.py`, and a stale version bump in `pyproject.toml`/`uv.lock`. Those are gone: the PR is now one commit touching only `pick_aoi/tool.py` and its unit test.

https://claude.ai/code/session_01MTmz33SBvinHBpd3wLP25X


gtempus (approved):
Thank you, @yellowcap!

PR #813, opened 2026-09-03
feat(dashboards): add near-real-time monitoring sections

yellowcap:
Stacked on #808 (dashboard sections) — **review that first**; this PR's base
is `pzb-1293-dashboard-sections-api`, so the diff here is only the NRT work.

## What this adds

One call builds a **near-real-time monitoring section** on a dashboard: an
alerts chart, a map of those alerts, and satellite imagery of the same area
and period. Meant for a one-click button on the frontend.

```
POST /api/dashboards/{id}/sections/nrt-monitoring        {"days": 14}
POST /api/dashboards/{id}/sections/{section_id}/refresh  {"days": 30}
```

| Position | Widget | Content |
|---|---|---|
| 0 | `insight` | Daily alert area (ha) by confidence tier |
| 1 | `map` | Integrated alerts layer for the period |
| 2 | `map` | Sentinel-2 mosaic for the period end |

Title and description are written by a model from the chart data — the
description is the summary that states the key figures — with a templated
fallback so a model failure never loses a section whose data is gathered.

**No new widget or config shapes**, which is why the section renders with no
frontend change. Everything is additive: `type` and `config` on sections,
three fields on this endpoint's response, `tile_url`/`tilejson_url` on the
mosaic response. An old client ignores all of it and still renders.

## Decisions worth reviewing

**Synchronous, not a job.** The endpoint does the whole build before it
answers — tens of seconds. One request, one render, no polling; chosen
deliberately over a job + `Retry-After` loop. It also keeps the
request-scoped user id bound for the custom-AOI geometry lookup.

**Sealed sections.** A section's content is a record of one build, so
editing it would make its own title untrue. The guard lives in
`dashboard_writer`, not the router, because the agent tools bypass routers
entirely; the router maps `SealedSectionError` to 409. Three back doors were
closed explicitly: deleting a section now always cascades for sealed types
(ungrouping would leave loose, editable widgets), `type` is not in the update
schema (or unsealing is one PATCH), and `update_insight_display` refuses an
insight a sealed widget references (it could rewrite the section's content
without touching a dashboard row).

**Layout is exempt.** `position`, and a config replacement whose only
differing keys are `size`/`sizes` — the frontend's own single/double setting.
The check is a **diff**, not a field whitelist: PATCH replaces config
wholesale, so "resize it" and "resize it and swap the tile_url" are the same
shape of request. Both sides are compared in the stored, relativized form.

**Two weeks, daily.** `DEFAULT_DAYS = 14`. That forced the alerts chart to
daily buckets — monthly ones collapse a fortnight to one or two points, and
daily is the resolution the analytics API returns. Days with no alerts stay
absent rather than zero-filled.

**Refresh rebuilds, it does not edit.** The recipe re-runs for the new
window and swaps the section's contents in one transaction, keeping the
section row so its id and place hold. The window lives on
`dashboard_sections.config` rather than being inferred from a widget's tile
dates — which also retired the double-click guard's config sniffing, so a
second recipe over the same dataset can't collide with this one. The previous
chart insight is deleted once nothing points at it.

**The agent's window change is gated by a rule, not a prompt.**
`update_nrt_monitoring_section` refuses to act without `confirmed=True`,
legitimate only after a `send_nudge` the user answered. A window change
replaces every figure on screen and deletes the previous ones; a model can
retry past an instruction, not past a refusal.

## Refactors that carry their own weight

- **`resolve_dataset_layer()`** (`src/agent/datasets/layers.py`) lifts the
  tile-URL rules out of the dataset-selection subagent, where they needed an
  LLM selection object and a pandas row. `get_tile_services_for_dataset()` is
  now an adapter over it, passing its *candidate* row so the extent filter on
  context layers is not quietly undone.
- **`widget_configs.py`** holds the map-widget snapshot projections that
  lived inside `add_map_widget`, so a layer added from chat and one added by
  a recipe are byte-identical.
- **`_visible_insights()`** shares the insight-expansion rule, so a widget's
  payload doesn't depend on which endpoint returned the dashboard.

## Verified live, not only mocked

Run against the real analytics API, real STAC/S3 and a real model call, which
found two bugs the mocked tests could not:

1. `AnalyticsHandler.pull_data` **mutates its input** (strips the GADM level
   suffix). One shared AOI dict meant every section silently lost its
   satellite widget and blamed "Could not load the geometry". Each consumer
   now gets a copy, and the test stand-in mutates like the real handler.
2. The generated description quoted raw analytics floats
   (`153409.28893796972 ha`). The prompt asks for report-style rounding and
   the chart copy the model reads is rounded; the stored chart keeps its
   precision.

Also verified: 14-day build → refresh to 60 days keeps the section id, moves
every widget's dates together, and deletes the stale insight; both migrations
round-trip up and down.

## Tests

1027 pass. `tests/api/test_nrt_monitoring.py` (built section, imagery
degradation, analytics failure, double-click guard, `force`, refresh, auth),
`tests/api/test_dashboards.py` (section `type`, one test per blocked action,
layout allowed, the smuggling refusal), plus unit tests for the layer
resolver, the summary fallbacks and rounding, the agent-side seal and the
enforced confirmation.

The one failing test on this branch is pre-existing and unrelated:
`test_metadata_version_matches_pyproject` compares installed package
metadata against `pyproject`, and the venv needs a `uv sync`.

## Not in scope

Scheduled refresh (on demand only, so a section ages honestly), multi-dataset
recipes, subscribe-to-alerts, and a dedicated imagery-widget endpoint —
though imagery widgets are now creatable in two calls, since the mosaic
endpoint returns renderable URLs.

## Frontend

`docs/nrt-monitoring-handoff.md` is the contract: the one call, the timing,
degradation to two widgets, which controls to hide inside a sealed section,
and which to keep (reorder, resize). Design notes and open points in
`docs/nrt-monitoring-section-plan.md`.

**Chat is gated:** the agent tools live in the `nrt-monitoring` skill, which
only the experimental profile loads, and `ff` itself requires an
admin/superuser/machine user. The REST endpoints are not gated.

PZB-1295


yellowcap:
Closed in favor of #822

PR #814, opened 2026-09-03
fix(traces): account for llm usage outside the agent loop

yellowcap:
## Why

We use Langfuse to measure cost per query, and the number was wrong in two independent ways.

`langfuse_traces.turn_tokens` was never a Langfuse figure. `parse_trace` summed `usage_metadata` off the AIMessages in `trace.output`, and only the agent's own model calls put an AIMessage into `AgentState` — a tool that calls an LLM returns a ToolMessage. So every tool-internal LLM call was counted nowhere, **even the ones Langfuse had traced correctly all along**. Meanwhile `total_cost` was `trace.totalCost`, which aggregates all observations. The two columns measured different scopes, and `trace.totalCost` turns out to be null on real traces anyway — so cost per query was not merely understated, it was absent.

Measured on one production trace (`62168a8a1b9269b9d95fdadcbd8dccc4`):

| owner | tokens | cached | cost |
|---|---|---|---|
| agent (5 calls) | 35,466 | 16,092 | $0.011132 |
| `pick_aoi` | 1,479 | 0 | $0.000842 |
| `pick_dataset` | 5,718 | 0 | $0.003327 |
| `generate_insights` → text generator | 1,828 | 0 | $0.001304 |
| **trace total** | **44,491** | 16,092 | **$0.016604** |

The parser recorded 35,466 — identical to the agent-only sum, to the token. It dropped 20% of tokens and **33% of cost**.

Cost share exceeds token share because of caching: agent calls are ~45% prompt-cache reads, every tool call is `input_cache_read: 0`. Tool calls are disproportionately expensive per token, so dropping them skews cost harder than tokens.

Separately, `generate_insights` showed only its stage-2 text call. Stage 1 — `GeminiCodeExecutor`, which calls `google.genai` directly through `run_in_executor` — produced **no observation at all**, despite `codeact_parts` proving it ran two model turns. No LangChain runnable means no callback.

### How much the new instrumentation adds

Unknown, and unknowable until it ships — that is the nature of the gap. The 20%/33% above is what the *parser* was dropping from observations Langfuse already had; the code executor emitted nothing to measure.

A **floor** can be reconstructed from the repo plus this trace, though. Rebuilding the executor's own prompt via `build_analysis_prompt` and counting the code and execution output it echoed back:

| | tokens (~) |
|---|---|
| prompt skeleton, dataset guidance stripped | 2,359 |
| code + execution output echoed back | 920 |
| model turns in the code-exec loop | 2 |
| **floor, prompt resent per turn** | **~5,600** |

That is **~13% on top of the 44,491 this PR now counts**, and it is a floor in three ways: the skeleton excludes the dataset's guidelines, cautions and code instructions; it excludes the inlined CSV (this trace's `statistics.data` was empty, loaded from the DB by id, so its real size is not recoverable); and this was the simplest possible query — one dataset, one year, one AOI.

It also understates cost more than tokens, because the executor runs on `CODING_MODEL`, whose default is `gemini-3.1-pro-preview` — materially pricier per token than the `gemini-3-flash-preview` everything else in this trace used. Worth confirming which model production actually has configured.

Treat ~13% as the lower bound on insight turns and expect the real figure to be higher. `derived.cost_by_component` will report it exactly once this is deployed.

## What changed

**Token columns read Langfuse usage.** `parse_observations` aggregates the trace's generations. `turn_*` now covers every LLM call in the turn; new columns carry the split (`agent_tokens`, `tool_tokens`, `agent_cost`, `tool_cost`, `generation_count`), and `derived.tokens_by_component` / `cost_by_component` break the tool half down per tool. `agent_tokens + tool_tokens == turn_tokens` by construction. `total_cost` falls back to the observation sum when `trace.totalCost` is null.

Langfuse reports `input` net of cached tokens; LangChain's `input_tokens` includes them. The parser adds them back, or the series would silently drop by the cache-hit rate.

**Observations are fetched per window, not per trace** — a handful of requests instead of N — reusing the existing retry/page-halving machinery, with a 30-minute upper pad for traces straddling a window boundary. If that fetch fails the window still ingests on message-derived usage; `derived.usage_source` records which, and `agent_tokens`'s fill rate in the run row is the alert.

**The code executor is instrumented.** A Langfuse generation span around `_call_model` reports `usage_metadata`, mapping Gemini's keys onto Langfuse's (cached tokens split out of `input`, thinking tokens folded into `output`). One span covers the retry loop, so its latency is the wall clock the caller waited. Not rewritten onto `ChatGoogleGenerativeAI`: the executor depends on `executable_code` / `code_execution_result` / `inline_data`, which LangChain models poorly.

**Out-of-turn calls are traced.** Thread naming, custom-area naming and language detection each ran in their own request with no ambient handler. They now get `auxiliary_config(...)` — their own handler, tagged `auxiliary`, joined to the thread by session id. Ingest drops tagged traces: they carry no `AgentState`, so they would land as zero-token `EMPTY` rows and dilute every per-turn average. Their cost stays visible in Langfuse under the thread's session.

**Chains are named** (`select_aoi`, `select_dataset`, `search_blogs_agent`, …) so components are attributable. Verified that passing `run_name` alone still inherits ambient callbacks.

**`langfuse-model-prices`** reports models Langfuse records usage for but prices at zero — an unrecognised model name silently zeroes that model's whole share of cost.

## Two notes for review

**A bug I introduced and caught.** I first fetched only `type=GENERATION`. But the agent/tool label lives on a generation's **ancestors** — both an agent call and a tool's call are named `ChatGoogleGenerativeAI` — so the parent walk found nothing and the whole turn read as tool spend. Ingest now fetches all types; `test_component_attribution_needs_the_ancestor_spans` pins it. Related: `derived.usage_split_agrees` compares the observation-derived `agent_tokens` against the independent message-derived figure, so if LangGraph renames its `model` node, drift surfaces as `False` instead of a silent zero — which a fill-rate check cannot catch, because zero is not null.

**`trace.output` arrives double-JSON-encoded.** `parse_trace` required a dict, and `parse_state` classified a non-dict output as "not applicable" rather than a violation — so it would zero every metric *and* keep the drift monitor quiet. `parse_trace` now decodes JSON-string `output`/`input`. The raw downloaded export parses completely, including `aoi_name` and `primary_dataset_name` that a string output had been discarding.

## Migration

`f3a8c1d5b204` adds five nullable columns and recreates `langfuse_traces_analytics`. Verified up → down → up on a throwaway database.

`PARSER_VERSION` → 3. **`turn_tokens` changes meaning**: it steps up for any turn whose tools call an LLM. Re-run `ingest-langfuse-traces --backfill --since` over the window you analyse, or the series has a step where the parser version changes.

## Testing

661 unit + 354 API tests pass. New: `test_langfuse_fetch.py` (8, the fetch client had none), `test_langfuse_tracing.py` (12), `test_gemini_executor_tracing.py` (4), plus 17 in the parse/ingest suites. Validated end-to-end by running the real production trace through `build_row`: `turn_tokens` 35,466 → 44,491, `total_cost` NULL → $0.016605 ($0.011132 agent / $0.005473 tool).

## Known remaining gap

Embeddings are still unaccounted. `pick_dataset`'s retriever call goes through LangChain's `GoogleGenerativeAIEmbeddings`, so it *is* traced — it appears as a `RETRIEVER` observation — but LangChain embeddings report no token usage, so Langfuse records it with empty usage and zero cost (visible in this trace as `[RETRIEVER] VectorStoreRetriever usage={} cost=0`). It is one short query embedding per `pick_dataset` call, so the amount is small, but it is not zero and this PR does not fix it.

## Not in scope

`src/api/services/nrt_summary.py` is on an unmerged branch; its summary call still needs a `run_name` when that lands.

### DIFF
--- a/db/alembic/versions/f3a8c1d5b204_add_llm_usage_split_to_langfuse_traces.py
+++ b/db/alembic/versions/f3a8c1d5b204_add_llm_usage_split_to_langfuse_traces.py
@@ -1,0 +1,136 @@
+"""add agent/tool llm usage split to langfuse_traces
+
+Revision ID: f3a8c1d5b204
+Revises: c9f1a2b3d4e5
+Create Date: 2026-09-03 11:20:00.000000
+
+Token columns used to be derived from the AIMessage stream, which only carries
+the calls the agent made itself — an LLM call inside a tool returns a
+ToolMessage and so was counted nowhere. Parser v3 sources them from the trace's
+generation observations instead and records the split:
+
+* ``agent_tokens`` / ``agent_cost`` — what the agent spent itself (this is what
+  ``turn_tokens`` used to hold).
+* ``tool_tokens`` / ``tool_cost`` — the remainder, spent by LLM calls inside
+  tools. Roughly a third of cost on the measured production trace.
+* ``generation_count`` — how many LLM calls the turn made in total.
+
+``turn_tokens`` steps up for any turn whose tools call an LLM, so re-run
+ingestion over the window you care about (``ingest-langfuse-traces --backfill
+--since``) to put existing rows on the new basis. The analytics view is
+recreated to expose the new columns.
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "f3a8c1d5b204"
+down_revision: Union[str, None] = "c9f1a2b3d4e5"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+_TABLE = "langfuse_traces"
+_VIEW = "langfuse_traces_analytics"
+
+_NEW_COLUMNS = (
+    ("agent_tokens", sa.Integer()),
+    ("tool_tokens", sa.Integer()),
+    ("agent_cost", sa.Float()),
+    ("tool_cost", sa.Float()),
+    ("generation_count", sa.Integer()),
+)
+
+# Recreated (not CREATE OR REPLACE'd) because adding columns to a view's select
+# list changes its shape, which REPLACE rejects.
+_CREATE_VIEW = f"""
+CREATE VIEW {_VIEW} AS
+SELECT
+    id,
+    session_id,
+    user_id,
+    environment,
+    trace_timestamp,
+    outcome,
+    has_answer,
+    answer_is_refusal,
+    had_tool_call,
+    tool_error_count,
+    aoi_name,
+    aoi_type,
+    primary_dataset_name,
+    has_insight,
+    is_global,
+    insight_id,
+    turn_input_tokens,
+    turn_output_tokens,
+    turn_tokens,
+    turn_tool_calls,
+    agent_tokens,
+    tool_tokens,
+    agent_cost,
+    tool_cost,
+    generation_count,
+    latency_seconds,
+    total_cost,
+    prompt,
+    (
+        row_number() OVER (
+            PARTITION BY COALESCE(session_id, id)
+            ORDER BY trace_timestamp DESC NULLS LAST, id DESC
+        ) = 1
+    ) AS is_final_turn_in_thread
+FROM langfuse_traces;
+"""
+
+_PRIOR_VIEW = f"""
+CREATE VIEW {_VIEW} AS
+SELECT
+    id,
+    session_id,
+    user_id,
+    environment,
+    trace_timestamp,
+    outcome,
+    has_answer,
+    answer_is_refusal,
+    had_tool_call,
+    tool_error_count,
+    aoi_name,
+    aoi_type,
+    primary_dataset_name,
+    has_insight,
+    is_global,
+    insight_id,
+    turn_input_tokens,
+    turn_output_tokens,
+    turn_tokens,
+    turn_tool_calls,
+    latency_seconds,
+    total_cost,
+    prompt,
+    (
+        row_number() OVER (
+            PARTITION BY COALESCE(session_id, id)
+            ORDER BY trace_timestamp DESC NULLS LAST, id DESC
+        ) = 1
+    ) AS is_final_turn_in_thread
+FROM langfuse_traces;
+"""
+
+
+def upgrade() -> None:
+    for name, type_ in _NEW_COLUMNS:
+        op.add_column(_TABLE, sa.Column(name, type_, nullable=True))
+    op.execute(f"DROP VIEW IF EXISTS {_VIEW};")
+    op.execute(_CREATE_VIEW)
+
+
+def downgrade() -> None:
+    op.execute(f"DROP VIEW IF EXISTS {_VIEW};")
+    op.execute(_PRIOR_VIEW)
+    for name, _ in reversed(_NEW_COLUMNS):
+        op.drop_column(_TABLE, name)
--- a/src/agent/subagents/pick_aoi/tool.py
+++ b/src/agent/subagents/pick_aoi/tool.py
@@ -348,10 +348,12 @@ async def select_best_aoi(
         ]
     )
 
-    # Chain for selecting the best location match and returning the src_id
+    # Chain for selecting the best location match and returning the src_id.
+    # run_name is what this call is labelled in Langfuse, and so what
+    # derived.cost_by_component attributes its spend to.
     AOI_SELECTION_CHAIN = (
         AOI_SELECTION_PROMPT | SMALL_MODEL.with_structured_output(AOIId)
-    )
+    ).with_config(run_name="select_aoi")
 
     selected_aoi_index = await AOI_SELECTION_CHAIN.ainvoke(
         {
@@ -747,7 +749,7 @@ class Geocoder:
         chain = (
             GEOCODER_EXTRACTION_PROMPT
             | SMALL_MODEL.with_structured_output(PlaceQuery)
-        )
+        ).with_config(run_name="extract_place_query")
         return await chain.ainvoke({"question": question})
 
     async def lookup(
--- a/src/agent/tools/update_insight_display.py
+++ b/src/agent/tools/update_insight_display.py
@@ -173,7 +173,9 @@ async def update_insight_display(
         )
 
     current = Insight.from_orm_row(row)
-    revised = await InsightDisplayReviser().revise(current, instruction)
+    revised = await InsightDisplayReviser().revise(
+        current, instruction, config={"run_name": "revise_insight_display"}
+    )
 
     try:
         updated = _apply_revision(current.charts, revised)
--- a/src/api/cli.py
+++ b/src/api/cli.py
@@ -812,6 +812,85 @@ def backfill_turn_fields_command(batch_size: int, dry_run: bool):
     asyncio.run(_run())
 
 
+@cli.command("langfuse-model-prices")
+@click.option("--hours", type=int, default=24, help="Look-back window.")
+@click.option("--environment", default=None, help="Filter to one environment.")
+def langfuse_model_prices_command(hours: int, environment: Optional[str]):
+    """Report which models Langfuse is failing to price.
+
+    Langfuse computes cost by matching an observation's model name against its
+    model-definition table. A model it does not recognise — a preview name, or a
+    newly configured one — is recorded with usage but **zero cost**, so
+    cost-per-query silently understates by that model's whole share, with
+    nothing in the data to say so.
+
+    For every model flagged UNPRICED here, add a model definition with its
+    prices in Langfuse (Settings → Models), then re-run
+    ``ingest-langfuse-traces --backfill --since`` over the affected window to
+    recompute the stored costs.
+    """
+    from src.api.services.langfuse.fetch import LangfuseClient
+
+    until = datetime.now(timezone.utc)
+    since = until - timedelta(hours=hours)
+    client = LangfuseClient.from_env()
+    # Every type, then filter on "has a model and reported usage": embeddings
+    # are billable too, and an unpriced embedding model is just as invisible as
+    # an unpriced chat one.
+    observations = client.fetch_observations_window(
+        since, until, environment, obs_type=None
+    )
+
+    # model -> [calls, tokens, cost, priced_calls]
+    stats: dict = {}
+    for o in observations:
+        model = o.get("model")
+        usage = o.get("usageDetails") or {}
+        if not model or not usage:
+            continue
+        cost = o.get("totalCost") or 0
+        row = stats.setdefault(model, [0, 0, 0.0, 0])
+        row[0] += 1
+        row[1] += int(usage.get("total") or 0)
+        row[2] += float(cost)
+        if cost:
+            row[3] += 1
+
+    if not stats:
+        click.echo(f"No billable LLM calls in the last {hours}h.")
+        return
+
+    click.echo(
+        f"{'model':38} {'calls':>7} {'tokens':>10} {'cost':>11}  status"
+    )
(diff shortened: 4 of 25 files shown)


src/api/services/langfuse/parse.py
```diff
@@ -436,6 +451,187 @@ def _detect_language(
...
+def _int(value: Any) -> int:
+    try:
+        return int(value or 0)
+    except (TypeError, ValueError):
+        return 0
+
+
+def _obs_cost(obs: dict[str, Any]) -> float:
+    """Cost of one observation, preferring Langfuse's own total."""
+    total = obs.get("totalCost")
```
01painadam, line 523: `_obs_cost` reads `totalCost` / `inputCost` / `outputCost`, but the observations list endpoint returns `calculatedTotalCost`, `calculatedInputCost`, `calculatedOutputCost` and `costDetails` (see `ObservationsView` in langfuse 3.10.1; `totalCost` exists only on traces). Against real API data this returns 0.0 for every generation, so `agent_cost`, `tool_cost` and `cost_by_component` are all zero, and because `total_cost` falls back to the observation sum it becomes 0.0 rather than NULL, which the fill-rate monitor reads as healthy.

The fixture in `test_langfuse_parse.py` uses `totalCost` too, which is why the suite is green. How were the observations for the $0.016605 validation obtained? Suggest reading `calculatedTotalCost` with `costDetails.total` as fallback, plus a fixture captured from a real list response.

src/api/cli.py
```diff
@@ -812,6 +812,85 @@ async def _run():
...
+    )
+
+    # model -> [calls, tokens, cost, priced_calls]
+    stats: dict = {}
+    for o in observations:
+        model = o.get("model")
+        usage = o.get("usageDetails") or {}
+        if not model or not usage:
+            continue
+        cost = o.get("totalCost") or 0
```
01painadam, line 851: Same key problem as `_obs_cost`: observations carry `calculatedTotalCost`, not `totalCost`, so as written this reports every model as UNPRICED.

db/alembic/versions/f3a8c1d5b204_add_llm_usage_split_to_langfuse_traces.py
```diff
@@ -0,0 +1,136 @@
...
+    ("agent_tokens", sa.Integer()),
+    ("tool_tokens", sa.Integer()),
+    ("agent_cost", sa.Float()),
+    ("tool_cost", sa.Float()),
+    ("generation_count", sa.Integer()),
+)
+
+# Recreated (not CREATE OR REPLACE'd) because adding columns to a view's select
+# list changes its shape, which REPLACE rejects.
+_CREATE_VIEW = f"""
```
01painadam, line 49: This view body is the original `8391583bd224` shape. `b1f7a3c9d2e5` (already on main) rewrote the view to read the stored `is_final_turn_in_thread` and expose `turn_index`. As written, upgrade drops `turn_index` from `langfuse_traces_analytics` and reverts the flag to a recomputed window, and `_PRIOR_VIEW` below restores the pre-`b1f7` shape, so downgrade is wrong too.

Please start from that revision's `_VIEW_NEW` and add the five columns. The up/down/up check passes because it only proves the migration runs, not the view's columns.

src/api/services/langfuse/parse.py
```diff
@@ -436,6 +451,187 @@ def _detect_language(
...
+    seen: set[str] = set()
+    cur = by_id.get(obs.get("parentObservationId") or "")
+    while cur is not None:
+        cid = cur.get("id") or ""
+        if cid in seen:  # defensive: a cycle would otherwise hang ingestion
+            break
+        seen.add(cid)
+        if str(cur.get("type") or "").upper() == "TOOL":
+            return cur.get("name") or "unknown_tool"
+        if cur.get("name") == _AGENT_NODE_NAME:
```
01painadam, line 555: This returns `agent` at the first ancestor named `model`, but a nested agent inside a tool has its own `model` node: `search_blogs` runs a deepagents agent built on `langchain.agents.create_agent`, whose LLM node is also `model`. The walk hits that before the enclosing `search_blogs` TOOL span, so blog-search spend is credited to the agent, `tokens_by_component` never gets a `search_blogs` key, and `usage_split_agrees` is False on every such turn (the message-derived figure excludes those calls).

Suggest returning the nearest TOOL ancestor anywhere up the chain, and `agent` only when there is none.

src/agent/subagents/pick_aoi/tool.py
```diff
@@ -348,10 +348,12 @@ async def select_best_aoi(
         ]
     )
 
-    # Chain for selecting the best location match and returning the src_id
+    # Chain for selecting the best location match and returning the src_id.
+    # run_name is what this call is labelled in Langfuse, and so what
```
01painadam, line 352: Small correction: `_component` attributes by the nearest TOOL span, not by chain `run_name`, so this spend lands under `pick_aoi` whatever the name here. The name is still useful in the Langfuse UI; just worth fixing the comment (same in `blog.py`).

src/api/services/langfuse/parse.py
```diff
@@ -436,6 +451,187 @@ def _detect_language(
...
+# --------------------------------------------------------------------------- #
+# The LangGraph model node. A generation whose nearest named ancestor is this
+# chain is a call the agent made itself; anything under a TOOL observation was
+# made by that tool.
+_AGENT_NODE_NAME = "model"
+AGENT_COMPONENT = "agent"
+OTHER_COMPONENT = "other"
+
+
+def _obs_usage(obs: dict[str, Any]) -> tuple[int, int, int, int]:
```
01painadam, line 497: Only `input_cache_read` is added back. The LangChain handler also subtracts every `output_token_details` value from `output` and stores it as `output_reasoning`; `langchain_google_genai` reports thoughts that way and all three models in `llms.py` have thinking on. So `turn_output_tokens` undercounts thinking tokens and `turn_input_tokens + turn_output_tokens != turn_tokens`.

Suggest summing every `input_*` key into input and every `output_*` key into output.

src/shared/langfuse_tracing.py
```diff
@@ -0,0 +1,214 @@
...
+    def _int(value: Any) -> int:
+        try:
+            return int(value or 0)
+        except (TypeError, ValueError):
+            return 0
+
+    cached = _int(getattr(usage, "cached_content_token_count", None))
+    prompt = _int(getattr(usage, "prompt_token_count", None))
+    output = _int(getattr(usage, "candidates_token_count", None)) + _int(
+        getattr(usage, "thoughts_token_count", None)
```
01painadam, line 170: Two things here. (1) Thoughts are folded into `output`, but the handler records LangChain calls with `output` net of reasoning plus a separate `output_reasoning`; emitting `output_reasoning` here keeps one convention across the table. (2) `tool_use_prompt_token_count` is not read. For code execution that is the echoed code and results (the ~920-token bucket in the floor estimate), so `input`, and therefore the cost Langfuse computes, omits it while `total` includes it.

src/api/services/langfuse/ingest.py
```diff
@@ -399,6 +441,82 @@ async def ingest_window(
...
+            # Every type, not just generations: which component a generation
+            # belongs to is written on its ANCESTORS (the LangGraph ``model``
+            # chain, or the TOOL span above it), never on the generation
+            # itself — both an agent call and a tool's call are named
+            # "ChatGoogleGenerativeAI". Fetch generations alone and the parent
+            # walk finds nothing, so every call lands in ``other`` and the
+            # agent/tool split reads as all-tool.
+            None,
+        )
+    except LangfuseFetchError as e:
```
01painadam, line 497: Only `LangfuseFetchError` is caught, but `_get_page` calls `raise_for_status()` for 4xx and `.json()` can raise, and `run_ingestion` has the same narrow catch. A 400 or 403 from the observations endpoint (a key without observation read scope, or an older instance rejecting `toStartTime`) aborts the whole run with the run row left in `running`, instead of degrading to message usage as the docstring promises.

src/api/services/langfuse/parse.py
```diff
@@ -460,24 +656,68 @@ def _detect_language(
...
+        from_messages = combined.get("turn_tokens")
+        combined["agent_tokens_from_messages"] = from_messages
+        combined["usage_source"] = "observations"
+        combined.update(usage)
+        # The agent/tool split hinges on the LangGraph model node still being
+        # named ``model`` (see _component). Rename it upstream and every call
+        # silently becomes "other" — which a fill-rate check cannot catch,
+        # because zero is not null. This flag can: it goes False.
+        combined["usage_split_agrees"] = (
+            usage["agent_tokens"] == from_messages if from_messages else None
```
01painadam, line 712: Falsy check: a legitimate message-derived 0 against non-zero observation agent tokens gives None rather than False, so the drift flag stays silent in one of the cases it is meant to catch. `is not None` would do.

src/api/services/langfuse/ingest.py
```diff
@@ -24,13 +30,21 @@
...
 
 # Langfuse caps list page size (a 300 request 400s); 50 is proven-safe.
 MAX_PAGE_SIZE = 50
 
+# A trace near the end of a window has children that start after the window
+# closes, so the observation fetch reaches past ``to_ts`` by this much. Observed
+# trace latency is tens of seconds; the pad is deliberately far larger, and
+# over-fetching is free because observations for traces outside this window's
+# trace set are discarded.
+OBSERVATION_WINDOW_PAD = timedelta(minutes=30)
```
01painadam, line 46: Operational note rather than a bug: with the default 24-hour chunk this now pulls every observation type for the chunk plus the pad, at page size 50. Traces to observations is roughly 1:30 or more, so a multi-month backfill is an order of magnitude more requests against an instance the fetch module already documents as failing under load, and a chunk that halves down to page size 1 aborts the run. Smaller backfill chunks may be needed.

tests/unit/agent/test_gemini_executor_tracing.py
```diff
@@ -0,0 +1,155 @@
...
+
+    _run(executor._call_model("gemini-3-flash-preview", [{"text": "x"}]))
+    assert attempts["n"] == 3
+    assert len(spans) == 1
+    assert spans[0].recorder.calls == [
+        {"usage": {"input": 9_000, "output": 300, "total": 9_300}}
+    ]
+
+
+def test_call_succeeds_with_langfuse_unconfigured(executor):
```
01painadam, line 143: This opens a real client when `LANGFUSE_*` is present in the environment; running the suite locally with the repo `.env` attempts to export spans to the configured host. Monkeypatching `get_client` keeps it hermetic.

src/agent/tools/update_insight_display.py
```diff
@@ -173,7 +173,9 @@ async def update_insight_display(
         )
 
     current = Insight.from_orm_row(row)
-    revised = await InsightDisplayReviser().revise(current, instruction)
+    revised = await InsightDisplayReviser().revise(
+        current, instruction, config={"run_name": "revise_insight_display"}
```
01painadam, line 177: Nit: `display_reviser.py` already sets this `run_name` via `with_config`, so this is redundant.

src/api/services/langfuse/ingest.py
```diff
@@ -376,17 +402,33 @@ async def ingest_window(
...
+        )
+    obs_by_trace = await _fetch_observations(
+        client,
+        from_ts,
+        to_ts,
+        environment,
+        page_size,
+        {tid for t in traces if (tid := t.get("id"))},
+    )
     stats = WindowStats(fetched=len(traces))
```
01painadam, line 427: Nit: `fetched` now excludes auxiliary traces, so `traces_fetched` in the run row undercounts what was actually fetched. Worth counting before the split or recording the dropped count alongside.

PR #815, opened 2026-09-04
PZB-1231: improve IFL layer responses

guilherme-n:
## Pull request type

Please check the type of change your PR introduces:
- [x] Bugfix

## What is the current behavior?

Issue Number: [PZB-1231](https://gfw.atlassian.net/browse/PZB-1231)

When tree cover loss is filtered by the `intact_forest` context layer, the agent calls the result "intact forest loss" — in chart titles, in the emissions chart, and in the conversation name. That names a different measurement: the reduction in IFL extent, which the IFL dataset publishes separately. Reviewers flagged the chart title "Annual Intact Forest Loss in DRC (2001-2025)" as misleading, since the chart actually plots tree cover loss occurring *inside* intact forest.

## What is the new behavior?

- Titles and prose now read "Tree Cover Loss in Intact Forests", and "Carbon Emissions from Tree Cover Loss in Intact Forests" for the emissions chart, keeping the usual cadence / area / period qualifiers.
- The rule is applied in both instruction paths, since they feed different generators: `presentation_instructions` drives the insight prose, `code_instructions` drives the chart titles. Fixing only the first left the titles wrong in 3 of 4 runs.
- `generate_thread_name` builds its own prompt and never sees the dataset catalog, so the rule is repeated there.
- `docker-compose.dev.yaml`: pin the migrate image to `linux/amd64` so the dev stack builds on Apple Silicon.

Verified against a local backend: three consecutive runs produced correct titles, prose and thread name, including one with a different AOI to confirm the example in the prompt is not being copied verbatim.

`tree_cover_loss_from_fires.yml` carries the wording rule in `presentation_instructions` only; its `code_instructions` was left alone since that dataset does not report emissions, but by the same routing logic its titles may still need it.

## Demo
Screenshot 2026-09-04 at 14 40 44

[PZB-1231]: https://gfw.atlassian.net/browse/PZB-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yellowcap:
after some 🐰 -holing I undid some changes i had pushed here. But there are some gaps 

## The naming rule doesn't reach every generator that writes text

The wording fix lives in the dataset catalog, but the catalog isn't read by
everything that names a metric. Four generators can produce a metric name, and
each reads a different source:

| Generator | Reads | Writes |
| --- | --- | --- |
| Code executor (`build_analysis_prompt`) | `code_instructions` | the original chart titles |
| `InsightTextGenerator` | `presentation_instructions` | the prose |
| `InsightDisplayReviser` | **neither** | rewritten titles *and* prose |
| `generate_thread_name` | **neither** | the conversation name |

Copilot caught one instance by hand: the fires dataset had the rule in
`presentation_instructions` only, so its prose came out right while its chart
titles stayed wrong. Fixed in `c1e47d71`.

**Still open: `InsightDisplayReviser`.** It backs `update_insight_display` — the
"restyle this, don't recompute it" path. It regenerates chart titles and
narrative from scratch, and its prompt gets only the neutral `WORDING_GUIDE`:
no `code_instructions`, no `presentation_instructions`, no dataset at all. So on
an IFL-filtered insight, "shorten the titles" or "reword the summary" hands the
job to a model that was never told the rule, and "Intact Forest Loss" can come
back into a title the original generator got right.

The fix is cheap because the identifier is already there. Each chart persists
its `dataset_id` so a later revision can re-resolve colours against the same
catalog entry (`src/agent/subagents/analyst/charts/model.py:63-66`). The same id
resolves `presentation_instructions` — the reviser just needs to be handed them,
the way it's already handed `WORDING_GUIDE`.

### Two smaller things

**The two instruction fields disagree on plurality.** `code_instructions`
mandates "Tree Cover Loss in **Intact Forests**"; `presentation_instructions`
mandates "intact forest" and gives the same example title spelled singular. The
prose will refer to a chart by a name the chart doesn't have.

**The lead sentence still licenses what's being banned.** Both YAMLs open with
`Use "tree cover loss" rather than "deforestation" only when neither primary nor
intact forest is applied` — so when `intact_forest` *is* active, "deforestation"
is permitted. The rule bans "intact forest loss" but not "deforestation of

yellowcap (approved):
Looks good thanks @guilherme-n !

PR #816, opened 2026-09-07
feat(agent): add an OpenRouter backend for open-model evals

yellowcap:
## What this adds

An optional OpenRouter backend for the model registry, so the agent can run
on open-weight models.

Set `OPENROUTER_MODELS` to a comma-separated list of full OpenRouter model
IDs. Each ID becomes one `ChatOpenAI` in `MODEL_REGISTRY`, keyed by an `or:`
slug built from the part after the `/`:

```
OPENROUTER_MODELS=deepseek/deepseek-v4-pro,z-ai/glm-5.3,qwen/qwen3.8-2.4t-a95b
MODEL=or:deepseek-v4-pro
```

Use those slugs in `MODEL`, `SMALL_MODEL` and `FALLBACK_MODELS`. The list is
empty by default, so nothing changes unless you set it. A non-empty list
without `OPENROUTER_API_KEY` raises at import, instead of a 401 on the first
agent turn.

An entry can also pin providers:

```
OPENROUTER_MODELS=meta-llama/llama-4-maverick@google-vertex+parasail
```

OpenRouter routes one model across several providers, and their capabilities
differ. Two of the five providers that serve `llama-4-maverick` do not
support tool calling, so a default-routed request can land somewhere that
cannot run the agent. A pinned entry keeps its first provider in the slug
(`or:llama-4-maverick@google-vertex`), so the pinned and the default-routed
variant can both be registered and compared.

No new dependency: `langchain-openai` was already pinned.

## The models in `.env.example`

The example list holds the three best models from a 24-case GOLD pilot over
ten open models:

| model | rows passed | p50 | max | run cost |
|---|---|---|---|---|
| `or:deepseek-v4-pro` | 24/24 | 77s | 146s | $0.29 |
| `or:glm-5.3` | 24/24 | 71s | 367s | $0.68 |
| `or:qwen3.8-2.4t-a95b` | 23/24 | 65s | 88s | $0.82 |
| `gemini` (yardstick) | 22/24 | 47s | 71s | — |

`glm-5.3` matched `deepseek-v4-pro` on accuracy but carried a 367s worst
case, which is the difference between a usable and an unusable interactive
answer. Latency figures include the shared Gemini code-execution step.

## What the sweep does not measure

`generate_insights` cannot move off Gemini. The analyst subagent sends
dataframes to Gemini and lets Gemini run Python in its own server-side
sandbox, and `CODING_MODEL` bypasses the registry. So a sweep holds
`CODING_MODEL` and `SMALL_MODEL` on Gemini and varies only `MODEL`. It
measures orchestration. The analysis and output buckets are partly Gemini's
score.

## Also in this PR

- `gemini-flash-3.8` registered as a pinned version, not the
  `gemini-flash-latest` alias. An alias silently changes what an eval run
  measured.
- `docs/openrouter-open-model-evals-plan.md`: the sweep plan, the model
  slate, the two-stage run procedure and the pilot findings.

## Test plan

- `tests/unit/test_openrouter_registry.py` (15 tests): slug building,
  provider pinning, the empty default, the missing-key error, and that
  registry entries carry the right model ID and `extra_body`.
- Ten models ran the 24-case GOLD pilot through this backend against a local
  API.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,11 +15,27 @@ COOKIE_SIGNER_SECRET_KEY="secret-cookie-key-123"
 DATASET_EMBEDDINGS_DB=<name-of-dataset-embeddings-db>
 # Model Configuration
 # Available options: sonnet, haiku, gemini, gemini-flash, gpt, phi4, smollm2
+# Plus one "or:<name>" slug per model listed in OPENROUTER_MODELS below.
 MODEL=gemini
 SMALL_MODEL=gemini-flash
 FALLBACK_MODELS=gemini-flash,gemini-flash-lite
 CODING_MODEL=gemini-3.1-pro-preview # Full Google model IDs required (used by genai SDK directly, not LangChain)
 CODING_FALLBACK_MODELS=gemini-2.5-pro,gemini-3-flash-preview # Full Google model IDs required
+
+# OpenRouter backend (optional). List full OpenRouter model IDs; each is
+# registered under an "or:" slug built from the part after the "/", so
+# "deepseek/deepseek-v4-pro" becomes "or:deepseek-v4-pro". Use those slugs in
+# MODEL, SMALL_MODEL and FALLBACK_MODELS. A non-empty list needs the API key.
+# Note: CODING_MODEL never uses this backend. generate_insights runs Python in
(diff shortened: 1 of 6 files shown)


yellowcap:
we can revisit later

PR #818, opened 2026-09-08
fix(pick_dataset): skip index documents with no catalog dataset PZB-1341 

### LINKED ISSUE #817: PZB-1341 pick_dataset fails with "No data found for dataset ID: 0": published RAG indexes still contain the removed DIST dataset
## Summary

`pick_dataset` fails with `No data found for dataset ID: 0` when the RAG retriever returns the deleted DIST-ALERT dataset. Every dataset embeddings index in S3 (v1 through v9) still contains that document. The catalog does not.

Observed in this trace, 2026-09-08 08:45:58 UTC:
https://langfuse.staging.globalnaturewatch.org/project/wri_lcl/traces/e0c8797d8d063ec57b62c3b496bcb403?timestamp=2026-09-08T08%3A45%3A58.677Z&display=details

Latency 0.21 s:

```
pick_dataset
Input:  {'query': 'daily integrated deforestation alerts for August 2026',
         'start_date': '2026-08-01', 'end_date': '2026-08-31'}
Output: null
Status: No data found for dataset ID: 0
```

## Cause

Doc id `0` is DIST-ALERT (`Global all ecosystem disturbance alerts`). PR #802 (commit 4e490ba4, 2026-08-27) removed that dataset from the catalog, so `DATASETS` now holds ids 1-12.

`rag_candidate_datasets` maps each retrieved doc id directly to a `dataset_id` and raises if no dataset matches:

https://github.com/wri/project-zeno/blob/main/src/agent/subagents/pick_dataset/tool.py#L90-L96

The index is a build artifact of the catalog, but no pipeline rebuilds or publishes it when the catalog changes. Only `tools-tests.yml` regenerates it, into a throwaway CI `data/` directory.

Doc ids in the published indexes (`s3://zeno-static-data/`):

| index | built | docs | contains id 0 |
| --- | --- | --- | --- |
| v3 (repo default, `src/shared/config.py:34`) | 2026-03-25 | 10 (0-9) | yes |
| v7 (production, `values-production.yaml` on `main`) | 2026-07-06 | 12 (0-11) | yes |
| v8 | 2026-08-07 | 13 (0-12) | yes |
| v9 | 2026-08-19 | 13 (0-12) | yes |

v9 is the newest index and predates the DIST removal by 8 days. A bump from v7 to v9 therefore does not correct this fault. It only adds LGMS (id 12).

## Effect

Queries with daily or alert wording are nearest to the DIST-ALERT document, so the retriever puts doc id 0 in the top 5 for `k=5`. The failure is deterministic for that class of query, not intermittent. Four valid candidates are discarded with it, because one unknown doc id stops the whole tool.

The affected deployment is staging, as the Langfuse host in the trace link above shows. `develop` runs image `e6f8d726` (2026-09-03, no DIST in the catalog) against index v9 (DIST present). Production on `main` pins image `0c7ed1a5` (2026-08-04), which still contains `global_all_ecosystem_disturbance_alerts_dist_alert.yml`, together with v7. That pair is consistent and does not fail.

The trace itself reports `Env: production`, which does not identify the environment. `src/api/services/chat.py:23` reads `STAGE`, but the Helm values set `GNW_STAGE` only, so all deployments label their traces `production`.

Production is exposed as soon as the pending deploy change merges. It removes the production image pin, which moves production to image `e6f8d726` on index v9 — the same mismatch that staging has now.

## Actions

1. Build a v10 index from the current catalog, publish it, and point both Helm values files at it:

   ```
   DATASET_EMBEDDINGS_DB=gnw-dataset-index-gemini-v10 uv run python src/ingest/embed_datasets.py
   aws s3 cp data/gnw-dataset-index-gemini-v10 s3://zeno-static-data/
   ```

2. Make `rag_candidate_datasets` skip an unknown doc id and write a warning, in place of raising. Return the standard no-dataset result if no candidate remains. A stale index must degrade, not stop the turn.

3. Add a check that compares the index doc ids with the catalog dataset ids, and fails the build or the deploy on a mismatch.

4. Never overwrite a published index name. The local `data/gnw-dataset-index-gemini-v3` file was rebuilt on 2026-08-31 with 12 docs, but the S3 object of the same name is the 10-doc March file. The repo default points at that name, so a pod that uses the default gets an index without integrated alerts and without LGMS. Increment the version for each rebuild.

5. Set `STAGE` in the Helm values, or read `GNW_STAGE` in `chat.py`, so trace environment labels are correct.

yellowcap:
Fixes #817.

## Problem

`pick_dataset` failed with `No data found for dataset ID: 0` on staging ([trace](https://langfuse.staging.globalnaturewatch.org/project/wri_lcl/traces/e0c8797d8d063ec57b62c3b496bcb403?timestamp=2026-09-08T08%3A45%3A58.677Z&display=details)):

```
Input:  {'query': 'daily integrated deforestation alerts for August 2026', ...}
Output: null
Status: No data found for dataset ID: 0
```

Doc id 0 is DIST-ALERT, removed from the catalog in #802. The embeddings index is a build artifact of `src/agent/datasets/catalog/`, but it is published separately as an S3 object that the pods sync into `data/`. No pipeline rebuilds it when the catalog changes, so every published index up to v9 still holds that document. Queries with daily or alert wording are nearest to it, which made the failure deterministic for that class of query.

`rag_candidate_datasets` raised on the first unusable doc id, so the four valid candidates retrieved alongside it were lost too.

## Changes

**`pick_dataset` tolerates a stale index**
- `rag_candidate_datasets` skips a document whose id has no catalog dataset and keeps the valid candidates. It logs a warning with the skipped ids, the index name and the catalog ids. A non-numeric id is skipped in place of raising on `int(doc.id)`.
- `_log_index_catalog_skew` runs once when the index loads and reports both directions of skew: indexed but not in the catalog (this bug), and in the catalog but not indexed (v7 could never retrieve LGMS). A matching index logs one confirmation line.
- `DatasetSelector.resolve` returns a plain message if no candidate survives retrieval or the profile exclusions, in place of the `KeyError` that an empty DataFrame produced in `_drop_excluded_datasets`. New i18n key `pick_dataset.retrieval_unavailable`.

**The index build guards itself** — `embed_datasets.py` verifies the dumped doc ids against the catalog and exits non-zero on a mismatch. Its docstring states the publish procedure, and the rule against overwriting a published version name.

**New default index version** — `DATASET_EMBEDDINGS_DB` now defaults to `gnw-dataset-index-gemini-v10`, built from the current catalog (12 datasets, ids 1-12). The old default named `v3`, whose S3 object is a 10-document file from March that holds neither integrated alerts nor LGMS. The local `data/gnw-dataset-index-gemini-v3` file had been rebuilt in place under the same name, which hid the difference.

**Trace environment labels** — `LANGFUSE_TRACING_ENVIRONMENT` falls back to `GNW_STAGE`, the variable the Helm values actually set. `STAGE` is set nowhere, so every environment labelled its traces `production`, which is why the failing pod first looked like production.

## Deployment

This PR alone does not restore retrieval on staging. Two steps go with it:

1. Publish the index: `aws s3 cp data/gnw-dataset-index-gemini-v10 s3://zeno-static-data/`
2. Point `DATASET_EMBEDDINGS_DB` at `v10` in `values-staging.yaml` and `values-production.yaml` (project-zeno-deploy).

Publish before the Helm change, or the pods sync a name that does not exist.

Production is on image `0c7ed1a5` (2026-08-04), whose catalog still contains DIST-ALERT, paired with index v7. That combination is consistent and does not fail. Moving production to v10 needs an image that postdates #802, which the pending deploy change does by removing the production image pin.

## Tests

`tests/unit/agent/tools/test_pick_dataset_index_skew.py`: a stale id is skipped and valid candidates kept, a non-numeric id is skipped, an empty shortlist returns a message rather than raising, and both skew-logging branches. `tests/unit` passes (636).


yellowcap:
Copied v10 to s3 ✅ 

Deployed v10 to staging ✅ 

Confirmed problem solved in staging ✅

gtempus:
Sharing my two cents from an architectural and systems-thinking level, along with protecting the value stream...
**Don't let this hold up your review/fix** I'll create a technical debt Jira ticket. It's been on my mind for a while.

Two things I'd love us to internalize as a team, before the specifics:

1. **Defensive code for a lookup that "shouldn't" fail is a signal, not a solution.** Skip-if-missing, log-if-mismatch, catch-and-degrade all mean an invalid state got reachable. Fixing the lookup handles today's occurrence; asking *why the state was reachable* usually finds the bigger fix.
2. **Two things that must always agree but ship through different paths will eventually disagree — that's not bad luck, that's the design.** The fix is to remove the second path (one release unit, one source of truth), not to make disagreement survivable. We've paid for this once already on the frontend — `project-zeno-next` carried its own hardcoded dataset registry for a year, drifting from the backend, patched by a long run of hotfixes — until it was deleted in favor of fetching the backend's copy directly.

## Root cause

The dataset catalog (deployed with every image) and the RAG index (a hand-built S3 object, manually versioned, manually re-pointed per environment via Helm) are two artifacts with no shared trigger to keep them in sync. This PR makes that safe to tolerate; it doesn't remove the coupling, so it can recur with the next catalog change.

## Soft deletes vs. hard deletes

Worth a second look before merge: #802 deleted DIST-ALERT's catalog entry outright rather than retiring it. That's a hard delete — the id simply stops existing, and nothing stops a future dataset from being assigned `0` again. A soft delete (a `status: retired` field, entry kept in the repo, id permanently reserved) costs one field and never frees the key. The difference matters a lot more once catalog and index can drift, which they can here: under a hard delete, a stale index entry for a removed id has nothing to resolve to and fails loud (today's crash) or gets silently skipped (this PR's fix). Under a soft delete, it resolves to a known, retired dataset every time — no special-casing needed. And the sharper risk is the reuse case: if id `0` is ever reassigned to a *new* dataset while some environment still runs a stale index, that index's old DIST-ALERT embedding now silently answers under the new dataset's identity — wrong data returned, not an error raised, and invisible to the skew check since it only diffs id sets, not content. Hard delete makes that reachable; soft delete makes it impossible by construction. Treat dataset ids as permanently reserved once assigned, never recycled.

Soft delete also simplifies this PR's diff, not just its safety. Once a retrieved doc id can never fail to resolve, `rag_candidate_datasets`'s skip loop, the `stale_doc_ids` warning, `_document_dataset_id`'s `None` case, the empty-DataFrame branch in `DatasetSelector.resolve`, and `_log_index_catalog_skew` all become unreachable — along with most of `test_pick_dataset_index_skew.py`, which exists specifically to cover those branches. That's the tell: code written to tolerate a bad state is a cost that soft delete removes entirely, whereas hard delete imposes it forever.

## Proposal 
I'm not completely married to this implementation. We can discuss. My main goal is to have folks start thinking about the first principles outlined above. Both when implementing solutions and resolving issues over time. If we can get better about asking these questions, we'll almost guarantee continuous improvement of the system.

I sketched an approach that removes the coupling instead of tolerating it — versioning the index by a hash of the catalog, built in CI, checked at pod readiness:

```mermaid
sequenceDiagram
    actor Dev
    participant Repo as catalog/*.yml
    participant CI as CI: build-index job
    participant Embed as embed_datasets.py
    participant API as Embeddings API
    participant S3 as S3 (content-addressed)
    participant Build as Image build
    participant Pod as App pod (boot)
    participant Tool as pick_dataset tool

    Dev->>Repo: edit/add/retire a dataset YAML
    Dev->>CI: open PR (path filter: catalog/**)
    CI->>Repo: read DATASETS
    CI->>CI: hash = sha256(sorted catalog contents)
    CI->>S3: HEAD gnw-dataset-index-<hash>
    alt index already published for this hash
        S3-->>CI: exists — nothing to build
    else no index for this hash
        CI->>Embed: run with --hash <hash>
        Embed->>API: embed each dataset doc
        API-->>Embed: vectors
        Embed->>Embed: assert dumped ids == catalog ids (exit 1 on mismatch)
        Embed->>S3: publish gnw-dataset-index-<hash> (new name, never overwritten)
    end
    CI-->>Dev: required check "index matches catalog" ✅ — merge unblocked
    Dev->>CI: merge to main
    CI->>Build: build app image, embed catalog hash as an image label
    Build->>Pod: deploy (no Helm value to hand-edit — hash travels inside the image)
    Pod->>Pod: read own image label -> expected hash
    Pod->>S3: fetch gnw-dataset-index-<expected hash>
    S3-->>Pod: index bytes
    Pod->>Pod: readiness probe: index hash == image's catalog hash?
    alt mismatch
        Pod-->>Pod: fail readiness — pod never serves traffic
    else match
        Note over Pod: catalog and index are provably the<br/>same artifact — divergence can't reach a serving pod
    end
    Tool->>Pod: rag_candidate_datasets(query)
    Pod-->>Tool: candidates (every retrieved id resolves — same hash guarantees the id sets are identical)
```

yellowcap:
Thanks @gtempus — the first-principles framing is right, and the id-reuse case is the sharpest thing in it. I've pushed a fix for that part in 19cfe5b3. Below is what I took, what I left, and why.

## Taken: ids are permanently reserved

You're right that nothing stopped a future dataset from taking `0` again, and right that this is the failure that returns wrong data instead of an error. It is also already more concrete than the comment assumes: `db/alembic/versions/a1b2c3d4e5f6_...py:63` backfills `statistics.dataset_id = 0` for DIST-ALERT rows, and `GET /api/insights?dataset_id=` filters on that column (`src/api/routers/insights.py:130`). So id 0 is already reserved by persisted data and by a public API contract. Reuse would silently re-attribute historical insights, on top of the stale-index problem.

`RETIRED_DATASET_IDS` in `src/agent/datasets/config.py` now names the reserved ids, and `_load_datasets` refuses a catalog that reuses one. Reuse is impossible by construction, and the guard fires at import, so no environment can boot on such a catalog.

`_log_index_catalog_skew` also now reports a retired indexed id apart from one the catalog has never held. The first says the index predates a removal; the second has no such explanation.

That is the whole reuse hazard closed, in ~20 lines, with no new filter points.

## Left: `status: retired` on the catalog entry

I traced what the soft delete would cost here before deciding. `DATASETS` is read at nine call sites that all currently mean "everything we offer", and a retired entry has to be filtered at six of them:

- `src/ingest/embed_datasets.py` — skip it at embed time, or it stays discoverable
- `embed_datasets.py`'s new build assert — catalog ids ≠ index ids **by design** once retired entries are deliberately unembedded, so the assert needs a new "indexable ids" concept or it fails every build
- `_log_index_catalog_skew` — same, or it logs an error on every boot
- `pick_dataset/schema.py:50` — the `dataset_id` validator would accept `0` as valid model output
- `skills/capabilities.py:12` — the agent would list DIST-ALERT to users as available
- `datasets/palette.py:54` and `datasets/layers.py:90` → `/api/datasets/catalog` — would publish DIST's `series_color` and tile URL to the frontend

Two of the filters are free: `DEFAULT_EXCLUDED_DATASETS` (`agent_config.py:133`) already does exactly this for LGMS. The other four are new, and each is a place a future catalog change can forget the flag. If one is missed on the selection path, `pull_data` reaches a handler branch #802 deleted — a worse failure than the raise this PR replaces.

Two corrections to the "tolerance code becomes unreachable" claim:

1. **The empty-shortlist branch is not about delete semantics.** I reproduced it: `KeyError 'dataset_name'` comes from `_drop_excluded_datasets` indexing a column-less DataFrame. It's a pre-existing bug on the LGMS exclusion path and stays reachable under any delete policy.
2. **Not embedding a retired dataset does not fix the reported bug.** A stale index still holds DIST-ALERT's vector; skipping it at embed time only takes effect from the next rebuild. What stops the wrong pick is a shortlist filter — the same shrink-the-shortlist behavior this PR already produces. And `_log_index_catalog_skew`'s second direction (in the catalog, never indexed — the v7/LGMS fault) is untouched by soft delete.

So the skip loop shrinks to a filter and the rest stays. I'd rather not pay six filter points and a permanent "ids don't match by design" caveat in two guards for that.

## Agreed: the real fix is one release unit

Your (2) is the root cause and I want it. One fact that makes it cheaper than the sketch: the index is **758 KB, 208 KB gzipped** — 12 documents of 3072-dim vectors. And this repo already bakes an S3 build artifact into the image: `docker-build.yml:47` passes `WRI_INSIGHTS_S3_URI`, `scripts/wri_insights_snapshot.py` unpacks it during the build, while the pods' init container does `aws s3 sync s3://bucket /app/data` for runtime data.

Moving the dataset index from the sync path to the build path removes the hand-edited Helm value and makes drift unreachable — no catalog hash, no readiness probe, no content-addressed naming. The image carries the catalog and its index as one artifact, which is the "one release unit, one source of truth" outcome your diagram is after, with much less machinery.

Happy for that to be the technical-debt ticket, and I'll take it. Keeping it out of this PR since the fix is deployed and verified on staging.

gtempus:
Thanks for taking a look and adopting some suggestions, @yellowcap!
Here is the [Jira ticket](https://gfw.atlassian.net/browse/PZB-1342) for the RAG deployment redesign.
Here is the [Jira ticket](https://gfw.atlassian.net/browse/PZB-1350) for establishing the Catalog boundary.

Regarding the soft delete. I don't expect any changes here, but it does make for a good follow-on discussion about tech debt, too. I asked claude to summarize for me so I could share more in-depth and not be so hand-wavy.

# `DATASETS` has no boundary — a design smell PR #818 exposed, not caused

Context: [PR #818](https://github.com/wri/project-zeno/pull/818) fixes a
crash in `pick_dataset` caused by a stale RAG index still holding a document
for a dataset removed from the catalog in #802. During review, a soft-delete
alternative (`status: retired` on the catalog entry, instead of deleting it
outright) was proposed and then declined — not because retirement is a bad
idea, but because retrofitting it onto today's code would mean touching six
independent read sites. That cost estimate is correct, and it is the
symptom this document is about. The underlying problem predates #818 and is
not specific to retirement — it should be tracked as its own tech-debt item,
separate from and not blocking #818.

Related, orthogonal proposal: [`catalog-index-coupling-proposal.md`](./catalog-index-coupling-proposal.md)
addresses *why the catalog and the index can disagree at all*. This document
addresses a different question: *why does agreeing on one rule
("is this dataset visible?") require six places to agree with each other?*
The two can be worked independently, in either order.

## The smell

`DATASETS` (`src/agent/datasets/config.py:49`) is a plain `list[dict]`
loaded once from YAML. It has no repository, no aggregate, no query method —
just the raw rows. Nothing stands between "the full contents of the catalog
files" and "the set of datasets a given consumer should actually see."

As a result, every consumer that needs a *subset* of the catalog reaches
into the raw list and writes its own filter. Nine-plus call sites import
`DATASETS` directly:

- `src/ingest/embed_datasets.py`
- `src/agent/subagents/pick_dataset/tool.py`
- `src/agent/subagents/pick_dataset/schema.py`
- `src/agent/datasets/palette.py`
- `src/agent/datasets/dates.py`
- `src/agent/datasets/handlers/analytics_handler.py`
- `src/agent/agent_config.py`
- `src/agent/skills/capabilities.py`

None of them talk to a shared abstraction. Each one is, in effect, also
playing the role of "the repository" — deciding for itself, in its own
words, what counts as offerable.

## This isn't hypothetical — it's already inconsistent today

The catalog already has one cross-cutting visibility rule:
`DEFAULT_EXCLUDED_DATASETS` (`agent_config.py:133`) hides LGMS from normal
traffic. Three independent implementations of "is this excluded" exist for
it, and they don't agree:

| Site | Applies the exclusion? |
|---|---|
| `pick_dataset/tool.py` — `_drop_excluded_datasets` | Yes — filters the RAG candidate `DataFrame` |
| `skills/capabilities.py` — `load_datasets_info` | Yes — a second, independently hand-written loop |
| `datasets/palette.py` — `_build_palettes` | **No** — no `excluded` check anywhere in it |

That last row is a live leak, not a theoretical risk: LGMS's `series_color`
is published today via `/api/datasets/catalog` even though the same
dataset is hidden from selection and from the capabilities listing. Nobody
decided this leak should exist — it's what happens by default when an
invariant is enforced by convention at N call sites instead of owned by one
model.

## How #818 draws it out of hiding

The soft-delete proposal asked for one more visibility rule: "retired
datasets are never offered." Tracing what that would actually cost showed
it has to be re-applied at (at least) six sites — the embed step, the build
assertion, the skew logger, the `dataset_id` validator, the capabilities
listing, and the palette/`\`/api/datasets/catalog\`` path — because none of
them share a definition of "offerable" to extend.

That cost is real, and it is a legitimate reason to keep `status: retired`
out of #818. But the cost isn't a property of retirement — it's a property
of the architecture retirement was asked to fit into. The same six-way
fan-out was already latent for `DEFAULT_EXCLUDED_DATASETS`; retirement just
made it expensive enough to notice, and the LGMS/palette leak above shows
the fan-out was already failing silently before retirement was ever
proposed.

## What "pulling it up" means here

In DDD/hexagonal terms: `DATASETS` today is an adapter's raw output (parsed
YAML) handed directly to callers, with no port in between. The fix is not a
new field on the dict — it's a small `Catalog` (or `DatasetRepository`)
object that owns the *behavior* "which datasets are visible," exposed
through a couple of methods:

- `Catalog.all()` — every dataset ever assigned an id, retired included.
  Needed by `_load_datasets`'s id-reuse guard, the embed-index skew check,
  and anything resolving a *historical* reference (e.g. `statistics.dataset_id`,
  `GET /api/insights?dataset_id=`) that must keep working for a retired id.
- `Catalog.offerable(excluded: frozenset[str] = frozenset())` — everything
  minus retired datasets minus the given profile's exclusions. This one
  method is what `pick_dataset`, `capabilities.py`, and the
  `/api/datasets/catalog` palette path should all call through.
- `Catalog.get(dataset_id)` — a single lookup, returning `None` for an
  unknown *or* retired id, replacing the ad hoc
  `[ds for ds in DATASETS if ds["dataset_id"] == x]` written out at four
  different sites today.

Once "offerable" is a method instead of a convention, profile-exclusion and
retirement become the *same* mechanism — one predicate on `Dataset`,
evaluated once — instead of two independently invented, inconsistently
applied filters. Adding retirement then costs one clause in
`Catalog.offerable()`, not six call-site edits, and the palette leak closes
as a side effect rather than needing its own fix.

## Before / After

**Before** — every consumer reads the raw list and re-derives its own
notion of "visible," so implementations diverge (and one omits the check
entirely):

```mermaid
sequenceDiagram
    participant YAML as catalog/*.yml
    participant Config as config.py: DATASETS (raw list)
    participant Tool as pick_dataset/tool.py
    participant Schema as pick_dataset/schema.py
    participant Caps as skills/capabilities.py
    participant Palette as datasets/palette.py
    participant API as /api/datasets/catalog

    YAML->>Config: parsed once at import
    Note over Config: plain list[dict] — no query surface,<br/>no owner of "is this visible"

    Tool->>Config: import DATASETS
    Tool->>Tool: _drop_excluded_datasets(df, excluded)<br/>own exclusion filter + own retired-id skip loop

    Schema->>Config: import DATASETS
    Schema->>Schema: `dataset_id in [ds["dataset_id"] for ds in DATASETS]`<br/>own inline lookup, no retired-aware guard

    Caps->>Config: import DATASETS
    Caps->>Caps: for ds in DATASETS: skip if name in excluded<br/>reimplements Tool's rule, independently, by hand

    Palette->>Config: import DATASETS
    Palette->>Palette: for ds in DATASETS: build palette entry
    Note over Palette: no excluded check at all
    Palette->>API: publish series_color for EVERY dataset
    Note over API: LGMS leaks through here today,<br/>despite being "hidden" everywhere else
```

**After** — a `Catalog` owns the predicate once; every consumer calls
through it and gets the same answer:

```mermaid
sequenceDiagram
    participant YAML as catalog/*.yml
    participant Catalog as Catalog (repository)
    participant Tool as pick_dataset/tool.py
    participant Schema as pick_dataset/schema.py
    participant Caps as skills/capabilities.py
    participant Palette as datasets/palette.py
    participant API as /api/datasets/catalog

    YAML->>Catalog: parsed once at import
    Note over Catalog: owns Dataset entities +<br/>offerable() = not retired, not profile-excluded

    Tool->>Catalog: offerable(profile.excluded_datasets)
    Catalog-->>Tool: candidates — retired & excluded already removed

    Schema->>Catalog: get(dataset_id)
    Catalog-->>Schema: Dataset, or None if unknown/retired

    Caps->>Catalog: offerable(profile.excluded_datasets)
    Catalog-->>Caps: same predicate, same result as Tool — by construction

    Palette->>Catalog: offerable(profile.excluded_datasets)
    Catalog-->>Palette: same predicate — LGMS and any retired id excluded automatically
    Palette->>API: publish series_color only for offerable datasets
    Note over Catalog: retirement and profile-exclusion<br/>are one rule, defined once
```

## Scope

This is a refactor of the read boundary around `DATASETS`, touching every
call site listed above — real work, and not something to fold into a bug
fix. #818 should ship as-is (hard delete, id permanently reserved via
`RETIRED_DATASET_IDS`, tolerant retrieval). This document is the case for a
follow-up tech-debt ticket: introduce a `Catalog`/`DatasetRepository`
boundary with an `offerable()` query, migrate the nine current call sites
onto it, and fix the palette exclusion gap in the same pass. Retirement
(`status: retired`) becomes a one-line addition to `offerable()` once that
boundary exists, rather than a six-site change today.

## Further reading

Naming the smell:
- Martin Fowler, "Shotgun Surgery"— the *Refactoring* smell for exactly what happened here: one conceptual change (retirement) forced small edits in six unrelated places, because no one place owned the rule. (Fowler & Beck, *Refactoring: Improving the Design of Existing Code*, 2nd ed., Addison-Wesley, 2018.)
- Martin Fowler, "Primitive Obsession" — same book; the root cause of the shotgun. `DATASETS` is `list[dict]`, so there is no type to hang an `offerable()`/`is_retired` method on, and every consumer re-derives the rule against raw keys instead of calling a method.
- Martin Fowler, ["AnemicDomainModel"](https://martinfowler.com/bliki/AnemicDomainModel.html) (bliki, 2003) — the general anti-pattern: objects that are just data, with all behavior (the visibility rule, in this case) living in the services/call sites that consume them instead of on the model.

The pattern that resolves it:
- Eric Evans, *Domain-Driven Design: Tackling Complexity in the Heart of Software* (Addison-Wesley, 2003) — the source for Entity, Aggregate, and Repository; the case for putting invariants ("is this dataset offerable") on the model rather than scattering them across clients.
- Vaughn Vernon, *Implementing Domain-Driven Design* (Addison-Wesley, 2013) — more operational than Evans; his chapters on Aggregates and Repositories are the closest match to sizing something as small as the proposed `Catalog`.
- Martin Fowler et al., *Patterns of Enterprise Application Architecture* (Addison-Wesley, 2002), the Repository pattern — "mediates between the domain and data mapping layers... acting like an in-memory domain object collection." `Catalog.offerable()`/`.get()` is a direct instance of this.

Broader framing this document leans on:
- Alistair Cockburn, ["Hexagonal Architecture"](https://alistair.cockburn.us/hexagonal-architecture/) (2005) — the ports-and-adapters vocabulary used above: `DATASETS` today is an adapter's raw output hitting callers with no port in front of it.
- Sandi Metz, *Practical Object-Oriented Design* (Addison-Wesley, 2018) — an OOD (rather than DDD) route to the same conclusion: an object's job is to know one thing well and answer questions about it, which is what none of the nine call sites can currently do without duplicating logic.

Adjacent, worth a skim:
- Alexis King, ["Parse, Don't Validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) (2019) — reframes `schema.py`'s `dataset_id` validator and the four ad hoc `[ds for ds in DATASETS if ...]` lookups: they re-check the same fact repeatedly instead of parsing once into a `Dataset` (or `None`) that the rest of the code can trust.

gtempus (approved):
See discussion thread. My comments are at the design/architecural level.
Thanks, @yellowcap!

yellowcap:
@gtempus thanks for pointing this out. I agree that his is not an ideal solution. I think our yaml approach is hitting harder and harder limits. If we have time for refactors, that would be a good place to start, as it is a source of a lot of if/else code. Lets see if and when we want to make time for this kind of work.

PR #819, opened 2026-09-08
fix(pick_dataset): report the date range applied to the tile URL

yellowcap:
## Where the problem comes from

`pick_dataset` builds the tile URL for date-driven layers and appends the date filter itself
(`src/agent/subagents/pick_dataset/tool.py`, the `INTEGRATED_ALERTS_ID` branch of
`get_tile_services_for_dataset`):

```python
elif selection_result.dataset_id == INTEGRATED_ALERTS_ID:
    tile_url += f"&start_date={start_date}&end_date={end_date}"
```

The GFW tile service honours those params — verified against
`gfw_integrated_dist_alerts` at z9/167/268 (Rondônia), which returns a different
raster per window:

| query | bytes |
|---|---|
| none | 17766 |
| `start_date=2024-01-01&end_date=2024-01-31` | 15483 |
| `start_date=2023-12-01&end_date=2026-09-08` | 20341 |

So the filter works. What fails is the range fed into it.

`start_date` / `end_date` are **optional** tool args on `pick_dataset`. When the model
omits them, `revise_date_range` fills them from the dataset's own coverage:

```python
if start_date is None:
    start_date = available_start   # 2023-12-01 for Integrated alerts
if end_date is None:
    end_date = available_end       # date.today(), the YAML has no end_date
```

The tile URL then covers the full ~21 months. `pull_data` takes its own
`start_date`/`end_date` and re-resolves them independently, so the chart shows the
window the user asked for while the map shows everything — the reported symptom
("selected the last two weeks, it showed the last year").

Nothing made that visible:

- `pick_dataset` discarded the clamp flag `revise_date_range` returns, into `_`.
- The flag would not have fired anyway: `range_clamped` is computed **after** the
  `None` fill-in, so omitted args are indistinguishable from a request for exactly
  the full coverage.
- The tool message never named the applied range. Its only date field is
  `content_date`, the dataset's static YAML coverage string
  (`"1 December 2023 - present"`), not the window that was applied.
- The session block (`src/agent/middleware.py`) renders `Date range:` from the
  top-level `state["start_date"]`/`["end_date"]`, which only `pull_data` writes.
  `pick_dataset` puts its dates under `state["dataset"]`.

The agent therefore had no way to know which window it had put on the map, and
neither did the user.

## What this changes

Adds an `## Applied date range` section to the `pick_dataset` tool message, followed
by a note when the applied range is not what was asked for:

- **nothing requested** — says so, states that the map layer is scoped to the full
  coverage, and tells the model to call `pick_dataset` again with dates if the user
  wanted a shorter period
- **one bound missing** — names which one was defaulted
- **clamped** — quotes the requested range and says to tell the user about the
  adjustment (matching the wording `pull_data` already uses)
- **applied as asked** — no note, the range line already says it

`revise_date_range` keeps its signature. Whether a bound was defaulted is read from
the incoming tool args, which `resolve()` already has, so this needs no change to
`dates.py` and no churn in its 17 existing test call sites.

## Scope

Reporting only — this makes the mismatch visible to the model and in traces. It does
not change which dates get applied. Still open, deliberately out of scope:

- `pick_dataset` does not write the top-level `state["start_date"]`/`["end_date"]`, so
  the session block still shows the previous `pull_data` range while the model decides
  the next call.
- The `analyze` / `pull-data` skills mention computing dates but never say to pass them
  to `pick_dataset`, and the tool docstring calls them "optional … narrow the range",
  which reads as skippable.
- The frontend legend truncates any range to years (`buildYearParam`), so a two-week
  filter renders as `YEAR 2026` regardless. Separate PR in `project-zeno-next`.

## Tests

`tests/unit/agent/tools/test_pick_dataset_date_note.py` — five cases on the note
itself, plus two that drive `pick_dataset.ainvoke` with the selector mocked out and
assert the message and the tile URL agree. One of those reproduces the reported bug:
no dates in, full coverage on the tile URL, note now present.

```
7 passed in 0.10s
```

Part of #821.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.9.2.1"
+version = "2026.9.8"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/subagents/pick_dataset/tool.py
+++ b/src/agent/subagents/pick_dataset/tool.py
@@ -206,6 +206,42 @@ async def select_best_dataset(
     )
 
 
+def _applied_date_range_note(
+    requested_start: Optional[str],
+    requested_end: Optional[str],
+    effective_start: str,
+    effective_end: str,
+    range_clamped: bool,
+) -> str:
+    """Explain how the applied date range relates to the requested one.
+
+    The tile URL is scoped to the applied range (see
+    get_tile_services_for_dataset), so a range the model did not intend
+    silently changes what the map shows. Returns "" when the applied range
+    is exactly what was asked for and needs no explanation.
+    """
+    if requested_start is None and requested_end is None:
+        return (
+            "No date range was requested, so this is the dataset's full "
+            "coverage. The map layer is scoped to this range. If the user "
+            "asked for a shorter period, call pick_dataset again with "
+            "start_date and end_date."
+        )
+    if requested_start is None or requested_end is None:
+        bound = "start" if requested_start is None else "end"
+        return (
+            f"No {bound} date was requested, so the dataset's own {bound} of "
+            "coverage was used. The map layer is scoped to the range above."
+        )
+    if range_clamped:
+        return (
+            f"The requested range ({requested_start} to {requested_end}) "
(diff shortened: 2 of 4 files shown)


yellowcap:
can revisit later

PR #820, opened 2026-09-08
feat(datasets): declare tile date conventions and serve them from the catalog

yellowcap:
## Where the problem comes from

Tile URLs exist in two hand-maintained copies. The backend owns them in
`src/agent/datasets/catalog/*.yml`; the frontend keeps a second copy in
`project-zeno-next/app/constants/datasets.ts` (`DATASET_CARDS[].tile_url`).
Nothing syncs them and no test compares them.

The frontend copy exists because only one of its three "add layer to map" paths
goes through the agent:

| path | tile URL source |
|---|---|
| chat `pick_dataset` | backend `dataset.tile_url` — correct |
| Data Catalog / default card | `DATASET_CARDS[].tile_url` — local copy |
| suggested-dataset nudge | `DATASET_BY_ID[id].tile_url` — local copy, even though a backend nudge triggered it |

So a layer's URL depends on how it was added. The copies have already drifted:

- **Global land cover** — backend templates the year and fills it from the
  resolved range (`items/global-land-cover-{year}`); the frontend pins
  `items/global-land-cover-2024`. Their colormaps also differ, class 1 being
  `[254, 254, 204, 255]` on the backend and `[254, 254, 204]` on the frontend.
- **Grasslands** — same shape: `items/grasslands-{year}` vs a pinned
  `items/grasslands-2022`.
- **Integrated alerts** — the case that surfaced this. The backend appends
  `&start_date=&end_date=`; the frontend card has no date params, and its only
  date-scoping code appends `&start_year=&end_year=`, which is the tree-cover-loss
  convention and wrong for this layer. The card also declares no default years,
  so nothing is appended at all and the layer renders unfiltered.

That last one is the general problem. **The date-filtering convention is
per-dataset** — query params by date, query params by year, or a year in the item
path — and the frontend hardcodes exactly one of the three. Worse, the backend did
not declare the convention either: `get_tile_services_for_dataset` inferred it from
hardcoded lists of dataset ids, so any second consumer had to re-derive it.

## What this changes

Makes the convention declarative and serves it, so there is one source of truth
instead of two copies and two inference sites.

**1. `tile_date_filter` in the catalog YAML** — one of `date_params`,
`year_params`, `year_in_path`, or absent for a layer with no time dimension.
Added to the five datasets that have one.

**2. `get_tile_services_for_dataset` reads it** instead of matching dataset ids:

```python
date_filter = getattr(selected_row, "tile_date_filter", None)
if date_filter == YEAR_PARAMS:
    ...
elif date_filter == DATE_PARAMS:
    tile_url += f"&start_date={start_date}&end_date={end_date}"
elif date_filter == YEAR_IN_PATH:
    tile_url = tile_url.format(year=end_date.year)
```

Behaviour is unchanged for every dataset — same branches, same output, keyed off
the YAML rather than `INTEGRATED_ALERTS_ID` / `LAND_COVER_CHANGE_ID` /
`GRASSLANDS_ID`, which are no longer imported here. The magic `canopy_cover = 30`
becomes `DEFAULT_CANOPY_COVER`, shared with the registry.

**3. `src/agent/datasets/layers.py`** — a `LAYERS` registry alongside the existing
`PALETTES` one, deriving per dataset: absolute `tile_url`, `date_filter`,
`start_date`, `end_date` (None = ongoing), `content_date_fixed`, and
`threshold_values` / `default_threshold` when the URL carries `{threshold}`.

Served URLs are absolute and keep their placeholders literal — `{z}/{x}/{y}`
always, plus `{threshold}` and `{year}` where they apply. `year_in_path` URLs are
`str.format` templates in the YAML, so their tile scheme is escaped as
`{{z}}/{{x}}/{{y}}`; the registry unescapes it. Relative eoapi paths are
absolutised with `rstrip("/")`, matching `src/shared/tile_urls.py` — the frontend's
`EOAPI_HOST` default has a trailing slash and currently produces `//raster/`.

Three invariants fail at import rather than at render time: an unknown
`tile_date_filter`, a param-appending convention on a URL with no query string
(the append uses `&`), and `year_in_path` with no `{year}`. A fourth — a
`{threshold}` placeholder with no `canopy_cover` values to fill it — was found by
its own test and is now enforced too.

**4. `/api/datasets/catalog` gains a `layers` array.** Additive: `datasets`
(colors) is untouched, and the frontend's Zod schema strips unknown keys, so the
deployed client is unaffected. The two lists stay independent because a dataset can
have tiles without colors (Integrated alerts, tree cover loss due to fires) and
colors without tiles.

## Scope

Backend only. The frontend PR that consumes `layers` and deletes
`DATASET_CARDS[].tile_url` depends on this being deployed, and is separate.

Not addressed here: the legend truncates any range to years
(`buildYearParam`), so a two-week filter renders as `YEAR 2026` whatever the
tile URL says. Also frontend, also separate.

## Tests

- `tests/unit/agent/test_dataset_layers.py` — registry derivation, placeholder
  unescaping, absolutisation, and each of the four import-time invariants.
- `tests/unit/agent/tools/test_tile_services.py` — rewritten: one case per
  convention plus a no-convention case, and a guard that fails if a new
  convention lands in the catalog with no coverage here.
- `tests/api/test_metadata.py` — three cases on the new array: every tiled
  dataset present with absolute URLs, the per-dataset conventions, and the
  colors/layers independence.

```
27 passed in 4.62s
```

Part of #821.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "project-zeno"
-version = "2026.9.2.1"
+version = "2026.9.8"
 description = "Language Interface for Maps & WRI/LCL data APIs."
 readme = "README.md"
 requires-python = "==3.12.8"
--- a/src/agent/datasets/catalog/global_land_cover.yml
+++ b/src/agent/datasets/catalog/global_land_cover.yml
@@ -4,6 +4,7 @@ context_layers: null
 parameters: null
 variables: null
 tile_url: /raster/collections/global-land-cover-v-2/items/global-land-cover-{year}/tiles/WebMercatorQuad/{{z}}/{{x}}/{{y}}.png?colormap=%7B%221%22%3A%20%5B254%2C%20254%2C%20204%2C%20255%5D%2C%20%222%22%3A%20%5B185%2C%20185%2C%2030%5D%2C%20%223%22%3A%20%5B36%2C%20110%2C%2036%5D%2C%20%224%22%3A%20%5B116%2C%20214%2C%20180%5D%2C%20%225%22%3A%20%5B107%2C%20174%2C%20214%5D%2C%20%226%22%3A%20%5B172%2C%20209%2C%20232%5D%2C%20%227%22%3A%20%5B255%2C%20241%2C%20131%5D%2C%20%228%22%3A%20%5B232%2C%20118%2C%2093%5D%2C%20%229%22%3A%20%5B255%2C%20205%2C%20115%5D%7D&assets=asset&expression=asset%2A%28asset%3C10%29%2A%28asset%3E0%29&asset_as_band=True
+tile_date_filter: year_in_path
 license: CC by 4.0
 geographic_coverage: Global
 resolution: 30 x 30 meters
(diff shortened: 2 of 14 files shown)


yellowcap:
we can revisit later

PR #822, opened 2026-09-09
feat(dashboards): add analysis templates for curated sections

yellowcap:
Replaces the near-real-time part of #813, rebuilt as a **system** rather than
one feature. #813 stays open as the source of the pieces this leaves out.

## What this is

An **analysis template** answers a standing question about an area — "what
has been disturbed here in the last two weeks?" — by writing a whole
dashboard section at once: a chart, the map layers that belong with it, and
a title and description written from the data.

It is not a chat workflow. Nothing is picked by a model along the way, so
the same request always produces the same section. The only model call
writes the words.

`nrt-monitoring` is the first template: a chart of integrated disturbance
alerts, a map of those alerts, and satellite imagery of the same area and
period.

**The agent is the only caller**, behind the experimental profile. No REST
endpoints in this PR — the system is built so a route can be added later
without changing a template.

## The measure of success

A second template is **one module, one registry entry, one skill section**.
No new tool, no `agent_config` change, no `dashboard_writer` change, no
migration, no router change. `docs/analysis-templates.md` is that recipe.

## The parts

| Where | What |
|---|---|
| `analysis_templates/registry.py` | the templates, as metadata only |
| `analysis_templates/base.py` | the interface, the result and error types, the shared helpers |
| `analysis_templates/writer.py` | the two transactional writes, and the `config` shape |
| `analysis_templates/nrt_monitoring/` | the first template |
| `agent/tools/analysis_sections.py` | two generic agent tools |
| `skills_md/analysis-templates.md` | the workflow the model follows |

## Decisions worth reviewing

**The registry holds metadata, not templates.** It imports nothing else, and
a test enforces that. Two callers need to know a template exists without
paying for what it does: `dashboard_writer` reads `SEALED_SECTION_TYPES` on
every dashboard write, and the agent's tool prompt lists the templates at
import time. `get_template()` imports a module on first use. A template's
`name` **is** its `dashboard_sections.type`, so there is one list of names
rather than two.

**A template takes the loaded dashboard, not an id.** It then picks its own
area and runs its own "already built" check, so the generic layer needs no
hook for either. It writes its own one-line summary, so the tools never
mention alerts, imagery or periods.

**One error family.** Everything a template raises is a `TemplateError`
whose message is written for the user. Each tool has one `except` clause; a
new template adds no error handling anywhere.

**Parameters are the template's own pydantic model**, with
`extra="forbid"`. Bounds are stated once, and a parameter a template does
not take is refused rather than ignored.

**Three generic agent tools**, not a set per template. Nothing in them names
a parameter — `params` is a dict the template's own model validates — so a
template driven by a threshold or a comparison area rather than a date range
uses them unchanged.

| The user says | Tool | What runs |
|---|---|---|
| "monitor this area" | `add_analysis_section(template, params?)` | `template.build()` |
| "refresh it", "anything new?" | `refresh_analysis_section(section?)` | `template.refresh()` with the parameters the section recorded |
| "show me 3 months instead" | `reconfigure_analysis_section(params, confirmed, section?)` | `template.refresh()` with new parameters |

A template implements one `refresh` and cannot tell the last two apart,
which is the point: re-running a section and moving it are the same
operation with different parameters. `writer` records what a section was
built with, so a refresh needs nothing from the caller
(`base.stored_params`), and `template.describe(section)` renders those
parameters for a message, because only the template knows what they mean.

Only `reconfigure` is confirmed, and it is gated by a **refusal**, not a
prompt: without `confirmed=True` the tool does nothing and tells the model
to `send_nudge` first. It answers a different question than the one on
screen and the previous answer is deleted, and a model can retry past an
instruction but not past a refusal. `refresh` asks the same question again,
so there is nothing to agree to.

## Sealed sections

A section a template wrote is read-only: its content is a record of one
build — one area, one period, the data as it was — so editing it would make
its own title and description untrue.

The rule lives in `dashboard_writer`, not the router, because the agent
tools bypass routers entirely. The router maps `SealedSectionError` to 409,
the agent tools to a reply that says what to do instead.

**Layout is exempt**: `position`, and a config replacement whose only
differing keys are `size`/`sizes`. The check is a **diff**, not a field
allowlist — PATCH replaces a config wholesale, so "resize it" and "resize it
and swap the tile_url" are the same shape of request.

Three back doors are closed with it:

- deleting a sealed section always cascades, or ungrouping would leave
  loose, editable widgets behind;
- `type` is absent from the update request and fixed to `"default"` on the
  create request, or a section would be one PATCH away from unsealed and a
  caller could create a section nothing can ever fill;
- `update_insight_display` refuses an insight a sealed widget references —
  the one path that could rewrite a section's content without touching a
  dashboard row.

The seal is ~198 lines of code over 8 files: one frozenset, two guard
functions, six call sites in the writer. A template never meets it — it
writes in one transaction and replaces in another, neither of which is
guarded.

## The refactor that carries its own weight

`resolve_dataset_layer()` (`src/agent/datasets/layers.py`) lifts the
tile-URL rules out of the dataset-selection subagent, where they needed an
LLM selection object and a pandas row. A dataset's catalog `tile_url` is a
template — relative to the eoapi host, carrying a canopy-cover threshold, a
year or a date range — and nothing outside a chat turn could resolve one.
`get_tile_services_for_dataset()` is now the adapter over it, passing its
*candidate* row so the extent filter on context layers is not quietly
undone.

`widget_configs.py` holds the map-widget config projections that lived
inside `add_map_widget` (`+46` new, `−49` deleted there), so a layer added
from chat and one added by a template are byte-identical.

**This is the commit with real regression risk** — it rewrites the
`pick_dataset` path every chat turn uses. It is `b31394fc`, self-contained,
and can be pulled into its own PR if you would rather review it separately.

## Deliberately not here

| Left out | Why |
|---|---|
| the text-highlight markup (#813 `0132ec8d`) | unrelated feature, its own PR |
| `POST .../sections/nrt-monitoring` and the refresh route, the `NrtSection*` schemas | the agent is the only caller for now |
| `MosaicCreateResponse.tile_url` / `tilejson_url` | unrelated to sections |
| the two frontend handoff docs (475 lines) | a contract for endpoints this PR does not add |
| `title`, `description`, `force`, `window_days`, `max_cloud_cover` | no caller passed them; the imagery ones are template constants |
| alerts bucketed by day instead of month | changes `POST /api/analyze` output for existing clients — see below |

## Known limits, written down rather than solved

- **One area.** A template covers the dashboard's first AOI. Portfolios are
  a dashboard-level question.
- **Parameters reach the model as a dict**, described in the registry's
  `params_help` rather than shown in a tool signature. If that proves
  unreliable as templates grow, the answer is per-template tools generated
  from the registry, not a parameter in the generic layer.
- **Synchronous.** A build runs to completion, which takes tens of seconds.
  That decision gets made again when a route is added.
- **The alerts chart is still monthly**, so the 14-day default window
  renders as one or two points. Moving it to daily buckets renames the
  chart's `x_axis` from `month` to `alert_date`, which is visible to
  existing `/api/analyze` clients, so it is left for its own PR.

## Verification

775 tests pass. New: `tests/api/test_analysis_templates.py` (built section,
imagery degradation, analytics failure, already-built guard, the
parameter round trip a refresh depends on),
`tests/unit/agent/tools/test_analysis_section_tools.py` (dispatch, the
two rebuild verbs, the enforced confirmation), `tests/unit/api/analysis_templates/test_registry.py`
(loops over **every** registered template, so a new one is covered the day
it is added — its parameter model, its `describe`, and the registry's
import-lightness),
`tests/api/test_dashboards.py` (one test per blocked action, layout allowed,
the smuggling refusal), plus the layer resolver and summary tests.

The migration round-trips up and down against a scratch database.

Two behaviours the mocked tests of #813 could not have found, kept here with
assertions: every consumer gets its own copy of the AOI, because
`AnalyticsHandler.pull_data` rewrites its input in place; and the figures
the model reads are rounded, because it will otherwise quote
`153409.28893796972 ha`.

PZB-1295

https://claude.ai/code/session_01Vs3qv4aWbUtdHEKS7uX8ZJ

### DIFF
--- a/src/agent/datasets/layers.py
+++ b/src/agent/datasets/layers.py
@@ -1,0 +1,178 @@
+"""Resolve a catalog dataset into a renderable map layer.
+
+A dataset's ``tile_url`` in the catalog is a template: it may be relative to
+the eoapi host, and it may carry a canopy-cover threshold, a year, or a date
+range that only the caller knows. Turning it into a URL a map can render is
+the same job whether the dataset was picked by the LLM in chat or named by an
+API caller, so it lives here rather than inside the dataset-selection
+subagent.
+
+``pick_dataset`` adapts its own selection objects onto this function; the
+the analysis templates call it directly. One definition means a map widget
+built by a template and one built from chat carry the same URL.
+"""
+
+from dataclasses import dataclass, field
+from datetime import date
+from typing import Optional
+
+from src.agent.datasets.config import DATASETS
+from src.agent.datasets.handlers.analytics_handler import (
+    FOREST_CARBON_FLUX_ID,
+    GRASSLANDS_ID,
+    INTEGRATED_ALERTS_ID,
+    LAND_COVER_CHANGE_ID,
+    TREE_COVER_ID,
+    TREE_COVER_LOSS_BY_DRIVER_ID,
+    TREE_COVER_LOSS_BY_FIRES_ID,
+    TREE_COVER_LOSS_ID,
+)
+from src.shared.config import SharedSettings
+
+#: Datasets whose tiles are cut at a canopy-cover threshold.
+_THRESHOLD_DATASETS = (
+    TREE_COVER_LOSS_ID,
+    TREE_COVER_ID,
+    TREE_COVER_LOSS_BY_DRIVER_ID,
+    TREE_COVER_LOSS_BY_FIRES_ID,
+    FOREST_CARBON_FLUX_ID,
+)
+DEFAULT_CANOPY_COVER = 30
+
+#: Tree cover loss tiles carry whole years, and only these are published.
+TCL_TILE_YEARS = range(2001, 2026)
+
+
+@dataclass(frozen=True)
+class DatasetLayer:
+    """One dataset rendered for one date range: what a map needs, nothing
+    else. The prose fields of the catalog record stay out by design — this
+    is what gets snapshotted into a dashboard widget."""
+
+    dataset_id: int
+    dataset_name: str
+    tile_url: str
+    start_date: str
+    end_date: str
+    context_layer: Optional[str] = None
+    #: Companion layers to draw with the dataset, as
+    #: ``{"name": ..., "tile_url": ...}``.
+    context_layers: list[dict] = field(default_factory=list)
+    #: The parameters the tile URL was built with, as
+    #: ``{"name": ..., "values": [...]}``.
+    parameters: Optional[list[dict]] = None
+
+
+def get_dataset_record(dataset_id: int) -> dict:
+    """The catalog record for a dataset id; raises ``ValueError`` if unknown."""
+    record = next(
+        (ds for ds in DATASETS if ds["dataset_id"] == dataset_id), None
+    )
+    if record is None:
+        raise ValueError(f"Dataset not found: {dataset_id}")
+    return record
+
+
+def _canopy_cover(parameters: Optional[list[dict]]) -> int:
+    for parameter in parameters or []:
+        if parameter.get("name") == "canopy_cover" and parameter.get("values"):
+            return max(parameter["values"])
+    return DEFAULT_CANOPY_COVER
+
+
+def resolve_dataset_layer(
+    dataset_id: int,
+    start_date: str,
+    end_date: str,
+    *,
+    context_layer: Optional[str] = None,
+    parameters: Optional[list[dict]] = None,
+    record: Optional[dict] = None,
+) -> DatasetLayer:
+    """Build the renderable layer for a dataset over a date range.
+
+    ``start_date`` / ``end_date`` are ISO dates the caller has already
+    clamped to the dataset's coverage (``datasets.dates.revise_date_range``).
+    ``context_layer`` names one of the dataset's context layers by its
+    catalog ``value``; ``parameters`` are ``{"name", "values"}`` dicts, of
+    which only ``canopy_cover`` currently changes a URL.
+
+    ``record`` overrides the catalog lookup with the dataset row the caller
+    already holds. ``pick_dataset`` passes its candidate row, whose context
+    layers have been filtered to the ones that actually cover the selected
+    area — resolving from the catalog instead would quietly undo that
+    filter.
+    """
+    record = record if record is not None else get_dataset_record(dataset_id)
+    start = date.fromisoformat(start_date)
+    end = date.fromisoformat(end_date)
+
+    tile_url = record["tile_url"] or ""
+    if tile_url and not tile_url.startswith("http"):
+        tile_url = SharedSettings.eoapi_base_url + tile_url
+
+    context_layers: list[dict] = []
+    if context_layer and record.get("context_layers"):
+        selected = next(
+            (
+                layer
+                for layer in record["context_layers"]
+                if layer.get("value") == context_layer
+            ),
+            None,
+        )
+        if selected:
+            context_layers.append(
+                {
+                    "name": selected.get("value"),
+                    "tile_url": selected.get("tile_url"),
+                }
+            )
+
+    if dataset_id in _THRESHOLD_DATASETS:
+        canopy_cover = _canopy_cover(parameters)
+        if dataset_id != TREE_COVER_ID:
+            canopy_tile_url = next(
+                (
+                    parameter["tile_url"]
+                    for parameter in record.get("parameters") or []
+                    if parameter["name"] == "canopy_cover"
+                ),
+                None,
+            )
+            if canopy_tile_url:
+                context_layers.append(
+                    {
+                        "name": "canopy_cover",
+                        "tile_url": canopy_tile_url.replace(
+                            "{threshold}", str(canopy_cover)
+                        ),
+                    }
+                )
+        tile_url = tile_url.replace("{threshold}", str(canopy_cover))
+
+    if dataset_id in (TREE_COVER_LOSS_ID, TREE_COVER_LOSS_BY_FIRES_ID):
+        if end.year in TCL_TILE_YEARS:
+            tile_url += f"&start_year={start.year}&end_year={end.year}"
+        else:
+            tile_url += (
+                f"&start_year={TCL_TILE_YEARS.start}"
+                f"&end_year={TCL_TILE_YEARS.stop - 1}"
+            )
+    elif dataset_id == INTEGRATED_ALERTS_ID:
+        tile_url += f"&start_date={start_date}&end_date={end_date}"
+    elif dataset_id in (LAND_COVER_CHANGE_ID, GRASSLANDS_ID):
+        # Annual raster item in the URL; the dates are already clamped to the
+        # dataset's own coverage.
+        tile_url = tile_url.format(year=end.year)
+
+    return DatasetLayer(
+        dataset_id=dataset_id,
+        dataset_name=record["dataset_name"],
+        tile_url=tile_url,
+        start_date=start_date,
+        end_date=end_date,
+        context_layer=context_layer,
+        context_layers=context_layers,
+        parameters=parameters,
+    )
--- a/src/agent/tools/analysis_sections.py
+++ b/src/agent/tools/analysis_sections.py
@@ -1,0 +1,336 @@
+"""The agent's two doors onto the analysis templates.
+
+An analysis template builds a whole dashboard section in one call — see
+``src.api.services.analysis_templates``. These two tools are generic: they
+resolve the dashboard, look the named template up in the registry, hand it
+its own validated parameters and report the summary it wrote. Neither knows
+what any template contains, so a new template needs no new tool and no
+change here — it is one registry entry and one skill section.
+
+Unlike the other dashboard primitives, these do not snapshot work the
+conversation already did: the template does the work itself. So neither
+needs pick_dataset or show_imagery to have run — only a dashboard with an
+area.
+
+A section a template writes is sealed afterwards: the editing tools refuse
+its content. ``update_analysis_section`` is the one exception, and it
+rebuilds the section rather than editing it.
+"""
+
+from typing import Annotated, Dict, Optional
+
+from langchain_core.tools import tool
+from langchain_core.tools.base import InjectedToolCallId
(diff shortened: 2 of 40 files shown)


src/api/repositories/dashboard_writer.py
```diff
@@ -431,6 +784,9 @@ async def remove_section(section_id, *, delete_widgets: bool = False) -> bool:
         if section is None:
             return False
 
+        if section.type in SEALED_SECTION_TYPES:
+            delete_widgets = True
+
```
01painadam, line 789: Deleting a sealed section removes its widgets but leaves the template-built insight behind, while replace_section_widgets deletes the insight it replaces via delete_unreferenced_insights.

The two paths should agree: collect the widgets' insight_ids before the delete and pass them to delete_unreferenced_insights when the section is sealed, so a deleted monitoring section does not leave an orphan chart in the user's insight list.

src/agent/datasets/layers.py
```diff
@@ -0,0 +1,178 @@
...
+
+A dataset's ``tile_url`` in the catalog is a template: it may be relative to
+the eoapi host, and it may carry a canopy-cover threshold, a year, or a date
+range that only the caller knows. Turning it into a URL a map can render is
+the same job whether the dataset was picked by the LLM in chat or named by an
+API caller, so it lives here rather than inside the dataset-selection
+subagent.
+
+``pick_dataset`` adapts its own selection objects onto this function; the
+the analysis templates call it directly. One definition means a map widget
```
01painadam, line 11: small typo - the the analysis

src/api/services/analysis_templates/registry.py
```diff
@@ -0,0 +1,138 @@
...
+        raise UnknownTemplateError(name)
+    module = import_module(f"{_PACKAGE}.{entry.module}")
+    template = module.TEMPLATE
+    # The registry and the module each name the template. They must agree,
+    # or a section would be written under a type nothing can refresh.
+    assert template.entry.name == entry.name, (
+        f"{entry.module}.TEMPLATE is '{template.entry.name}', "
+        f"registered as '{entry.name}'"
+    )
+    return template
```
01painadam, line 129: This assert is the only thing stopping a registry entry and its module from disagreeing on the template name, and assert statements are stripped under python -O.

Raise a real exception (a RuntimeError or a registry-specific error) so the check holds in every runtime.

src/api/services/analysis_templates/writer.py
```diff
@@ -0,0 +1,130 @@
...
+def _config(
+    template: "AnalysisTemplate[Any]",
+    content: SectionContent,
+    params: BaseModel,
+) -> dict:
+    return {
+        "template": template.entry.name,
+        "params": params.model_dump(mode="json"),
+        **content.config,
+    }
```
01painadam, line 43: **content.config is spread after the reserved keys, so a template that puts template or params in its own config silently overwrites the generic layer's record and breaks refresh.

Spread content.config first and set the two reserved keys last, or refuse the collision.

src/api/services/analysis_templates/writer.py
```diff
@@ -0,0 +1,130 @@
...
+    params: BaseModel,
+) -> TemplateResult:
+    """Swap a section's whole contents for freshly gathered ones.
+
+    The section row survives, so its id and its place on the dashboard
+    hold. Insights the previous widgets showed are deleted once nothing
+    points at them: they were this section's own content, for parameters
+    the section no longer covers.
+    """
+    written = await dashboard_writer.replace_section_widgets(
```
01painadam, line 102: On a published dashboard (is_public=True), the insight this refresh writes is private (persist_insight default) while the one it deletes was made public by the publish cascade. Viewers of the public dashboard lose the chart after a refresh.

Either copy the dashboard's is_public onto the new insight in replace_section_widgets/add_section_with_widgets, or re-run the cascade after a template write.

src/api/services/analysis_templates/nrt_monitoring/recipe.py
```diff
@@ -0,0 +1,343 @@
...
+        language=language,
+    )
+    if not analysis.data.success:
+        raise DataUnavailableError(
+            f"Could not retrieve alert data for '{aoi['name']}': "
+            f"{analysis.data.message}"
+        )
+
+    # Charts only, no narrative: the section's description carries the words.
+    insight_id = await persist_insight(
```
01painadam, line 166: The insight is persisted here, before the imagery search, the summary call and the section write. If anything after this point fails (dashboard deleted, DuplicateInsightWidgetError, a DB error on commit, an unexpected exception from the analytics or imagery client), the user is left with an orphan chart-only insight that shows up in GET /api/insights.

Persist the insight immediately before writer.write_section/replace_section, or wrap the rest of the gather in a try that deletes it on failure.

src/api/services/analysis_templates/nrt_monitoring/recipe.py
```diff
@@ -0,0 +1,343 @@
...
+        Insight(charts=analysis.charts),
+        user_id=user_id,
+        thread_id="",
```
01painadam, line 169: The insight carries no statistics_ids, so it has no dataset or AOI provenance: the insight filters on GET /api/insights?dataset_id=...&aoi_source=... will never find it and "how was this generated" has nothing to show.

AnalyzeService/AnalyticsHandler produce statistics rows; consider threading their ids through as the deterministic analyze job does.

src/api/services/analysis_templates/nrt_monitoring/recipe.py
```diff
@@ -0,0 +1,343 @@
...
+        params: NrtParams,
+        language: str = DEFAULT_LANGUAGE,
+    ) -> TemplateResult:
+        aoi = first_aoi(dashboard)
+        if aoi is None:
+            raise DataUnavailableError(
+                f"Dashboard '{dashboard.name}' has no area to monitor."
+            )
+
+        start_date, end_date = await resolve_period(params.days)
```
01painadam, line 265: resolve_period is called here and again inside _gather, so the period checked for duplicates and the period stored can differ if the two calls straddle midnight (or if the clamp changes between them). Resolve once and pass start_date/end_date into _gather.

src/agent/tools/analysis_sections.py
```diff
@@ -0,0 +1,453 @@
...
+    first.
+    """
+    state = state or {}
+    dashboard, target, template = await _resolve_target(
+        state, dashboard_id, section, tool_call_id, "refresh_analysis_section"
+    )
+    if dashboard is None:
+        return template  # the error command
+
+    return await _rebuild(
```
01painadam, line 337: stored_params builds the template's params model from what the section stored, with extra="forbid". A section written before a parameter was renamed or removed, or with a value outside a later-tightened bound, raises ValidationError here, which is not a TemplateError and is not caught.

Wrap it like _parse_params in reconfigure and return an error telling the model to reconfigure_analysis_section with fresh parameters (or fall back to the defaults, which is what the base.stored_params docstring promises for the missing-field case).

src/agent/tools/analysis_sections.py
```diff
@@ -0,0 +1,453 @@
...
+    try:
+        parsed = _parse_params(template, params)
+    except ValidationError as error:
+        return _params_error(template, error, tool_call_id)
+
+    # The refusal is the guard, not the instruction in the docstring: this
+    # answers a different question than the one on screen, and the previous
+    # answer is deleted. A model can retry past an instruction, not past a
+    # refusal.
+    if not confirmed:
```
01painadam, line 396: confirmed is set by the model, so this refusal is a prompt-level guard, not enforcement; the comment reads as though it were stronger. Either soften the comment or record the nudge in state (send_nudge could set a flag) and check that here.

src/agent/tools/common.py
```diff
@@ -98,10 +123,29 @@ def format_sections(dashboard: DashboardOrm) -> str:
...
+        + (
+            " [read-only]"
+            if section.type in dashboard_writer.SEALED_SECTION_TYPES
+            else ""
+        )
+        for section in sections
     )
 
 
+def _section_or_sealed(row):
```
01painadam, line 136: This refusal assumes the caller is adding a widget, but edit_dashboard_section resolves its target through resolve_section too (edit_dashboard_section.py:41).

Naming a sealed section for a retitle returns "cannot take new widgets. Put the widget in another section" instead of the sealed-edit message, and the sealed_error_command branch in that tool becomes reachable only when section is omitted. Either let resolve_section return the sealed row and have each caller decide (edit falls through to update_section, which raises properly), or give _section_or_sealed a caller-neutral message.

01painadam (commented):
Well-structured feature work. Claude says its "good to merge" after looking at four things: 
1. the insight row is persisted before the imagery search, the summary call and the section write, so any later failure leaves an orphan in the user's insight library
2. on a published dashboard a refresh swaps a public insight for a private one, so viewers lose the chart
3. deleting a sealed section orphans its insight while a refresh cleans up, an inconsistency
4. refresh_analysis_section can raise an uncaught pydantic ValidationError when a section's stored parameters no longer fit the template. 

None of these look too hard to fix and none touches the design.

Architecture observations (Claude):

- The registry-as-leaf design (registry.py imports nothing) is what lets dashboard_writer know the seal list without loading analytics or model clients. It does mean src/api/repositories now imports from src/api/services, an upward dependency the codebase did not have before. Acceptable because the leaf is metadata only, and a unit test (test_the_registry_stays_import_light) pins that property.
- "Deterministic" holds for the data and the widgets. The title and description are written by a model call, so two builds for the same area and period produce the same chart and layers but not the same words. The fallback pair is deterministic. Worth knowing when reading the docs, which say the only model call writes the words, and the PR body, which says identical sections.
- A build is synchronous inside a tool call and takes tens of seconds (analytics pull, mosaic search, model call). Documented as a known limit; fine for the experimental profile, a design decision for any HTTP route later.
- The refactor of get_tile_services_for_dataset onto resolve_dataset_layer is behaviour-preserving for every dataset in the catalog, and quietly fixes two latent crashes (unmatched context layer name, missing canopy tile URL).

yellowcap:
Immutability is no a good pattern, it complicates things unnecessarily. Closing this in favor of a simplified #840

PR #823, opened 2026-09-14
fix(charts): lead the LGMS analysis with the annual-average chart

01painadam:
Swaps the order of the charts returned by `LGMSChartGenerator`: the annual-average hierarchical chart now leads at position 0, with the three time-series roll-ups (full detail / by category / summary) following at positions 1-3. Their relative order is unchanged.

## What was tested

- `tests/unit/api/services/test_chart_generators.py` updated to the new positions; the lead position of the annual-average chart is pinned in `test_lgms_hierarchical_chart_is_axis_less`
- Full unit tier: 635 passed
- pre-commit (ruff, ruff-format, mypy) clean


yellowcap (approved):
So tidy! Nice ordering! 🧹

PR #824, opened 2026-09-15
fix(pick_dataset): anchor the selector on today's date

01painadam:
## What changed

The dataset selector (`pick_dataset` subagent) is now told today's date. `DATASET_SELECTOR_PROMPT` gains a `{today}` placeholder with a two-line rule beside the existing "use the dataset's own available range" sentence: a `content_date` that ends in "present" runs up to today, so a request for the current year or recent months is inside its range. `select_best_dataset` takes an optional `today: date` (default `date.today()`, the same source `tool.py` already uses for the tile end date) so the rendered prompt is testable.

## Why

The selector sees coverage only as free text (`content_date: 1 December 2023 - present`) and nothing about the current date, so it read "present" against its own training-time present. Production trace `099ba22993ccae4e61056626213ee5f0` (2026-09-15, Land & Carbon Lab team account): *"Now add disturbance alerts to the map for 2026"* → pick_dataset answered *"No dataset selected: … no data available for that year. The Integrated Alerts dataset … only contains data from December 2023 to the present."* The orchestrator then told the user alerts *"only provide data up to the present (late 2024)"*, while the same session had just analysed alerts through September 2026.

Reproduced live before the change (same query, integrated-alerts candidates, dates 2026-01-01..2026-12-31): 0/2 selected, reason *"Since 2026 is in the future, no dataset c…"*.

## What was tested

- `uv run pytest tests/unit`: 638 passed. New `tests/unit/agent/tools/test_pick_dataset_today.py` (3 tests, dependency-free: a capturing stand-in for `SMALL_MODEL`) asserts the rendered system prompt states the given date, carries the "present" rule, and defaults to the real today.
- `uvx pre-commit run --all-files`: green (calver bump included).
- Live, real Gemini, after the change: "disturbance alerts for 2026" and "Now add disturbance alerts to the map for 2026" → Integrated alerts selected **6/6**; control "tree cover loss in 2026" → declined **3/3** (TCL's `content_date` ends 2025, as it should).
- A first draft used a separate "Dates" section; a same-day A/B on "net carbon sinks versus sources" (2024 dates, 8 trials each) showed it made the selector hedge (4 declines vs 2 without). The trimmed two-line form in this PR selects the flux dataset **8/8**. Commit history keeps that step visible.
- `tests/tools/test_pick_dataset.py -k "expected_dataset or context_layer or hallucinated"` (real LLM, 65 cases): 64 passed / 1 failed on two runs, a **different** case each time (flux, then TCL-by-fires) with 3-4 reruns per run, i.e. the suite's existing non-determinism; the case that failed on the first run passed on the second.

## Notes for reviewers

- Only `tool.py` formats `DATASET_SELECTOR_PROMPT`; `tests/unit/agent/tools/skills/test_skills.py` reads it but does not format it.
- A GOLD case for "alerts for the current year" is being added to gnw-gold-evals separately so this regression is caught by the release gate.
- Related session findings, not in this PR: PZB-1406 (protected-area analytics failing since 2026-08-25) and PZB-1392 (protected-area name resolution).


danscales (approved):
Looks good!

Issue #36, opened 2024-12-06
Add human in the loop checkpoints


yellowcap:
### Goal
Implement checkpoints that interrupt agents as they works through its tasks, giving the user an opportunity to given feedback that adjusts the plan. This feedback could take many forms (prompt or UI) for example

- When an agent uses the `location-finder` tool which returns multiple candidate areas, the user should have an opportunity to select the correct one (or refine their prompt and try again)
- When an agent decides to add a data layers to the map, the LLM should activate a UI elements so the user can select the layers they want to see

The implementation should allow the front end to facilitate user interaction at these decision-points.

### Scope
API implementation only

### User Stories & Acceptance Criteria
- When I am interacting with an LLM, I want opportunities to interact with the LLM, so that I can correct the LLM or refine my prompt to get to a better/more correct result
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]
  - Given [context], when [specific action], then [response or consequence to action]

### Risks & questions
- Should these decision points be hard coded?
- what options should the model present to the user?
  - we need to list and explore these against the use case & designs e.g. select an area from candidates

### Solution Hints
https://daoleo.medium.com/humans-in-a-loop-harnessing-persistence-and-memory-efficiency-213b7f40a756

### Future work
...

yellowcap:
We have a simple HIL checkpoint for selecting areas. Refinements are going into new tickets.

PR #826, opened 2026-09-15
fix(pick_aoi): rank protected areas on their leaf name, not the shared designation

01painadam:
Tickets: **PZB-1392** (the bug), **PZB-1407** (durable ingest follow-up), **gnw-gold-evals PR #36** (the `aoi/designations` cohort).

Scorer only. The retrieval half moved to **PZB-1411** on `fix/pick-aoi-word-similarity-fallback`.

## The bug

Asked for a protected area by its English name plus its designation, `pick_aoi` silently returned a same-designation area in another country. Production traces, 2026-09-15:

| prompt | resolved to | should be |
|---|---|---|
| Show me Sankuru National Reserve | Samburu, National Reserve, **KEN** | Sankuru, Réserve Naturelle, **COD** |
| Go to Yaguas National Park | Yanga, National Park, **AUS** | Yaguas, Parque Nacional, **PER** |
| Show me Chiribiquete National Park | Chiquibul, National Park, **BLZ** | La Serranía de Chiribiquete, **COL** |

`src/ingest/ingest_wdpa.py` composes `aois.name` as `"<leaf>, <designation>, <ISO3>"`, and the designation is in the source's own language. A source-narrowed search returns rows that all share segments 2 and 3, so scoring whole names ranked on what every candidate has in common:

```
'Sankuru National Reserve' vs 'Samburu, National Reserve, KEN'   0.4674  <- won
'Sankuru'                  vs 'Sankuru, Réserve Naturelle, COD'  0.4442
```

`", National Reserve, "` is most of the string. `subtype` is constant across WDPA so hierarchy contributed nothing, and protected areas raise no `aoi_choice` nudge (`check_duplicate_aois` is GADM-only), so the wrong polygon was analysed and the answer presented as fact.

## The fix

Selection scores the **leaf** — the only segment that identifies the place — with a smaller whole-name term that keeps the parent hierarchy meaningful, the unchanged hierarchy preference, and an exact-name term.

**A stored name matches if it matches under either word-order convention:** as a leading run of the term leaf (English, designation last — `"Sankuru National Reserve"` → `"Sankuru"`), or as a symmetric word-for-word covering of it (every other convention, designation first). Symmetric so a name is rewarded neither for being short nor for being long. Order-free coverage is why native-order terms work at all: `alternatives` exists to carry native spellings, and a term that scores the right row near zero can only ever raise a decoy.

**An exact name wins outright** (+0.25). `_format_aoi_candidate` builds an `aoi_choice` option from a row's full stored name and documents that resubmitting it re-resolves to that row. No weighting can promise that: GADM children often repeat their parent's name (`"Senga, Senga, Butezi, Ruyigi, Burundi"`), so child and parent tie on every name term and hierarchy hands the user the parent — a coarser area than the one they clicked. The bonus is a guarantee only while it exceeds the widest hierarchy gap, so that is **asserted at import time** beside the existing hierarchy assertion, derived from `_HIERARCHY_SCORES` rather than hard-coded. Both sides are accent-, case-, punctuation- and whitespace-normalised, so de-accented English keys equal to the accented stored name.

**No prefix bonus.** The leaf comparison already scores an exact leaf 1.0, so any bonus on top double-counts — and a bonus gated on "not an exact leaf" rewards every candidate *except* the exact match, which is how an intermediate version of this branch scored `"Niger"` 0.8500 against itself and 0.8583 against `"Nigeria"`.

## Evidence

Two fixtures, and a harness built specifically because the first one gave false confidence.

`tests/fixtures/aoi_designation_candidates_v1.json` records production's frames for the 22-case cohort. Tuned against that alone, an earlier version of this branch scored 20/22 while regressing five separate things it could not see — every case there carries *both* search terms *and* a source-narrowed frame. `test_scoring_regressions.py` adds one probe per blind spot and records what `main` scores on each, because "better than the fixture said" is not the bar; "no worse than what shipped" is.

| probe | main | this branch |
|---|---|---|
| cohort, both terms | 14/22 | **21/22** |
| single term, user wording only | 7/22 | **13/22** |
| exact full-name round trip, wrong | 0/150 | **0/150** |
| un-narrowed frames | 3/4 | **3/4** |
| native-order terms | 0/3 | **2/3** |
| `Niger` > `Nigeria` | yes | yes |
| bare `Para` > `Paraná` | yes | yes |
| recorded replay frames moved | — | **0** |

Pinned ladders all hold, with wider margins: exact/prefix/unrelated, country/state/site, and Lisboa over `"Lisbon, Forest Preserve, USA"` — which on `main` holds by **0.000802** and here by 0.0986.

The round trip is the sharpest probe: 150 real GADM rows, each searched under its own stored name, which is exactly the `_format_aoi_candidate` contract.

## What still fails, and why

Three xfails, each with its mechanism in the test:

- **ch-aoi-134** (Okapi) is retrieval-bound, not scoring-bound. `similarity('Okapis, Réserve de Faune, COD', 'Okapi')` is 0.1724, under the 0.2 `pg_trgm.similarity_threshold`, so the row is in neither recorded frame and no scorer can reach it. **PZB-1411.** ch-aoi-149 is the control: asked natively, canonical `"Okapis"` does retrieve it, and that case passes.
- **`"Salonga"` un-narrowed** selects the country Tonga (0.7667) over the park (0.7012). The name evidence is right; a country simply outranks a named site by more hierarchy than the name can recover. Wants a taper in `_HIERARCHY_SCORES` and its own before/after. Fails on `main` too.
- **`"Parc National de l'Ivindo"`** selects `"Parc National de Toubkal"`, whose stored *leaf* is itself mostly designation words and so covers three of the term's four. No comparison of two strings can separate that; it needs the designation in its own column, which is SPEC-PR8. Fails on `main` too.

## Rejected on measurement

Weighting each word by how rare it is **within the retrieved frame** — the idea states the bug directly, since the segments every candidate shares should not decide. It scored *identically* to plain word coverage on every probe, so the frame sensitivity it introduces (a candidate's score depending on what else was retrieved; unit tests that can no longer score a pair in isolation) would buy nothing on this catalogue. Word coverage alone was also tried and costs 8 single-term cases, dropping to 5/22, below `main`'s 7.

## One known limit, stated so it is not rediscovered

The leaf comparison does **not** penalise a short name. A candidate whose whole leaf is `"Rio"` scores a perfect 1.0 against the term leaf `"rio pure national park"`, because `"rio"` is a leading run of it, and taking the better of two measures cannot take that back. Word coverage alone scores it 0.65 against 0.83 for `"rio pure"`; the better-of-two erases the difference. Separating that pair is left to the whole-name and exact-name terms, which do it. If a one-word decoy ever wins a frame, the fix is a coverage floor in `_leaf_similarity`, not a weight — that is written in the docstring.

## Deliberate change to a pinned test

`test_each_candidate_scores_against_its_best_term` asserted that the user's wording **alone** picked a foreign park and that only the canonical leaf rescued the intended row. That was this defect written down as an expectation. Rewritten as `test_the_designation_no_longer_decides_which_park_is_chosen`, asserting both term sets select Botum Sakor, with the four PZB-1392 pairs parametrised beside it.

## Rollback

`git revert`. Deliberately no feature flag: `aoi_normalizer_enabled` gates term construction *and* scorer-vs-LLM together, so flipping it would reintroduce PZB-1272 rather than roll this back, and a second boolean would double the behaviour surface with an "off" branch nothing exercises.

## Pre-existing gap, noted not fixed

`scripts/record_aoi_pick_aoi_fixtures.py`, named in `tests/tools/conftest.py`'s error messages, does not exist, so `tests/fixtures/aoi_pick_aoi_v2.json` cannot be regenerated by the documented command. Out of scope. This PR changes no SQL, so those recorded frames stay fully valid, which is what makes the replay suite a meaningful gate here.

## Contracts touched

None of the six cross-repo seams. `AgentState` unchanged, stream unchanged, `GET /api/aois` untouched, no new agent tool, no migration, no new setting.

## Gate

```
uv run pytest tests/unit tests/api
    1035 passed, 3 xfailed, 7 warnings in 260.08s (0:04:20)

AOI_PICK_AOI_FIXTURES_MODE=replay uv run pytest tests/tools/test_pick_aoi.py
    21 passed, 1 warning in 23.37s

uvx pre-commit run --all-files
    bump CalVer / trailing-whitespace / check-yaml / end-of-file-fixer /
    check-added-large-files / detect-private-key / ruff / ruff-format / mypy
    ... all Passed
```

Run against a private `postgis/postgis:17-3.5` container rather than the shared `:5434`, which is the documented workaround: on the shared instance this branch showed a *shifting* set of 8-11 teardown errors across two runs and an unrelated `test_traces_endpoints::test_list_pagination` failure on leftover rows, none in AOI code and none reproducible in isolation.

https://claude.ai/code/session_01LDw5V6swtikLFcMcmafYWF


01painadam:
Putting this back to draft. A high-effort review with live database probes found correctness regressions against `main`, and I reproduced the scorer half independently on this branch:

```
term 'Niger'  vs 'Niger'                 0.8500
term 'Niger'  vs 'Nigeria'               0.8583   <- wrong country wins
term 'Para'   vs 'Pará, Brazil'          0.7450
term 'Para'   vs 'Paraná, Brazil'        0.7567   <- reverses PZB-1272
term 'Kahuzi-Biega' vs 'Kahuzi-Biega, Parc National, COD'  0.5418
term 'Kahuzi-Biega' vs 'Kahuzi, Congo'                     0.5607   <- district beats the park
term 'Parque Nacional Yasuní' vs 'Yasuní, Parque Nacional, ECU'  0.3334
term 'Parque Nacional Yasuní' vs 'Parque Nacional Natural, …'     0.4777   <- native word order
```

Cause of the first two: retiring the exact-leaf bonus while keeping the prefix bonus behind `term_leaf != candidate_leaf` leaves the exact match as the only candidate denied a bonus. The existing Pará test passes only because its term carries a parent, so `startswith` is false and the bonus never fires.

Also found, not yet reproduced by me: a candidate whose whole leaf equals a one-word prefix scores a perfect 1.0 with no length weighting; feeding 150 real GADM rows back as their own full name picks the parent admin unit 11 times against 0 before, which breaks the disambiguation round-trip; `%>` is `word_similarity`, not containment, and admits junk at 0.6; the fallback pads the frame `check_duplicate_aois` reads, suppresses the empty-retry safety net, fires on the `place` term where it can only add designation-shared decoys, and inherits an ordering that can cut the row it exists to find; and the cohort fixture was recorded through `GET /api/aois`, the one caller deliberately excluded from the fallback, so it does not measure what ships.

The core insight stands and the original bug is real. The weighting and the retrieval half need another pass with a wider instrument.

PR #827, opened 2026-09-16
feat(insights): show curated LGMS charts without interpretation

yellowcap:
## Summary

The key goal of this PR is to show curated LGMS charts in the agent chat instead of routing LGMS through the generative (code-executor) insight path, which is unreliable for this dataset's sparse, section-flattened metrics.

- Add a catalog flag, `curated_only`, set on `land_ghg_inventory.yml`. A dataset marked this way always gets its deterministic chart generator's output and skips the code executor entirely — no interpretation, no comparisons across areas, no restyling.
- Share the curated-chart-building step (`src/api/services/charts/curated.py`) between the REST `AnalyzeService` path and the agent's `generate_insights` path, so there is one place that matches a generator, localises its text, and applies chart colors.
- Guard every surface that could leak LGMS values through another route: `pull_data` (refuses multi-area pulls), `update_insight_display` (refuses restyles), `inspect_view_context` (withholds chart rows), and the session-level instructions the model sees every turn.

## Next step (out of scope here)

Right now `curated_only` is all-or-nothing: a flagged dataset can *only* show curated charts, with a fixed narrative and no model-written insight text. A follow-up is needed to let the agent *choose* curated charts for other datasets (the 12 generators already registered in `DETERMINISTIC_GENERATORS` cover more than LGMS) while still allowing the generative path when the query calls for it. That requires eligibility rules per generator, and reusing `InsightTextGenerator` to narrate curated charts instead of staying silent. Design notes: `docs/curated-charts-in-agent-plan.md` ("Later" section).

## Test plan

- [x] `tests/unit/agent/tools/test_curated_only.py`
- [x] `tests/unit/api/services/test_curated_charts.py`
- [x] Verified live against a real `WRI_BEARER_TOKEN`: a Pará LGMS request returns the 4 curated charts with no executor run, and the agent declines a figure, an area comparison, and a pie chart request.


solomon-negusse (approved):
Looks good!

01painadam (commented):
A few notes:

1. The colour step in the shared builder is wrong, and collides with `analyze-chart-colors` (blocking)

`src/api/services/charts/curated.py:44-53` calls resolve_chart_colors after _localise, and without any slug stamping. Registry colours are keyed by stable English slug, so color_map gets built from display labels and every category falls through to the hashed fallback palette. Verified against the real generator for dataset 8:

- registry:  logging #52A44E   wildfire #885128   permanent_agriculture #E39D29
- color_map: Logging #F58518   Wildfire #54A24B   Permanent agriculture #FF9DA6

This hits the three datasets that have categories in the registry and a categorical curated chart: 1 (land cover pie), 3 (natural lands pie), 8 (driver pie). It is not cosmetic-only: `color_map` is persisted to `insight_charts.color_map` and served by GET `/api/insights/{id}` (src/api/routers/insights.py:49), and the FE feeds it straight into `formatChartData` (ChartWidget.tsx:440). Before this PR `/api/analyze` emitted no colours at all, so this introduces confidently-wrong colours on the one path in this PR with default-profile reach (LGMS itself is unaffected, dataset 12 has no palette entry).

Looks like the unmerged branch analyze-chart-colors already solves exactly this, with the constraint spelled out in its own docstring: "Must run before _localise: that rewrites category labels into the reader's language, and color_map keys have to stay stable English slugs", plus a stamp_category_slugs step. Both branches edit src/api/services/analyze.py and both want to own the ordering inside the curated builder, so they will conflict.

**Recommend** dropping the colour call from this PR and set chart.dataset_id explicitly instead (that is the only thing the curated-only guards actually need from it, and getting it as a side effect of resolve_chart_colors is a fragile coupling either way), then let analyze-chart-colors add the stamp-then-resolve-then-localise sequence inside build_curated_charts.

2. A curated_only dataset with no generator reports "no data" (worth fixing here)

In `_analyze_curated_only`, if no generator handles the flagged dataset, build_curated_charts returns [] and the model is told "No <dataset> data is available for <area>. Tell the user politely; do not try another analysis." Invalid config surfaces as missing data, which is the opposite of the repo's fail-loudly rule. The import cycle rules out checking this in curated_only.py itself, so a unit test (every curated_only id has an entry in DETERMINISTIC_GENERATORS) or a check in registry.py is the cheap version.

01painadam:
I'll add some evals to `gnw-gold-evals` too

PR #830, opened 2026-09-17
Add lgms layers

solomon-negusse:
## Add LGMS sector map layers

This adds real LGMS layers served by the tile cache service

**Layers.** `land_ghg_inventory.yml` declares five layers against the v1.0.3 dynamic tile endpoint, replacing the two that pointed at a shared COG mosaic:

| layer | measure |
|---|---|
| `lgms` | net flux, whole land sector |
| `lulucf` | net flux, vegetation + soil |
| `agriculture` | gross emissions, cropland + livestock |
| `cropland` | gross emissions |
| `livestock` | gross emissions |

**Layer selection.** The `selected_layer` examples in `DatasetOption` covered only the old two-layer set. They now name all five and make the hierarchy explicit: `cropland`/`livestock` for a specific source, `agriculture` for one spanning both, `lulucf` for land-use/vegetation, `lgms` for the whole sector.

**Narration fix.** The agent had no way to know a multi-layer dataset renders at all — the `pick_dataset` tool message reported only the dataset name and context layer, and the session line only `Dataset: <name> (context: <ctx>)`. The selector sees `layers` and picks correctly, but the main agent narrated from the yml prose, which describes LGMS purely as tables: asked about agriculture emissions it would display the agriculture layer while replying *"there is currently no direct map layer for agricultural emissions."* Both surfaces now report the available layers and which one is displayed


yellowcap (commented):
Thanks @solomon-negusse this will be great to have in the system. I have one concern about this now that there are more layers: we should probably add a description field for which one to pick. Just the single keyword seems a bit too narrow/brittle for a solid llm selection. I also think that currently the flux type is not explicitly given to the agent, and we can drop the tile_url from the selection context, as we are not using that later and I think it does not add to an improved selection. 

Below a plan that I produced with Claude through some converation about this:


A. Layer descriptions — essential. One change, end to end. The layer name is the only thing that carries meaning today, and a bare token like lulucf or agriculture means nothing to the selector, to the narrating agent, or to the user reading the legend. Everything else here is secondary to fixing that.

- Catalog. Add optional title and description to each entry in a yml layers list. name stays the machine id — it matches the layer= URL param, validate_selected_layer_for_dataset (schema.py:96) and the widget-config snapshot, so nothing downstream moves.

layers:
  - name: agriculture
    title: Agriculture
    description: >-
      Gross emissions from cropland and livestock combined. No removals,
      so this is not a net flux. Use cropland or livestock for one source.
    tile_url: "...&layer=agriculture&flux_type=gross_emissions"

  Each description must state the measure. Three of the five LGMS layers are gross emissions and two are net flux; that distinction currently exists only inside a query string, so the agent can display a gross-emissions layer and call it a net flux.
- Schema. Add the two fields to DatasetLayer (schema.py:22) and read them in get_dataset_layers (tool.py:716).
- Selector. Render layers as a prompt block — name, title, description — instead of the raw CSV cell of dicts. Follows _format_selection_hints (tool.py:201).
- Field description. With the catalog carrying the vocabulary, cut the hard-coded LGMS clauses from selected_layer (schema.py:43) down to generic guidance: the layer whose description best matches the query's focus, plus the existing null rule. Otherwise every new multi-layer dataset appends another glossary to a prompt every query pays for.
- Narration. Carry title/description into the widget-config projection (add_map_widget.py:110), the pick_dataset tool message and the session line, so the agent says "Agriculture (gross emissions, cropland + livestock)" and the frontend legend has a human label.

Ships in one PR. The yml fields can land first as a no-op if you want the review split.

B. Drop tile_url from the selector's view of layers. Confirmed safe: the CSV is a throwaway copy, and selected_row comes from the untouched frame. Two notes — the copy shares dict objects with candidate_datasets, so build new dicts rather than popping; and context_layers and parameters cells leak URLs into the same CSV, which is the same pure-noise argument but a wider prompt change, so it belongs in its own commit. This roughly pays for A's added tokens.

C. Optional measure: net_flux | gross_emissions. Makes the distinction checkable instead of prose the agent may skim. Only worth it if A's descriptions turn out not to hold.

solomon-negusse:
Thanks for the thoughtful review @yellowcap  - addressed in 6d5c3452941a04c322f05409cdd986ef69b1b8a1.

yellowcap (approved):
Looks good! 👍🏽

PR #831, opened 2026-09-17
fix(agent): fix comparison-guardrail attribution and false claims

yellowcap:
## Summary

Two related fixes to the LGMS/curated-only comparison guardrail:

- **Wording (PZB-1422):** the single-area guardrail shown when a curated-only dataset (e.g. LGMS) is queried with multiple AOIs made the dataset the grammatical subject of the limitation (e.g. "LGMS is available for one area at a time"), which reads as the underlying data being unable to support comparisons. Reworded so Horizon's curated-chart rendering is the subject, not the data.
- **False completion claims (PZB-1421):** a user asked to compare two areas on LGMS; the agent correctly declined the comparison, then claimed it had displayed charts for the second (declined) area — no such chart existed. Nothing in the prompt stopped the model from narrating past a decline: reusing the previously selected area's chart as if it answered the declined comparison, or inventing an unrelated technical reason (e.g. "the dataset is too large") for the one-area limit.

## Changes

- `src/agent/i18n.py`: reworded the `pull_data.single_area_only` string (en/es/fr/pt/id) shown verbatim when `pull_data` is called with a curated-only dataset and more than one AOI.
- `src/agent/subagents/analyst/tool.py`: reworded the fallback guard in `_analyze_curated_only` to attribute the limit to Horizon, and added an explicit "no chart was produced" instruction.
- `src/agent/datasets/curated_only.py`: extended `CURATED_ONLY_SESSION_NOTE` (shown in the session block on every turn for a curated-only dataset) with the same "Horizon limitation, not the data" framing and an explicit ban on inventing a different reason.
- `src/agent/graph.py`: added two Policy rules — never claim a chart/analysis was produced/shown/updated unless a tool actually returned one this turn, and generalized the existing "relay and stop" nudge rule to plain declines that offer no options (not just clickable nudges).

## Why this didn't become a code-level guard instead

Considered refusing early in `generate_insights` whenever more than one AOI is selected, but multi-area comparison is a real, tested feature for non-curated datasets (`tests/tools/test_generate_insights.py::test_generate_insights_comparison`) — a blanket guard there would have broken it. The curated-only path already refuses multi-area pulls in `pull_data.py`; the actual gap is that nothing stops the model from narrating past that refusal, since this codebase has no `return_direct`/hard-stop mechanism — enforcement is necessarily prompt-level.

## Test plan

- [x] `tests/tools/test_generate_insights.py` (58 tests, includes the multi-area comparison case) — pass
- [x] `tests/unit/agent/tools/test_curated_only.py` — pass
- [ ] Ask the agent to compare/rank multiple areas using LGMS data and confirm the reply attributes the limit to Horizon, not the data, offers to narrow to one area, and does not claim any chart was shown for the declined area
- [ ] Ask a global/world question on LGMS and confirm the agent does not invent a "dataset is too large" reason for asking to narrow to one area


01painadam:
Looks good across multiple tests

01painadam (approved):

01painadam:
Two suggested instruction follow-ups to tighten:

1. Make the decline rule require a verbatim quote of the tool's sentence, or the reworded strings keep getting paraphrased away.
  - pull_data — user-facing, quotable (src/agent/i18n.py:348)
  "Horizon shows {dataset_name} for one area at a time. It does not yet support comparing or ranking areas. Please     choose one area."
  -  generate_insights via _analyze_curated_only — hybrid, not quotable (src/agent/subagents/analyst/tool.py:623)
  "Horizon shows {dataset_name} for one area at a time; it does not yet support
  comparing or ranking areas. Tell the user politely that this is a Horizon
  limitation, not a limitation of the data, and ask them to choose one area"
2. (Future) For curated-only datasets, remove both interpretation licences and make the reply the fixed narrative plus a pointer to the info and metadata buttons.

yellowcap:
@01painadam would you make a follow-up ticket for the additional improvements?

PR #832, opened 2026-09-17
fix(insights): id agent charts by their insight so the frontend can group them

01painadam:
## Summary

The agent's LGMS charts reach the frontend as four unrelated cards: the three
net-flux roll-ups never fold behind a DETAIL pill, and an analysis's charts
show back-to-front in the pager. This gives every chart an id that names the
insight it belongs to, which is the handle the frontend groups on.

`to_frontend_dict()` now takes the owning `insight_id` and emits
`{insightId}-chart-{position}`; the three places that put charts into agent
state pass it. Omitting it keeps the previous bare `chart_{position}`.

Relates to PZB-1392 work on LGMS rendering; pairs with
wri/project-zeno-next#683.

## Why the backend

Three of the four fields the frontend needs were already correct after #827 —
LGMS is `curated_only`, so `generate_insights` routes to
`_analyze_curated_only` and builds its charts through the same
`build_curated_charts` → `LGMSChartGenerator` step `/api/analyze` uses. The
`chart_type`, `title` and `series_fields` all arrive right. Only the id was
wrong.

The frontend could have minted the id itself (`RestAnalysisGateway` already
does exactly that for the REST path), but the backend already owns this id,
and fixing it here also clears two latent bugs that affect **every** dataset,
not just LGMS:

- `orderInsightsForPager` falls back to a per-chart key and reverses, so the
  charts *within* one analysis render in reverse order.
- `chart_0` repeats across analyses in a session, so ids collide in the
  insight store.

Recall paths (`search_insights`, `update_insight_display`) are included
deliberately, so an LGMS analysis pulled back from history collapses like a
fresh one.

## Contract

`charts_data[].id` is a contract-2 payload. Consumers checked:

- No backend code emits `[Chart N]` markers.
- gnw-evals and gnw-gold-evals read `charts_data[].insight` and `[].type`
  only, never `id`.
- Frontend `develop` already carries the whole net-flux feature, so no
  frontend release is needed for the id itself; PR #683 adds the chat-side
  fold on top.

Non-matching ids stay valid: `chartBatchKey` reads them as "no siblings".

## Test plan

- [x] `pytest tests/unit` — 659 passed
- [x] `pytest tests/api` — 354 passed (against an isolated test Postgres on a
      named volume; the shared :5434 container is throwing the known
      bind-mount `InsufficientPrivilegeError` ownership noise)
- [x] `pre-commit run --all-files` — clean
- [x] New tests pin the `{insightId}-chart-{n}` shape at three levels: the
      model's id and its bare fallback, the four LGMS charts sharing one
      group key, and the agent's curated-only path end to end. The regex is
      copied verbatim from the frontend's `chartBatchKey`.

## Local end-to-end

Backend on this branch, frontend on `lgms/fix-agent-render` (#683),
`?agent_profile=experimental&ff=net-flux`, asking for the land GHG inventory
for Pará.

`GET /api/threads/{id}/state` for insight `ca224a67-…`:

| id | type | title | seriesFields |
|---|---|---|---|
| `…-chart-0` | hierarchical-bar | Net GHG Flux — Annual Average | 0 |
| `…-chart-1` | stacked-bar-with-line | Net GHG Flux — Full Detail | 11 |
| `…-chart-2` | stacked-bar-with-line | Net GHG Flux by Category | 6 |
| `…-chart-3` | stacked-bar-with-line | Net GHG Flux Summary | 3 |

On screen: the MEASURE pill on the annual-average chart, the DETAIL pill
offering Summary / Category / Full, and "2 of 2 available analyses" where
four charts were sent. The DETAIL pill is the direct proof — in the map
workspace `NetFluxToolbar` resolves siblings through `chartBatchKey` and
hides the pill when grouping fails.

The analytics API refuses this account (`role: USER` → 403 "Requires a
ResourceWatch admin account"), so the pull was served by a local stub at
`ANALYTICS_API_BASE_URL`, the seam the handler already documents. The stub
lived outside the repo and its rows are invented — the figures on screen are
not measurements, only the plumbing was under test.

https://claude.ai/code/session_01LDw5V6swtikLFcMcmafYWF


yellowcap (approved):
Looks good, Id-away

PR #834, opened 2026-09-18
PZB-1051: feat(custom-areas): accept GeoJSON in the area upload endpoint

yellowcap:
## Summary

This follow-up to #807 adds GeoJSON to `POST /api/custom_areas/upload`. This PR is stacked on #807 (base `feat/upload-csv-shp`). Merge #807 first.

- `.geojson` and `.json` files go to a new `parse_geojson` in `src/api/services/area_upload.py`. The write path does not change: the parser returns the same `ParsedFeature` list.
- The contract is the same as for CSV and shapefile:
  - Each feature needs a `name` property, matched case-insensitively (`name` and `NAME` together is an error).
  - Geometries must be `Polygon` or `MultiPolygon`.
  - Coordinates must be WGS84 lon/lat (RFC 7946). The CSV bounds check applies, with no slack.
  - The limit is 500 features.
  - Validation is all-or-nothing. Each error has a feature index (`feature N: …`).
- The top level must be a `FeatureCollection` or one `Feature`. A bare geometry is rejected because it has no properties for the name.
- A legacy `crs` member is accepted only if it names WGS84. Any other `crs` gives a 422, so a projected file is not stored at the wrong location.
- Other properties are stored without change, so numbers, booleans, lists and objects keep their JSON types. CSV properties are always strings.
- The 415 message and `docs/area-uploads.md` now include GeoJSON. KML stays out of scope.

## Frontend

The frontend PR wri/project-zeno-next#688 adds CSV and zip upload to the areas panel. It keeps the existing GeoJSON path through `POST /api/custom_areas`, which creates one merged area with a random name. After this PR merges, the dialog can send `.geojson` through `/upload` too. That is a product decision, because users then get one area per feature, and each feature needs a `name` property.

## Open point

The upload endpoint has no area-size check. The frontend's single-area GeoJSON path rejects areas smaller than 0.1 km² or larger than 10,000 km². This PR records that difference under "Out of scope" in the docs and does not add a limit.

## Test plan

- [x] `tests/api/test_area_upload.py`: 45 passed. The new tests cover an API upload with a mirror check, all per-feature errors in one 422 (through a `.json` filename), a single `Feature`, invalid files (UTF-8, JSON, top-level type, empty file, `features` not a list, non-WGS84 `crs`, duplicate name keys, malformed geometry, non-Feature entries), a WGS84 `crs` member, and the feature cap.
- [x] `tests/api/test_custom_area.py` passed.
- [x] Manual: a `.geojson` upload to a local API returns 200 and creates the area.

### DIFF
--- a/db/alembic/versions/e7c1f4a92b58_add_custom_area_properties_and_upload_batch.py
+++ b/db/alembic/versions/e7c1f4a92b58_add_custom_area_properties_and_upload_batch.py
@@ -1,0 +1,42 @@
+"""add custom_areas properties and upload_batch_id
+
+Revision ID: e7c1f4a92b58
+Revises: d4a1c7b93e02
+Create Date: 2026-08-29 12:00:00.000000
+
+``properties`` carries the non-geometry attributes of an uploaded feature and is
+projected into ``aois.properties`` by the custom-area mirror. ``upload_batch_id``
+groups the rows created by one file upload; a drawn area leaves it null. No
+index on ``upload_batch_id`` yet — nothing queries by it.
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+
+# revision identifiers, used by Alembic.
+revision: str = "e7c1f4a92b58"
+down_revision: Union[str, None] = "d4a1c7b93e02"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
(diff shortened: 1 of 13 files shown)


batpad (approved):

yellowcap:
Moving `.geojson` onto `POST /api/custom_areas/upload` deletes the only validation that the old GeoJSON path had and the new endpoint does not: a 0.1–10,000 km² area check. The old check applied to the merged area. The new endpoint creates one area per feature. Should the endpoint reject features outside that range before we ship this? If yes, should the check apply per feature? It also has an effect on CSV and shapefile uploads, which never had a limit.

(Context: the frontend change is on wri/project-zeno-next#688; the endpoint is wri/project-zeno#834.)

PR #835, opened 2026-09-18
PZB-1429: Set explicit end date

01painadam:
Explicitly set the LGMS dataset's end date to 2024. Leaving it open ended is causing the FE to set it to 2026:

Screenshot 2026-09-18 at 15 36 49


willian-viana (approved):
:shipit:

PR #839, opened 2026-09-23
Trace nudge clicks and message sources in Langfuse

01painadam:
## Why

A nudge click resubmits the option text as an ordinary chat message. Langfuse has no way to tell it apart from a typed message. The FE already sends `query_type: "human_input"` on a click, but `ChatRequest` has no such field, so pydantic drops it. Each trace carries only the user id and the session id. We can't measure nudge uptake by type, and the trace export can't flag clicks.

This is Part 1 of the nudge tracing and follow-up nudges plan. Part 2, replacing prose follow-up offers with a `follow_up` nudge, comes later and needs this data first.

## What changes

- **`ChatRequest`** gains `input_source` and `nudge_response`:
  - `input_source` is one of `typed | nudge | starter_prompt | dashboard_chip | dashboard_module | analyse_nudge | url_prompt | map_action`, default `typed`. Any other value is a 422.
  - `nudge_response` is `{type, option_index}`. Sending it with any source other than `nudge` is a 422. An empty `type` (the free-form `send_nudge` can produce one) is normalised to None rather than rejected.
  - Legacy `query_type: "human_input"` still maps to `nudge`, for FE builds that predate this change. It is resolved once in a model validator, and nothing downstream reads `query_type`.
- **`src/api/services/chat_trace.py`** (new, pure functions):
  - `match_pending_nudge` is a server-side cross-check against the thread's pending `state.nudge.options`. It is an exact match after stripping whitespace, and case-sensitive. It reuses the checkpoint read `stream_chat` already makes, so chat gets no new I/O and no new way to fail.
  - `build_turn_trace_metadata` builds the tags and metadata.
- **Per-turn Langfuse output:**
  - Trace name: `chat_turn`, set via `config["run_name"]`.
  - Tags: `input:<source>`, plus `nudge:<type>` for clicks.
  - Metadata: `input_source`, `nudge_type`, `nudge_option_index`, `nudge_match`, `nudge_match_type`, `nudge_match_index`, `ui_action_only`, `ff`, `page`. Keys whose value is None are dropped.
- **Ingest:**
  - `langfuse_traces` gains nullable `input_source`, `nudge_type`, `nudge_match`, plus an index on `input_source`.
  - `parse.py` reads these from trace metadata and falls back to the tags. Older traces stay NULL, never `typed`.
  - `PARSER_VERSION` goes to 3.

The agent never sees any of this; it is tracing only.

## Heads-up

- **Trace rename:** traces were all named `LangGraph` and are now named `chat_turn`. No code filters on the old name (zeno ingest, gnw-traces and the eval harnesses were checked), but **saved Langfuse UI views or dashboards that filter on `LangGraph` need updating**.
- **Migration:** revision `9f723bd32c77`, chained from `e7c1f4a92b58`. Apply it with `alembic upgrade head`. Filling in old rows would mean re-ingesting that window.
- **Known limitation:** `state.nudge` is last-write-wins and is never reset each turn. So `nudge_match` can report a stale option as a match, and can miss a click on an old button still shown in the chat history. This is documented in the docstring. The client's `input_source` is the primary signal.
- **Deploy order:** merge together with the FE PR below. This BE returns 422 for any `input_source` value it doesn't know. An older BE ignores the new fields.

## Paired FE PR

wri/project-zeno-next: https://github.com/wri/project-zeno-next/pull/710

FE files affected: `app/types/chat.ts` (`InputSource`, `SendSource`, `ChatPrompt`), `app/utils/chatInputSource.ts` (the one serialiser), `chatStore.sendMessage`, and every caller of it.

## Testing

- `uv run pytest tests/unit tests/api`: 1126 passed, against a throwaway postgis:17-3.5. The later empty-type fix reran its touched unit files: 51 passed.
- `uvx pre-commit run --all-files`: all hooks passed.
- Migration upgrade, then downgrade -1, then upgrade: clean.
- Local e2e: the real BE ran from this branch, with the Langfuse SDK exporting to a local OTLP sink, and I decoded the exported spans.
  - A typed turn exports as name `chat_turn`, tags `["input:typed"]`, `nudge_match=false`.
  - "Tree cover loss in Santa Cruz" produced an `aoi_choice` nudge. Clicking option 0 exported tags `["input:nudge","nudge:aoi_choice"]`, `nudge_option_index=0`, `nudge_match=true` with the matched type and index.
  - An unknown source returned 422, and `nudge_response` with a `typed` source returned 422.
- To confirm on staging: that `nudge_match` arrives as a boolean after Langfuse's OTel ingest. Ingest accepts `"true"`/`"false"` strings as well, just in case.


yellowcap:
Closing this. Nudge offers are already visible through the nudge tool spans in Langfuse. We can get click uptake from the trace data we already ingest: each trace output holds the `nudge` state, so we can compare turn N's prompt against turn N-1's options in a query. That also works for old traces.

For tracing data alone, this change costs too much: a new API contract, a paired FE deploy, a migration, and a 422 on unknown `input_source` values that can break chat. If we need per-source data later (starter prompts, dashboard chips), we can do a smaller change: an optional free-form field that goes into the metadata and never rejects a request.

Paired FE PR: wri/project-zeno-next#710 (also closed).

